PR #55, opened 2018-07-15
[MRG] bug fixes, test additions, `meg_bids` -->`mne_bids` ... Paving the way for EEG inclusion

sappelhoff:
starting to include EEG into mne-bids - this is work in progress and I already open the PR to allow for feedback on the way


jasmainak:
okay, keep us posted when we should give it a try

sappelhoff:
btw, this is WIP for #34 *linking*.

@jasmainak I'd appreciate an interim review!

mne_bids/datasets.py
```diff
@@ -38,3 +41,67 @@ def fetch_faces_data(data_path, repo, subject_ids):
             tf.extractall(path=data_path)
             shutil.move(op.join(data_path, 'ds117'), target_dir)
             os.remove(tar_fname)
+
+
+def download_matchingpennies_subj(url_dict=None, directory='.'):
```
jasmainak: couple of questions

1. this dataset is already in BIDS format?
2. how big is it?
3. is it going to be used in an example somehow?

mne_bids/mne_bids.py
```diff
@@ -107,26 +115,55 @@ def _events_tsv(events, raw, fname, event_id, verbose):
...
-        Dictionary mapping a brief description key to an event id (value). For
-        example {'Go': 1, 'No Go': 2}.
+    trial_type : dict | None
+        Dictionary mapping a brief description key to an event id (event_value)
+        For example {'Go': 1, 'No Go': 2}. This corresponds to the 'trial_type'
+        column in BIDS.
+    event_type : dict | None
+        Dictionary mapping a brief description key to an event id (event_value)
+        if this is different from 'trial_type'. Corresponding to the
+        'event_type' column in BIDS.
```
jasmainak: is this a new column in BIDS-EEG? It doesn't exist in the main specification ...

mne_bids/mne_bids.py
```diff
@@ -328,14 +396,18 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
     events_data : str | array | None
         The events file. If a string, a path to the events file. If an array,
         the MNE events array (shape n_events, 3). If None, events will be
-        inferred from the stim channel using `find_events`.
-    event_id : dict
-        The event id dict
+        inferred from the stim channel using `mne.find_events`.
+    event_id : dict | None
+        The event id dict used to create a 'trial_type' column in events.tsv
+    event_type : dict | None
```
jasmainak: same question as before. Which field is this?

mne_bids/datasets.py
sappelhoff: 1. Yes, I added it to offer a quick and easy access to straight forward EEG files (only 2 markers corresponding to two conditions in a single task ... 10 electrodes EEG, that's it for the data)
2. The whole dataset something between 5 and 6 GB. For testing, I just download a channels file which is several kB - for actually doing something with it, one subj suffices, which is <1GB
3. Yes, I'll use it in the EEG example later on

mne_bids/mne_bids.py
sappelhoff: see above, "event_type" is a new BIDS EEG field to describe event_values (=event_id) in case the description is different from "trial_type". Generally, you can think about it this way: If we have an experimental trial, there might be several trigger codes nested in that trial: one signifying a button press, another signifying the onset of an additional stimulation or feedback, etc. --> these things can go into "event_type", while "event_id" covers the BIDS trial_type column (previously "condition" in this code)

mne_bids/mne_bids.py
```diff
@@ -107,26 +115,55 @@ def _events_tsv(events, raw, fname, event_id, verbose):
...
     sfreq = raw.info['sfreq']
     events[:, 0] -= first_samp
 
-    df = pd.DataFrame(np.c_[events[:, 0], np.zeros(events.shape[0]),
-                            events[:, 2]],
-                      columns=['onset', 'duration', 'condition'])
-    if event_id:
-        event_id_map = {v: k for k, v in event_id.items()}
-        df.condition = df.condition.map(event_id_map)
+    data = OrderedDict([['onset', np.c_[events[:, 0]].squeeze()],
```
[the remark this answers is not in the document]
jasmainak: yeah, because you don't explicitly say which axis to remove. It's better to index it with a 0, I believe ...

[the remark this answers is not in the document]
sappelhoff: how about `.reshape(-1)`?

mne_bids/mne_bids.py
```diff
@@ -509,17 +515,55 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
+                    raise ValueError('"%s" already exists. Please set'
+                                     ' overwrite to True.' % raw_file_bids)
             else:
-                raise ValueError('"%s" already exists. Please set overwrite to'
-                                 ' True.' % raw_file_bids)
+                if verbose:
+                    print('Writing data files to %s' % raw_file_bids)
+                # First assume we have single files
+                try:
+                    sh.copyfile(raw_file, raw_file_bids)
```
sappelhoff: @jasmainak for this most recent commit I need a bit of MEG experise.

What I changed is that if the actual data (not some BIDS meta data) is NOT present in the to-be-done BIDS-directory, it is getting copied there. This already needed a few tweaks to cover e.g., the multifile system for BrainVision (see code), which seems to work. Another tweak concerned CTF data, which apparently is always saved in a directory ending on `.ds`
____________
Now the issue is with KIT and BTi data ... see the travis [output](https://travis-ci.org/mne-tools/mne-bids/builds/405795025?utm_source=github_status&utm_medium=notification)

Due to the added line 561, I am now writing data to the BIDS dir, which has extensions that are not part of BIDS.

Before adding that line, it seems to me that no data was moved to the freshly made BIDS directory. Is that correct? How can we deal with this now?

.travis.yml
```diff
@@ -23,7 +23,8 @@ install:
       pip install --no-deps -e .
       cd ../
     - python setup.py install
-    - npm install -g https://github.com/INCF/bids-validator.git#0.26.0
+    # - npm install -g https://github.com/INCF/bids-validator.git#0.26.0
+    - npm install -g https://github.com/robertoostenveld/bids-validator.git#eeg-validator  # noqa use the EEG fork until merge
```
sappelhoff: This is the EEG fork for the bids-validator that we are working on. It's also applied for the BEP_006 branch in BIDS-examples, see [here](https://github.com/INCF/BIDS-examples/blob/bep006_eeg/.travis.yml) ... for MEG etc. it is the same

mne_bids/mne_bids.py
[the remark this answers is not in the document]
jasmainak: what is `np.c_[events[:, 0]].shape`?

mne_bids/mne_bids.py
```diff
@@ -236,13 +274,18 @@ def _sidecar_json(raw, task, manufacturer, fname, kind, verbose):
...
         Name of the task the data is based on.
     manufacturer : str
-        Used to define the coordinate system for the MEG sensors.
+        Manufacturer of the acquisition system. For MEG also used to define the
+        coordinate system for the MEG sensors.
     fname : str
         Filename to save the sidecar json to.
     kind : str
         Type of the data as in ALLOWED_KINDS.
+    eeg_reference : str | None
```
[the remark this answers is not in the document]
jasmainak: yes, but I meant to say that you should update the doc to reflect what happens when it is `None` :-)

mne_bids/mne_bids.py
jasmainak: okay your travis issue seems related to this issue on bids validator: https://github.com/INCF/bids-validator/pull/485 ? Am I right?

mne_bids/mne_bids.py
```diff
@@ -424,60 +506,64 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
+                                     ' overwrite to True.' % raw_file_bids)
             else:
-                raise ValueError('"%s" already exists. Please set overwrite to'
-                                 ' True.' % raw_file_bids)
+                if verbose:
+                    print('Writing data files to %s' % raw_file_bids)
+                # First assume we have single files
+                try:
+                    sh.copyfile(raw_file, raw_file_bids)
+                except IOError as e:
```
jasmainak: usually it's not a good practice to do `try-catch` constructs. You should try to avoid it if possible by checking file endings.

mne_bids/mne_bids.py
```diff
@@ -424,60 +506,64 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
+        else:
+            # All other data are single file formats
+            raw_file_bids_list = [raw_file_bids]
+            raw_file_list = [raw_file]
+
+        # Now copy&move or overwrite the files for the new BIDS directory
+        for raw_file, raw_file_bids in zip(raw_file_list, raw_file_bids_list):
+            if os.path.exists(raw_file_bids):
+                if overwrite:
+                    os.remove(raw_file_bids)
```
jasmainak: why is this necessary? Doesn't `sh.copyfile` overwrite the file?

mne_bids/tests/test_io.py
```diff
@@ -0,0 +1,29 @@
+"""Testing utilities for file io."""
+# Authors: Stefan Appelhoff <stefan.appelhoff@mailbox.org>
+#
+# License: BSD (3-clause)
+import pytest
+
+from mne_bids.io import _parse_ext, _read_raw
+
+
+def test__parse_ext():
```
jasmainak: accidental double underscore

sappelhoff: That was intentional (but maybe wrong): I was following the convention of defining a test function "fun" as `def test_fun`. In this case, fun was `_parse_ext`, thus `test_` + `_parse_ext` = double underscore. I take from your comment that this is not the convention?

mne_bids/mne_bids.py
sappelhoff: Mmmh yes, it seems like it although I lack the comprehensive knowledge about the file formats to be 100% sure ...

.travis.yml
jasmainak: okay great !

mne_bids/tests/test_io.py
jasmainak: oh well, I never followed this convention ...but as you like

mne_bids/mne_bids.py
jasmainak: maybe you can test it on your files and then put some subtle pressure on @chrisfilo to merge it :)

mne_bids/mne_bids.py
sappelhoff: What you mention could perhaps also be a solution. I am trying to stick to BEP_006 so let me try to argue --> I think that the event_type column might be used for really low level things like e.g.: Marker 1 signifies onset of a "famous face" trial then there is some jitter before the actual stimulus is shown, this could be Marker 2. --> for the "trial_type" column, which is described via event_id, these both markers get the entry "famous face" ... but in the "event_type" column, these might have the entries "trial onset" and "face_show_onset"

mne_bids/mne_bids.py
jasmainak: See if the PR actually fixes your problem. It could be a good check also for that PR

mne_bids/mne_bids.py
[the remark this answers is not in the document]
sappelhoff: hahaha, I was wondering about that as well. I didn't write this, I'll try `events[:, 0]` and see if this gets us into trouble

mne_bids/mne_bids.py
jasmainak: I guess there is still scope to change BEP_006 :) That's why it's a good idea to get your hands dirty to see if the standard makes sense in practice.

Anyway, it feels a bit like overengineering to me. If it's an optional field, maybe we could leave it out for now and address it in a future PR?

sappelhoff: if you insist, I'll remove it! :-) BEP_006 is unlikely to change, we're kind of wrapping things up now

mne_bids/mne_bids.py
sappelhoff: Yes, `sh.copyfile` overwrites ... sometimes we don't want that, that's why there is this clause I guess - I did not add it though, it's old

mne_bids/mne_bids.py
sappelhoff: okay, I'll fix it to:
```python
                if ext == '.ds':
                    sh.copytree(raw_file, raw_file_bids)
                else:
                    sh.copyfile(raw_file, raw_file_bids)
```

jasmainak: +1

mne_bids/mne_bids.py
```diff
@@ -251,6 +294,18 @@ def _sidecar_json(raw, task, manufacturer, fname, kind, verbose):
...
+    if isinstance(raw, BaseRaw):
+        recording_type = 'continuous'
+    elif isinstance(raw, BaseEpochs):
+        recording_type = 'epoched'
+    else:
+        recording_type = 'n/a'
+
+    if kind == 'eeg' and eeg_reference is None:
+        warn('No EEGReference found, defaulting to n/a')
+        eeg_reference = 'n/a'
```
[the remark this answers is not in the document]
jasmainak: See this code. Perhaps you can leverage something like this:

```py
import os.path as op
import numpy as np

import mne

data_path = op.join(mne.datasets.sample.data_path(), 'MEG', 'sample')
raw = mne.io.read_raw_fif(op.join(data_path, 'sample_audvis_raw.fif'),
                          preload=True)
raw.set_eeg_reference('average', projection=True)  # set EEG average reference

eeg_proj = raw.info['projs'][-1]
print(eeg_proj['data']['col_names'])
```

mne_bids/mne_bids.py
jasmainak: yeah, I'd like to keep things as simple and as automated as possible. I just noticed that there is also a field called `event_value`, which you can probably update automatically :-)

mne_bids/datasets.py
jasmainak: okay looking forward to the example!

mne_bids/mne_bids.py
sappelhoff: I copied your changes from the PR into my local source and then ran `sudo npm install -g .` ... the error persists

jasmainak: no time to look right now :( I need to run now. I'll look later if you don't figure out what's the issue

mne_bids/mne_bids.py
sappelhoff: okay, I'll remove it - `event_value` is inferred from mne's `events`

mne_bids/mne_bids.py
chrisfilo: merged and released

jasmainak: Thanks !

choldgraf:
I'll try and get to this soon-ish but I'm a bit swamped right now prepping for a talk next week, just FYI!

mne_bids/mne_bids.py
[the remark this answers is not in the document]
sappelhoff: Although I agree with trying to automate as much as possible, I think that we'll have to keep this field as manual (default None) ... your approach would only work for the case of applying an  MNE average reference to the data. All other references cannot be projections (see here: `ValueError: Setting projection=True is only supported for ref_channels="average".`)

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,66 @@
+"""
+=========================
+Use MNE-Bids for EEG Data
+=========================
+
```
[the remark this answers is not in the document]
jasmainak: you can into the "artifacts" tab of CircleCI. But right now, we don't have CircleCI configured for the repo. That would be another PR that would be welcome :-)

mne_bids/mne_bids.py
```diff
@@ -499,52 +499,43 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
     if len(events) > 0:
         _events_tsv(events, raw, events_tsv_fname, event_id, verbose)
 
-    # Before writing the neural data, cover some special cases:
-    # for FIF, we need to re-save the file to fix the file pointer
-    # for files with multiple parts
-    if ext in ['.fif', '.gz']:
-        raw.save(raw_file_bids, overwrite=overwrite)
+    # Write the neural data
+    # ---------------------
```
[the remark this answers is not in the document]
jasmainak: ok yeah. Although I would recommend doing it in a separate PR. Maybe it's a good idea to do a couple of separate PRs if the diff in this PR starts to grow really big ...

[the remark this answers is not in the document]
jasmainak: sounds good

mne_bids/mne_bids.py
```diff
@@ -499,43 +501,58 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
+            raw.save(raw_file_bids, overwrite=overwrite)
+        else:
+            if os.path.exists(raw_file_bids):
+                if overwrite:
+                    os.remove(raw_file_bids)
+                    sh.copyfile(raw_file, raw_file_bids)
+                else:
+                    raise ValueError('"%s" already exists. Please set'
+                                     'overwrite to True.' % raw_file_bids)
+
```
sappelhoff: Note that the saving procedure for MEG is as it was before this PR. there might be an error that somehow passed through the tests though: Checking the if-clauses, you save if:
- it's `.fif` or `.gz`
- if its any other MEG format AND the path exists AND `overwrite=True`

--> however, what if the format is NOT `.fif`/`.gz` and the path does not exist?

README.md
```diff
@@ -27,6 +33,17 @@ To check if everything worked fine, you can do:
 
 and it should not give any error messages.
 
+To run the tests using `pytest`, you need to have the cloned mne-python repository
+with a local pip-install instead of the mne-python package from pypi. Update your
+installation as follows:
+
+	$ git clone https://github.com/mne-tools/mne-python
+	$ cd mne-python
+	$ conda activate mne
```
jasmainak: this is confusing unless you also say that `environment.yml` from the installation page of MNE must be used

README.md
```diff
@@ -27,6 +33,17 @@ To check if everything worked fine, you can do:
 
 and it should not give any error messages.
 
+To run the tests using `pytest`, you need to have the cloned mne-python repository
```
jasmainak: I would move it to the contributions section. It's not relevant for regular users.

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,177 @@
+"""
```
jasmainak: I get:

```py
SyntaxError: Non-ASCII character '\xe2' in file /home/mainak/Desktop/projects/github_repos/mne-bids/examples/make_eeg_bids.py on line 145, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

when I tried to run this example

mne_bids/mne_bids.py
jasmainak: indeed, can you make a PR?

cc @monkeyman192

mne_bids/mne_bids.py
```diff
@@ -424,60 +493,66 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
+    # Check if it is EEG (preserving original file format)
+    # ----------------------------------------------------
+    elif ext in allowed_extensions_eeg:
+        if os.path.exists(raw_file_bids) and not overwrite:
```
jasmainak: this part should be common to both EEG and MEG

examples/make_eeg_bids.py
sappelhoff: How did you run it? I am using the mne-bids environment installed via conda and the mne-python and mne-bids `environment.yml` ... and then running `python mne_bids/examples/make_eeg_bids.py` gives me no errors ... (python 3 everything)

sappelhoff: but this is probably due to the `├──` pip characters I am using ... I thought they'd be pretty to show the directory structure. Generated with the `tree` bash function

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,177 @@
...
+# initial `dataset_description` on top!
+#
+# That's nice and it has saved us a lot of work. However, a few things are not
+# yet covered by MNE-BIDS for EEG data and will have to be added by hand. Most
+# importantly, the `README` file at the root of the directory, but also
+# `CHANGES`, `participants.tsv`, and some more specific entries.
+#
+# Remember that there is a convenient javascript tool to validate all your BIDS
+# directories called the "BIDS-validator", available here:
+# https://github.com/incf/bids-validator
```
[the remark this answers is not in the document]
sappelhoff: good point, I think I'll use: https://www.npmjs.com/package/bids-validator

... the web version does not contain a link to the source code as far as it seems? Should be changed maybe?

mne_bids/tests/test_utils.py
```diff
@@ -36,3 +41,50 @@ def test_make_folders():
...
+        copyfile_brainvision(src, dest)
+
+    # Assert we can move the new file
+    raw = read_raw_brainvision(os.path.join(new_data_dir,
+                                            new_basename + '.vhdr'))
+    assert isinstance(raw, BaseRaw)
+
+    # Assert that errors are raised
+    with pytest.raises(ValueError):
+        copyfile_brainvision('bogus.vmrk', 'bogus.vhdr')
```
jasmainak: can you also test that MNE is able to read in the copied files correctly? after the pointers have changed ...

mne_bids/utils.py
```diff
@@ -186,20 +204,40 @@ def make_dataset_description(path, name=None, data_license=None,
...
+    references_and_links : list | str | None
+        List of references to publication that contain information on the
+        dataset, or links.  Must be a list of strings or a single comma
+        separated string like ['a', 'b', 'c'].
     doi : str | None
         The DOI for the dataset.
+
+    Notes
+    -----
+    The required field BIDSVersion will be automatically field by mne_bids.
```
jasmainak: field -> filled ?

examples/make_eeg_bids.py
jasmainak: I used python 2.7 :-) That's why probably ...

examples/make_eeg_bids.py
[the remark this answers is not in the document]
jasmainak: You just right click and say "Inspect element" to see the source code :-) It's javascript ...

mne_bids/tests/test_utils.py
sappelhoff: see line 83, there is a typo in the comment though. It should read `# Assert we can read the new file`

mne_bids/utils.py
sappelhoff: +1

examples/make_eeg_bids.py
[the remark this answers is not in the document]
sappelhoff: True :-) I'll refer to both. But still made an issue on bids-validator, because that is easy and nice to have.

examples/make_eeg_bids.py
sappelhoff: mmmh as long as it's there, I might as well make a small effort for python 2 ;-) I replaced these with slightly more ugly dashes now that should not throw encoding errors anymore

README.md
sappelhoff: +1

README.md
sappelhoff: actually, the `environment.yml` in our mne-bids has `mne` as its name :-) so installing the mne-bids env.yml will give you an mne environment. That almost got me into trouble with my mne-python environment. Should I perhaps adjust the mne-bids environment.yml?

jasmainak: yeah but you need to first fetch the file and create the environment ... so you have 2 steps missing.

Also the `environment.yml` was really to make appveyor work and not necessarily for users

mne_bids/mne_bids.py
sappelhoff: yes, but let's wait until this one is merged? I'll have a lot of rebasing to do otherwise. For now, I will make an issue so that we don't forget

jasmainak: okay sounds good

mne_bids/mne_bids.py
sappelhoff: +1

mne_bids/tests/test_mne_bids.py
```diff
@@ -129,5 +132,7 @@ def test_brainvision():
                 output_path=output_path, kind='eeg', eeg_reference='Cz',
                 event_id={'test': 1})
 
-    cmd = ['bids-validator', output_path]
-    run_subprocess(cmd, shell=shell)
+    # Until EEG validator is merged, assert that this will be wrong.
+    with pytest.raises(subprocess.CalledProcessError):
+        cmd = ['bids-validator', output_path]
+        run_subprocess(cmd, shell=shell)
```
sappelhoff: once the bids-validator can manage EEG, we just habe to remove the `with pytest.raises()` line to make the test work.

jasmainak:
you're going to download the 356 MB file every single time the example is run? That's pretty tough on those without good internet connection :-)

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,181 @@
...
+#    ---------- README
+#    ---------- sub-xx
+#        ---------- modality
+#            ---------- sub-xx_task-yy_events.tsv
+#            ---------- sub-xx_task-yy_modality.ext
+#            ---------- sub-xx_task-yy_modality.json
+#
+#
+# Now, what does it actually look like?
+for root, dirs, files in os.walk(output_path):
```
jasmainak: should be a function since you duplicate it

jasmainak:
I get this when trying to build the doc:

```sh
/home/mainak/Desktop/projects/github_repos/mne-bids/doc/auto_examples/convert_ds000117.rst:77: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/mainak/Desktop/projects/github_repos/mne-bids/doc/auto_examples/make_eeg_bids.rst:3212: ERROR: Error in "code-block" directive:
maximum 1 argument(s) allowed, 13 supplied.

.. code-block:: python
   """
   Walk over a folder of files and create BIDS compatible folder.

   Parameters
   ----------
   subject_id : str
       The subject name in BIDS compatible format ('01', '02', etc.)
   task : str
       Name of the task the data is based on.
   raw_file : str | instance of mne.Raw
       The raw data. If a string, it is assumed to be the path to the raw
       data file. Otherwise it must be an instance of mne.Raw
   output_path : str
    """
/home/mainak/Desktop/projects/github_repos/mne-bids/doc/auto_examples/make_eeg_bids.rst:3228: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

examples/make_eeg_bids.py
jasmainak: in fact, you could probably add it as a public function in `mne-bids`. It's a nice way to visualize the folders

examples/make_eeg_bids.py
jasmainak: haha, my default env is still 2.7 for some reason. I should switch soon :)

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,181 @@
...
+# -------------------------
+#
+# First, we need some data to work with. We will use the example data that is
+# provided with the report on the BIDS extension for EEF data: The "Matching
+# Pennies" dataset. See here for more information: https://osf.io/cj2dr/
+#
+# For convenience, MNE-Bids has a function to download this data.
+
+# make a temporary directory to save the data to
+data_dir = _TempDir()
```
jasmainak: don't expose private function to user

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,181 @@
...
+
+###############################################################################
+# Now we have everything to start a new BIDS directory using our data.
+# To do that, we can use the high level function `raw_to_bids`, which is the
+# core of MNE-BIDS. Generally, `raw_to_bids` tries to extract as much meta data
+# as possible from the raw data and then format it in a BIDS compatible way.
+# `raw_to_bids` takes a bunch of inputs, most of which are however optional.
+# The required inputs are described in the docstring:
+#
+# .. code-block:: python
```
jasmainak: why not use sphinx gallery separators with the # symbol

sappelhoff:
> you're going to download the 356 MB file every single time the example is run? That's pretty tough on those without good internet connection :-)

won't the example be built once? ... and then it will be accessible over the documentation web page?

examples/make_eeg_bids.py
sappelhoff: Mmmh why not? The alternative would be to do some lines of codes with `os.mkdir`, potentially with an awkward `if os.path.exists`, or else making some hash to be sure it doesn't exist using `np.random.bytes()` --> seems longer and achieves the same

sappelhoff:
> I get this when trying to build the doc:

How are you trying to build the doc? So that I can do a bit of debugging myself. BTW: Shouldn't we build the docs automatically (and test them upon each build)? What would we need for that?

examples/make_eeg_bids.py
sappelhoff: I don't really know what you mean by that --> I just want to show the docstring of the required inputs. 

You mean I should do that using actual code, like `help(raw_to_bids)`? That would print out the entire docstring, whereas I only want to show the non-defaulting parameters that HAVE to be specified.

mne_bids/utils.py
```diff
@@ -21,6 +21,21 @@
 from .io import _parse_ext
 
 
+def print_dir_tree(dir):
+    """Recursively print a directory tree starting from `dir` [1].
```
sappelhoff: Should I add a test for this? It'd probably be a very trivial test, but for coverage reasons?

jasmainak: sure, +1 for that

examples/make_eeg_bids.py
jasmainak: private function should be private :-) you can save it in ~\mne_data\ if you like

jasmainak:
> How are you trying to build the doc? 

`make html`

> So that I can do a bit of debugging myself. BTW: Shouldn't we build the docs automatically (and test them upon each build)? What would we need for that?

you mean circle CI? sure that would be nice to set up

jasmainak:
> won't the example be built once? ... and then it will be accessible over the documentation web page?
if you do that, anyone trying to change the example will struggle to test things quickly.

examples/make_eeg_bids.py
jasmainak: oh I see, then I'd just remove this code block ...

sappelhoff:
>you're going to download the 356 MB file every single time the example is run? That's pretty tough on those without good internet connection :-)
>> won't the example be built once? ... and then it will be accessible over the documentation web page?
>>> if you do that, anyone trying to change the example will struggle to test things quickly

I am now downloading it to `~/mne_data/eeg_matchingpennies` instead of a temporary dir. So the downloading will happen only once :-)

sappelhoff:
@teonbrooks @choldgraf do you have time for a review?

@jasmainak any other comments?

jasmainak:
I have some urgent admin work to finish in the next couple of days. If @choldgraf or @teonbrooks can help in the meanwhile, it would be great. Otherwise, I'll take a look during the Neuroinformatics brainhack

choldgraf:
it's gonna be tough for me to find time (I'm getting married in 2 weeks and leaving for Europe in 1 week) but if I do, then I'll try and give this a review

examples/make_eeg_bids.py
teonbrooks: @jasmainak, wait, aren't we ending support for py2? ;) https://github.com/mne-tools/mne-python/pull/5256

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,173 @@
+"""
+=========================
+Use MNE-Bids for EEG Data
```
teonbrooks: Bids --> BIDS

examples/make_eeg_bids.py
```diff
@@ -0,0 +1,173 @@
...
+=========================
+
+In this example, we use MNE-Bids to create a BIDS-compatible directory of EEG
```
teonbrooks: Bids --> BIDS

mne_bids/datasets.py
```diff
@@ -38,3 +41,68 @@ def fetch_faces_data(data_path, repo, subject_ids):
...
+    ----------
+    url_dict : dict | None
+        Dictionary with keys corresponding to the type of BIDS file
+        and its extension. Dict values corresponding to valid urls
+        pointing to the respective files on the Open Science
+        Framework. MUST contain a key value pair
+        "subj_id: int" (see Examples). If None, will download data for sub-05.
+    subj_id : int | None
+        The subject identifier the url_dict corresponds to. If None,
+        will be set to 99.
```
teonbrooks: if it downloads `sub-05` by default, why not set this to `05`?

sappelhoff: it could be that a user specifies
```python
{'channels.tsv': 'https://osf.io/wzyh2/',
 'eeg.eeg': 'https://osf.io/mj7q8/'}
```
Where the links point to some other subject or not subject at all ... but I agree, this is a bit weird ... I'll think about a fix.

sappelhoff:
great, thanks for taking a look @teonbrooks.

Ready for merge @jasmainak ?

mne_bids/mne_bids.py
```diff
@@ -424,60 +463,35 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
-        The data as MNE-Python Raw object.
-
-    Returns
-    -------
-    events : array, shape = (n_events, 3)
-        The first column contains the event time in samples and the third
-        column contains the event id. The second column is ignored for now but
-        typically contains the value of the trigger channel either immediately
-        before the event or immediately after.
+    if ext in allowed_extensions_meg:
```
teonbrooks: you could generalize to use `ALLOWED_EXTENSIONS` since this will be used by all kinds

mne_bids/mne_bids.py
```diff
@@ -333,9 +367,9 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
     events_data : str | array | None
         The events file. If a string, a path to the events file. If an array,
         the MNE events array (shape n_events, 3). If None, events will be
-        inferred from the stim channel using `find_events`.
-    event_id : dict
-        The event id dict
+        inferred from the stim channel using `mne.find_events`.
+    event_id : dict | None
```
teonbrooks: since we haven't made a release yet, we could change this arg to be `trial_type` as you have in one of your helper functions if you think that would be clearer

mne_bids/tests/test_mne_bids.py
```diff
@@ -1,22 +1,24 @@
...
 import os.path as op
+from warnings import simplefilter
 
 import mne
 from mne.datasets import testing
 from mne.utils import _TempDir, run_subprocess
 
-from mne_bids import raw_to_bids
-from warnings import simplefilter
+from mne_bids.mne_bids import raw_to_bids
```
teonbrooks: any reason to not just use `from mne_bids import raw_to_bids` here?

sappelhoff: no reason, let's change it to what you suggest

mne_bids/mne_bids.py
sappelhoff: It's a dilemma ... one way to do it is closer to MNE, the other is closer to BIDS, and here we are in the middle of MNE-BIDS ;-) I am 50-50 on this. Someone else decide, please.

mne_bids/mne_bids.py
sappelhoff: this is already in preparation of the EEG if-else, which will go right below there. I just removed the EEG part in order to not merge it into 0.1.0 - that's why it might stick out a bit

teonbrooks: ok, that's fine

mne_bids/mne_bids.py
teonbrooks: i'm fine leaving it MNE-like here since users will be familiar with it. we can keep the internals more BIDS like as it is currently

mne_bids/mne_bids.py
```diff
@@ -363,12 +397,17 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
                         config=config, verbose=verbose)
         _, ext = _parse_ext(raw_file, verbose=verbose)
     elif isinstance(raw_file, BaseRaw):
+        # We got a raw mne object, get back the filename if possible
+        # NOTE: raw_file will be used as a string filename from here on!
```
jasmainak: maybe just rename as `raw_fname`?

mne_bids/utils.py
```diff
@@ -1,17 +1,49 @@
...
+from mne import read_events, find_events
+from mne.externals.six import string_types
+
+
+def print_dir_tree(dir):
+    """Recursively print a directory tree starting from `dir` [1].
+
+    References
+    ----------
+    .. [1]
```
jasmainak: incomplete ...

mne_bids/utils.py
```diff
@@ -1,17 +1,49 @@
...
-from mne.externals.six import string_types
+
 from .config import BIDS_VERSION
 
+import numpy as np
+from mne import read_events, find_events
+from mne.externals.six import string_types
+
+
+def print_dir_tree(dir):
```
jasmainak: you need to update `api.rst`

mne_bids/io.py
```diff
@@ -1,7 +1,17 @@
...
+#          Stefan Appelhoff <stefan.appelhoff@mailbox.org>
+#
+# License: BSD (3-clause)
 from mne import io
 import os
 
-ALLOWED_EXTENSIONS = ['.con', '.sqd', '.fif', '.gz', '.pdf', '.ds']
+allowed_extensions_meg = ['.con', '.sqd', '.fif', '.gz', '.pdf', '.ds']
+
+ALLOWED_EXTENSIONS = allowed_extensions_meg
```
jasmainak: This looks weird without the EEG extensions. Probably a candidate for next PR

jasmainak:
@sappelhoff thank you so much for simplifying the PR. Now it has a realistic chance of getting merged ;)

jasmainak:
you have merge commits in your history. Can you rebase and make a clean history? Thanks!

jasmainak:
I get this error:

```py
Unexpected failing examples:
/home/mainak/Desktop/projects/github_repos/mne-bids/examples/convert_ds000117.py failed leaving traceback:
  File "/home/mainak/Desktop/projects/github_repos/mne-bids/examples/convert_ds000117.py", line 72, in <module>
    event_id=event_id, output_path=output_path, overwrite=True)
  File "/home/mainak/Desktop/projects/github_repos/mne-bids/mne_bids/mne_bids.py", line 470, in raw_to_bids
    events = _read_events(events_data, raw)
  File "/home/mainak/Desktop/projects/github_repos/mne-bids/mne_bids/utils.py", line 333, in _read_events
    events = find_events(raw)
  File "<string>", line 2, in find_events
  File "/home/mainak/Desktop/projects/github_repos/mne-python/mne/utils.py", line 724, in verbose
    return function(*args, **kwargs)
  File "/home/mainak/Desktop/projects/github_repos/mne-python/mne/event.py", line 707, in find_events
    "length." % (n_short_events))
ValueError: You have 1 events shorter than the shortest_event. These are very unusual and you may want to set min_duration to a larger value e.g. x / raw.info['sfreq']. Where x = 1 sample shorter than the shortest event length.
```
This example works fine on `master`

sappelhoff:
hi @jasmainak 

> you have merge commits in your history. Can you rebase and make a clean history? Thanks!

What exactly are you proposing? Should I get rid of commits such as

```
   Merge pull request #31 from teonbrooks/cli
    
    [MRG] initial cli
```

?

I tried to do `git rebase master` from my branch, but that did not remove these messages.

jasmainak:
okay I rebased, added a cosmit and merged.

congrats @sappelhoff on the merge! :tada: Now you are the top contributor on `mne-bids` ;-)

jasmainak:
Looking forward to the EEG PR. Now, that the EEG validator PR is merged, I think it should be quite straightforward to do the testing on the MNE-BIDS end!

sappelhoff:
Definitely! What's the timeline for the 0.1.0 release? I'll start after that, I suppose.

mne_bids/mne_bids.py
```diff
@@ -424,60 +459,35 @@ def raw_to_bids(subject_id, task, raw_file, output_path, session_id=None,
...
-            raise ValueError('Events must have two dimensions, '
-                             'found %s' % events.ndim)
-        if events_data.shape[1] != 3:
-            raise ValueError('Events must have second dimension of length 3, '
-                             'found %s' % events.shape[1])
-        events = events_data
     else:
-        events = find_events(raw, min_duration=0.001)
-    return events
+        pass
```
jasmainak: @sappelhoff I just realized while reviewing https://github.com/mne-tools/mne-bids/pull/69 that this is not a good construct. You are violating two maxims of Zen of Python:

> Flat is better than nested.

and

> Errors should never pass silently. Unless explicitly silenced.

PR #259, opened 2019-08-14
DOC: Change Pypi description text

### LINKED ISSUE #258: Update text before release
The setup.py still states that this is experimental. I think it's grown up a fair bit :) this is the text that ends up on pypi. we should update it to reflect the package's aim

https://github.com/mne-tools/mne-bids/blob/f8f267b45ac36e1600ea9ceb5540299e1bf3ab21/setup.py#L17

sappelhoff:
closes #258

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,8 @@ if version is None:
     raise RuntimeError('Could not determine version')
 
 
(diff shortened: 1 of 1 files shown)


teonbrooks:
thanks @sappelhoff!

PR #271, opened 2019-10-01
[MRG] Deface mri during write_anat

alexrockhill:
PR Description
--------------

Using a numpy/scipy implementation, a shear method will be applied to erase the face for T1 data.

closes https://github.com/mne-tools/mne-bids/issues/239

progress on https://github.com/mne-tools/mne-bids/issues/269 

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -22,7 +22,7 @@ Current
 Changelog
 ~~~~~~~~~
 
--New feature in :func`mne_bids.write.write_anat` for shear deface of mri, by `Alex Rockhill`_ (`#271 <https://github.com/mne-tools/mne-bids/pull/271>`_)
+- New feature in :func`mne_bids.write.write_anat` for shear deface of mri, by `Alex Rockhill`_ (`#271 <https://github.com/mne-tools/mne-bids/pull/271>`_)
 
 
 Bug
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -22,6 +22,9 @@ Current
 Changelog
 ~~~~~~~~~
 
+-New feature in :func`mne_bids.write.write_anat` for shear deface of mri, by `Alex Rockhill`_ (`#271 <https://github.com/mne-tools/mne-bids/pull/271>`_)
+
+
 Bug
 ~~~
 
--- a/examples/convert_mri_and_trans.py
+++ b/examples/convert_mri_and_trans.py
@@ -157,7 +157,6 @@ plt.show()
 
 ###############################################################################
 # We can deface the MRI for anonymization
-#
 anat_dir = write_anat(bids_root=output_path,  # the BIDS dir we wrote earlier
                       subject=sub,
                       t1w=t1_mgh_fname,  # path to the MRI scan
--- a/examples/convert_mri_and_trans.py
+++ b/examples/convert_mri_and_trans.py
@@ -155,6 +155,27 @@ for point_idx, label in enumerate(('LPA', 'NAS', 'RPA')):
               title=label)
 plt.show()
 
+#
+# We can deface the MRI for anonymization
+#
+anat_dir = write_anat(bids_root=output_path,  # the BIDS dir we wrote earlier
+                      subject=sub,
+                      t1w=t1_mgh_fname,  # path to the MRI scan
+                      session=ses,
+                      raw=raw,  # the raw MEG data file connected to the MRI
+                      trans=trans,  # our transformation matrix
+                      deface=True,
+                      verbose=True  # this will print out the sidecar file
+                      )
+
+# Our MRI written to BIDS, we got `anat_dir` from our `write_anat` function
+t1_nii_fname = op.join(anat_dir, 'sub-01_ses-01_T1w.nii.gz')
+
+# Plot it
+fig, ax = plt.subplots()
+plot_anat(t1_nii_fname, axes=ax,
+          title='Defaced')
+plt.show()
 
 ###############################################################################
 # .. LINKS
--- a/examples/convert_mri_and_trans.py
+++ b/examples/convert_mri_and_trans.py
@@ -165,6 +165,7 @@ anat_dir = write_anat(bids_root=output_path,  # the BIDS dir we wrote earlier
                       raw=raw,  # the raw MEG data file connected to the MRI
                       trans=trans,  # our transformation matrix
                       deface=True,
+                      overwrite=True,
                       verbose=True  # this will print out the sidecar file
                       )
 
@@ -173,8 +174,7 @@ t1_nii_fname = op.join(anat_dir, 'sub-01_ses-01_T1w.nii.gz')
 
 # Plot it
 fig, ax = plt.subplots()
-plot_anat(t1_nii_fname, axes=ax,
-          title='Defaced')
+plot_anat(t1_nii_fname, axes=ax, title='Defaced')
 plt.show()
 
 ###############################################################################
--- a/examples/convert_mri_and_trans.py
+++ b/examples/convert_mri_and_trans.py
@@ -155,7 +155,7 @@ for point_idx, label in enumerate(('LPA', 'NAS', 'RPA')):
               title=label)
 plt.show()
 
-#
+###############################################################################
 # We can deface the MRI for anonymization
 #
 anat_dir = write_anat(bids_root=output_path,  # the BIDS dir we wrote earlier
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -675,7 +675,7 @@ def test_write_anat(_bids_validate):
     assert vox_sum > vox_sum3
 
     with pytest.raises(ValueError,
-                       match='The raw object and trans must be provided'):
+                       match='The raw object, trans and raw must be provided'):
         write_anat(output_path, subject_id, t1w_mgh, session_id, raw=raw,
                    trans=None, verbose=True, deface=dict(inset=2),
                    overwrite=True)
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -543,6 +543,7 @@ def test_set(_bids_validate):
 def test_write_anat(_bids_validate):
     """Test writing anatomical data."""
     # Get the MNE testing sample data
+    import nibabel as nib
     output_path = _TempDir()
     data_path = testing.data_path()
     raw_fname = op.join(data_path, 'MEG', 'sample',
@@ -647,9 +648,6 @@ def test_write_anat(_bids_validate):
                    trans=trans, verbose=True, deface=False, overwrite=True)
 
     # test deface
-    if not has_nibabel():  # pragma: no cover
-        raise ImportError('This function requires nibabel.')
-    import nibabel as nib
     anat_dir = write_anat(output_path, subject_id, t1w_mgh,
                           session_id, raw=raw, trans=trans_fname,
                           verbose=True, deface=True, overwrite=True)
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -656,7 +656,7 @@ def test_write_anat(_bids_validate):
 
     anat_dir2 = write_anat(output_path, subject_id, t1w_mgh,
                            session_id, raw=raw, trans=trans_fname,
-                           verbose=True, deface=dict(inset=20.),
+                           verbose=True, deface=dict(inset=25.),
                            overwrite=True)
     t1w2 = nib.load(op.join(anat_dir2, 'sub-01_ses-01_T1w.nii.gz'))
     vox_sum2 = t1w2.get_data().sum()
@@ -675,7 +675,7 @@ def test_write_anat(_bids_validate):
     with pytest.raises(ValueError,
                        match='The raw object, trans and raw must be provided'):
         write_anat(output_path, subject_id, t1w_mgh, session_id, raw=raw,
-                   trans=None, verbose=True, deface=dict(inset=20.),
+                   trans=None, verbose=True, deface=True,
                    overwrite=True)
 
     with pytest.raises(ValueError, match='inset must be numeric'):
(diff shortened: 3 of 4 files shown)


jasmainak:
Great start! Can you add an example + tests?

mne_bids/write.py
```diff
@@ -1072,4 +1073,59 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+                          '`overwrite` is set to False. File: "{}"'
+                          .format(t1w_basename))
+
+    nib.save(t1w, t1w_basename)
+
     return anat_dir
+
+
+def _deface_t1w(t1w_data, method, mri_landmarks, theta=35):
+    # x: L/R L+, y: S/I I+, z: A/P A+
```
jasmainak: can you put a link to the code you copied this from?

mne_bids/write.py
```diff
@@ -1072,4 +1073,59 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+    if method == 'smooth':
+        raise ValueError('Smooth method not yet implemented')
+    elif method == 'erase':
+        t1w_data[deface_indices] = 0
+    else:
+        raise ValueError('Deface argument %s not recognized' % method)
+    return t1w_data
+
+
+def rotate(v, thetas):
```
jasmainak: private function

mne_bids/write.py
```diff
@@ -1072,4 +1073,59 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+def _deface_t1w(t1w_data, method, mri_landmarks, theta=35):
+    # x: L/R L+, y: S/I I+, z: A/P A+
+    normal = np.cross(mri_landmarks[0], mri_landmarks[1])
+    normal /= np.linalg.norm(normal)
+    deface_indices = np.zeros(t1w_data.shape, dtype=bool)
+    for i in range(t1w_data.shape[0]):
+        for j in range(t1w_data.shape[1]):
+            for k in range(t1w_data.shape[2]):
+                # translate to NAS, and shift up
+                v = np.array([i, j, k]) - mri_landmarks[1]
```
jasmainak: better to avoid single letter variable letters except for indices. They are unintelligible and harder to grep

mne_bids/write.py
```diff
@@ -978,6 +979,10 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : str | None
+        The method for defacing: If `smooth` a guassian kernal will be
+        used to smooth the face area, if `erase`, the image values in
+        the face area will be erased. If None, no defacing will be performed.
```
jasmainak: You need to tell in docstring that `raw` and `trans` should be provided. Otherwise, the user may feel scolded for no fault of his own :)

mne_bids/write.py
```diff
@@ -978,6 +980,14 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : tuple (float, float, str, bool) | None
```
agramfort: I would not deface by default. I would propose:

False : don't
True : do with (0.05, 35., 'erase', False) as defaults
tuple : (float, float, str, bool) 

so I can just do True to have something reasonably ok.

alexrockhill:
image

alexrockhill:
image

mne_bids/write.py
```diff
@@ -978,6 +980,17 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : bool | tuple (float, float, str, bool) | None
+        If True, deface with default parameters, if parameters are provided,
+        the order is inset, theta, method, plot. Here `inset`
```
jasmainak: I'd rather use a dict instead of tuple

mne_bids/write.py
```diff
@@ -1072,4 +1089,60 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         fname = t1w_basename.replace('.nii.gz', '.json')
         _write_json(fname, t1w_json, overwrite, verbose)
 
+        if deface is not None:
+            # x: L/R L+, y: S/I I+, z: A/P A+
```
jasmainak: still missing where this is copied from

alexrockhill: I figured that out from messing with the image... that is the directionality in mri voxel space

jasmainak: I thought the code was copied from FieldTrip somewhere, no?

mne_bids/write.py
```diff
@@ -978,6 +980,14 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : bool | tuple (float, float, bool)
+        If False, no defacing is performed.
+        If True, deface with default parameters.
+        If parameters are provided, the order is inset, theta, plot.
+        Here `inset` is how far back as a fraction of mri space to start
+        defacing relative to the nasion (default 0.2), `theta` is the angle of
+        the defacing shear (default 35 degrees), and `plot_on`
+        is whether or not to plot the results (default False).
```
```
If dict, can contain keys:
    inset  : where to start defacing relative to nasion, expressed as fraction of MRI dimensions (default=0.2)
    theta : angle of defacing shear (default=35 degrees)
 ```
Please don't plot in the same function

File: mne_bids/write.py [hunk 9]
```diff
@@ -1070,6 +1084,52 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+            mask = (coords[..., 2] < 0)
+
+            t1w_data[mask] = 0.
+            # smooth decided against for potential lack of anonymizaton
+            # https://gist.github.com/alexrockhill/15043928b716a432db3a84a050b241ae
+
+            t1w = nib.Nifti1Image(t1w_data, t1w.affine, t1w.header)
+            if plot_on:
+                fig = t1w.orthoview()
+                fig.show()
```
jasmainak: you'll soon have people asking to modify the plots, let's not do that. Modular is better.

mne_bids/write.py
alexrockhill: No, just useful for space transformations if this has to be done again

doc/whats_new.rst
```diff
@@ -22,6 +22,9 @@ Current
 Changelog
 ~~~~~~~~~
 
+-New feature in :func`mne_bids.write.write_anat` for shear deface of mri, by `Alex Rockhill`_ (`#271 <https://github.com/mne-tools/mne-bids/pull/271>`_)
```
agramfort:
```suggestion
- New feature in :func`mne_bids.write.write_anat` for shear deface of mri, by `Alex Rockhill`_ (`#271 <https://github.com/mne-tools/mne-bids/pull/271>`_)
```

mne_bids/tests/test_write.py
```diff
@@ -639,4 +644,59 @@ def test_write_anat(_bids_validate):
     # specify trans but not raw
     with pytest.raises(ValueError, match='must be specified if `trans`'):
         write_anat(output_path, subject_id, t1w_mgh, session_id, raw=None,
-                   trans=trans, verbose=True, overwrite=True)
+                   trans=trans, verbose=True, deface=False, overwrite=True)
+
+    # test deface
+    if not has_nibabel():  # pragma: no cover
+        raise ImportError('This function requires nibabel.')
```
agramfort: don't do this like this. Use a decorator so that test is just skipped if nibabel is not installed. Now I don't think the whole test should be skipped with nibabel is not available.

mne_bids/write.py
```diff
@@ -978,6 +980,15 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : bool | dict
+        If False, no defacing is performed.
+        If True, deface with default parameters.
+        If dict, accepts the following keys:
+            `inset`: how far back as a fraction of mri space to start
+                     defacing relative to the nasion (default 0.2)
```
agramfort: does this 0.2 value match to centimeters or some meaningful unit?

mne_bids/write.py
```diff
@@ -978,6 +980,15 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+            `theta`: is the angle of the defacing shear in degrees relative
+                     to the normal to the plane passing through the anatomical
+                     landmarks (default 35)
```
```suggestion
                     landmarks (default 35).
```

mne_bids/write.py
```diff
@@ -1070,6 +1096,56 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
...
+            idxs_meg = apply_trans(translation(y=trans_y), idxs_meg)
+            idxs_meg = apply_trans(rotation(x=-np.deg2rad(theta)), idxs_meg)
+            coords = idxs_meg.reshape(t1w.shape + (3,))  # (*t1w_data.shape, 3)
+            mask = (coords[..., 2] < 0)   # z < 0
+
+            t1w_data[mask] = 0.
+            # smooth decided against for potential lack of anonymizaton
+            # https://gist.github.com/alexrockhill/15043928b716a432db3a84a050b241ae
+
+            t1w = nib.Nifti1Image(t1w_data, t1w.affine, t1w.header)
```
agramfort: finally can you make this code go in a private function that would be in charge of all the parameter validation.

mne_bids/write.py
agramfort: can you just clarify what 0.2 means. What is fraction of mri space?

alexrockhill: 0.2 is a fraction of field of view of the mri, I'll look into changing it, I think it can be expressed in metric units

alexrockhill: It doesn't look like RAS has sensible units... the head is 60 something which is about double what it should be in cm. Is it better as a fraction or as a somewhat arbitrary unit? I'll keep trying to figure out what the units are actually

alexrockhill: It's in mm, changed accordingly!

mne_bids/write.py
```diff
@@ -978,6 +1048,15 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : bool | dict
+        If False, no defacing is performed.
+        If True, deface with default parameters.
+        If dict, accepts the following keys:
+            `inset`: how far back in millimeters to start defacing
+                     defacing relative to the nasion (default 20)
```
```suggestion
                     relative to the nasion (default 20)
```

mne_bids/write.py
```diff
@@ -978,6 +1048,15 @@ def write_anat(bids_root, subject, t1w, session=None, acquisition=None,
         also be a string pointing to a .trans file containing the
         transformation matrix. If None, no sidecar JSON file will be written
         for `t1w`
+    deface : bool | dict
+        If False, no defacing is performed.
+        If True, deface with default parameters.
```
```suggestion
        If True, deface with default parameters.
        `trans` and `raw` must not be `None` if True.
```

examples/convert_mri_and_trans.py
```diff
@@ -155,6 +155,27 @@
               title=label)
 plt.show()
 
+#
```
jasmainak:
```suggestion
###############################################################################
```

examples/convert_mri_and_trans.py
```diff
@@ -155,6 +155,27 @@
...
+# We can deface the MRI for anonymization
+#
```
jasmainak:
```suggestion

```

agramfort:
thx heaps @alexrockhill !

jasmainak:
Great work!

waiting for next PR ;-)

maybe this issue? https://github.com/mne-tools/mne-bids/issues/269

alexrockhill:
I think @bloyl is working on this but maybe he needs help?

jasmainak:
no, we need to add a parameter on `mne-bids` end so people can share anonymized data. What @bloyl will do is definitely going to help though.

alexrockhill:
@agramfort and all, what would you think of adding the smooth deface with a warning and not recommended prominently everywhere? Talking to @bloyl, we realized that people cannot redo coregistrations with sheared faces.

agramfort:
When you write the anat you pass the trans that gets written in bids. You don’t need to redo the Coreg

bloyl:
I haven't look at what the output of the proposed defacing so I don't know how destructive it is to downstream processing... but I think its worth considering having options that are the least destructive as possible provided they enable some additional level of downstream processing. Not necessarily now but potentially as future improvements.

For instance, in besa I don't think the mne trans file is going to be helpful, since besa uses its own internal head frame. So I would either need to redo coreg (in besa) or back engineer the besa stuff. I don't know how nutmeg or curry or other software works but I imagine something similar might happen.

Similarly, I imagine that full defacing could potentially break freesurfer, brainsuite, civit etc, which would lock a person downloading the data into the surfaces, bems etc created by the initial processing. Besa again is a concrete example as they use their own bems, segmentations etc.

just my 2cnts.

alexrockhill:
@bloyl the pictures of the defacing in both cases are above in the comments in this PR and the code to do the smoothing yourself and check the outputs is here: https://gist.github.com/alexrockhill/15043928b716a432db3a84a050b241ae. I'll check the freesurfer recon and bem later when I have time.

bloyl:
How light are we trying to keep the mne-bids dependencies? 

If we allow a dipy or simileITK dependency we could probably implement something like this https://www.biorxiv.org/content/biorxiv/early/2018/03/02/275453.full.pdf

agramfort:
if it can stay mne with minimal deps ie numpy + scipy it's ideal

>

alexrockhill:
I would be willing to do the script implementation and reconstruction and interpolation tests to see if the image can be recovered but it would take some time 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jasmainak:
This is less a priority for me than implementing a full anonymization -> calling mne functions for anonymize

agramfort:
+1

PR #286, opened 2019-11-14
in get_head_mri_trans, we need to allow_maxshield=True

dnacombo:
When read_raw is called within get_head_mri_trans, it needs to read even if maxshield isn't applied (see https://github.com/brainthemind/CogBrainDyn_MEEG_Pipeline_BIDS/blob/5be2e50d90b305cc2f4e8d72039de64f8b5879c7/11-make_forward.py#L83)

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -179,7 +179,7 @@ def _handle_channels_reading(channels_fname, bids_fname, raw):
     return raw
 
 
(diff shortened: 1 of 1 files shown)


agramfort:
we'll do it this way https://github.com/mne-tools/mne-bids/pull/284

thanks anyway @dnacombo

PR #379, opened 2020-03-27
TESTS: Fix timestamp bug in tests, extract FIF anonymization tests to separate function

### LINKED ISSUE #378: test_write.test_fif() fails on master with MNE-Python 0.19.2 when system clock is not set to UTC
#### Describe the bug
When running `tests.test_write.test_fif()` on my local machine, the test fails with an an assertion error related to an issue dealing with dates.

#### Steps to reproduce
- install `mne-bids:master` and `mne-python:maint/0.19` or `mne-python` 0.19.2
- set your computer clock to a timezone different from GMT/UTC (CET, i.e. UTC+1h, in my case)
- run `python -m pytest ./mne_bids/tests/test_write.py::test_fif`

#### Expected results
The test should pass.

#### Actual results
An assertion fails.
```python
>       assert data['acq_time'][0] == meas_date.strftime('%Y-%m-%dT%H:%M:%S')
E       AssertionError: assert '2002-12-03T19:01:10' == '2002-12-03T20:01:10'
E         - 2002-12-03T20:01:10
E         ?            ^^
E         + 2002-12-03T19:01:10
E         ?            ^^
```

#### Additional information
**The problem does not exist with `mne-python:master`.**

I recently fixed a different issue in a test that was also related to time / UTC (#363). We will need to check if the issue described here is of the same nature.

Also we should figure if it's possible to also run a CI test on a system that doesn't use a clock set to UTC to catch those issues. (Because obviously, `master` passes all tests on our CI systems, even with `mne-python` 0.19.2)

hoechenberger:
PR Description
--------------
- Fixes timestamp handling in a test, which closes #378.
- Moves anonymization tests for FIF files to their own function, which can now be easily skipped (via the `@pytest.mark.skipif` decorator) on MNE-Python <0.20, where this feature is not available.
- Adds a test with time zone set to Berlin to Travis test matrix.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,13 @@ matrix:
                MNE_VERSION="maint/0.19"
           name: "Linux pip minimal"
 
+        # Linux with timezone = Berlin, i.e. UTC+1 (CET) or UTC+2 (CEST)
+        - os: linux
+          env: CONDA_ENV="environment.yml"
+               MNE_VERSION="master"
+               TZ="Europe/Berlin"
+          name: "Linux conda full, timezone Europe/Berlin"
+
 # Specify version of BIDS-validator to be used, 'master', or 'stable'
 env:
   global:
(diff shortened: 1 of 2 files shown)


agramfort:
thx @hoechenberger

PR #390, opened 2020-04-16
[MRG] Adding updated unit tests and read_raw_bids correction for electrode coordinates (in meters)

adam2392:
PR Description
--------------

Closes: #389 

If you want to see that the added unit test breaks the existing code, comment out the changes in `_handles_electrodes_reading`.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,8 @@ jobs:
               conda update --yes --quiet conda
               conda env create -f environment.yml --name=testenv
               source activate testenv
-              pip install mne --upgrade
+#              pip install mne
+              pip install --upgrade --no-deps https://api.github.com/repos/mne-tools/mne-python/zipball/master
               python setup.py develop
 
         - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
               conda update --yes --quiet conda
               conda env create -f environment.yml --name=testenv
               source activate testenv
-              pip install mne==0.20
+              pip install mne
               python setup.py develop
 
         - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,6 @@ jobs:
               conda update --yes --quiet conda
               conda env create -f environment.yml --name=testenv
               source activate testenv
-#              pip install mne
               pip install --upgrade --no-deps https://api.github.com/repos/mne-tools/mne-python/zipball/master
               python setup.py develop
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
               conda update --yes --quiet conda
               conda env create -f environment.yml --name=testenv
               source activate testenv
-              pip install mne
+              pip install mne --upgrade
               python setup.py develop
 
         - run:
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -23,7 +23,6 @@ Changelog
 ~~~~~~~~~
 
 - :func:`read_raw_bids` now reads in participants tsv data, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
-- add EEG electrodes reading to :func:`mne_bids.read_raw_bids` and writing to :func:`mne_bids.write_raw_bids` by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -30,6 +30,7 @@ Bug
 - Fix bug in :func:`write_raw_bids` where raw.info['subject_info'] can be None, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`read_raw_bids` will now read all channels from electrodes.tsv. Channels with coordinates 'n/a' will also be included but their location set to `np.nan` in the `raw` object, by `Adam Li`_ (`#393 <https://github.com/mne-tools/mne-bids/pull/393>`_)
 - Do not change an events array passed to :func:`write_raw_bids` via the ``events_data`` keyword argument, by `Alexandre Gramfort`_ (`#402 <https://github.com/mne-tools/mne-bids/pull/402>`_)
+- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate in electrodes.tsv and also read in possible iEEG coordinateframes by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -31,7 +31,7 @@ Bug
 - Fix bug in :func:`write_raw_bids` where raw.info['subject_info'] can be None, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`read_raw_bids` will now read all channels from electrodes.tsv. Channels with coordinates 'n/a' will also be included but their location set to `np.nan` in the `raw` object, by `Adam Li`_ (`#393 <https://github.com/mne-tools/mne-bids/pull/393>`_)
 - Do not change an events array passed to :func:`write_raw_bids` via the ``events_data`` keyword argument, by `Alexandre Gramfort`_ (`#402 <https://github.com/mne-tools/mne-bids/pull/402>`_)
-- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinate frames by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
+- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv, and also read in possible iEEG coordinate frames via the 'space' BIDs-entity by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -30,7 +30,7 @@ Bug
 - Fix bug in :func:`write_raw_bids` where raw.info['subject_info'] can be None, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`read_raw_bids` will now read all channels from electrodes.tsv. Channels with coordinates 'n/a' will also be included but their location set to `np.nan` in the `raw` object, by `Adam Li`_ (`#393 <https://github.com/mne-tools/mne-bids/pull/393>`_)
 - Do not change an events array passed to :func:`write_raw_bids` via the ``events_data`` keyword argument, by `Alexandre Gramfort`_ (`#402 <https://github.com/mne-tools/mne-bids/pull/402>`_)
-- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinateframes by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
+- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinate frames by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -23,6 +23,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`read_raw_bids` now reads in participants tsv data, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
+- add EEG electrodes reading to :func:`mne_bids.read_raw_bids` and writing to :func:`mne_bids.write_raw_bids` by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -24,6 +24,7 @@ Changelog
 
 - :func:`read_raw_bids` now reads in participants tsv data, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`mne_bids.write_raw_bids` now uses the 'space' BIDS-entity when writing iEEG electrodes and coordinate frames by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
+- Fix :code:`convert_ieeg_to_bids` to now use sample ECoG EDF data by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -23,6 +23,7 @@ Changelog
 ~~~~~~~~~
 
 - :func:`read_raw_bids` now reads in participants tsv data, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
+- :func:`mne_bids.write_raw_bids` now uses the 'space' BIDS-entity when writing iEEG electrodes and coordinate frames by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -31,7 +31,7 @@ Bug
 - Fix bug in :func:`write_raw_bids` where raw.info['subject_info'] can be None, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`read_raw_bids` will now read all channels from electrodes.tsv. Channels with coordinates 'n/a' will also be included but their location set to `np.nan` in the `raw` object, by `Adam Li`_ (`#393 <https://github.com/mne-tools/mne-bids/pull/393>`_)
 - Do not change an events array passed to :func:`write_raw_bids` via the ``events_data`` keyword argument, by `Alexandre Gramfort`_ (`#402 <https://github.com/mne-tools/mne-bids/pull/402>`_)
-- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate in electrodes.tsv and also read in possible iEEG coordinateframes by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
+- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinateframes by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
 
 API
 ~~~
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -68,6 +68,7 @@ _IEEG_COORDINATE_FRAME_DICT = dict(
 _VERBOSE_IEEG_COORDINATE_FRAME = {val: key for key, val
                                   in _IEEG_COORDINATE_FRAME_DICT.items()}
 
+
 # mapping subject information back to mne-python
 # XXX: MNE currently only handles R/L,
 # follow https://github.com/mne-tools/mne-python/issues/7347
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -64,6 +64,8 @@ _IEEG_COORDINATE_FRAME_DICT = dict(
     mni_tal=FIFF.FIFFV_MNE_COORD_MNI_TAL,
     ras=FIFF.FIFFV_MNE_COORD_RAS,
     fs_tal=FIFF.FIFFV_MNE_COORD_FS_TAL,
+    head=FIFF.FIFFV_COORD_HEAD,
+    unknown=FIFF.FIFFV_COORD_UNKNOWN
 )
 _VERBOSE_IEEG_COORDINATE_FRAME = {val: key for key, val
                                   in _IEEG_COORDINATE_FRAME_DICT.items()}
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -55,7 +55,11 @@ ALLOWED_EXTENSIONS = {'meg': allowed_extensions_meg,
                       'eeg': allowed_extensions_eeg,
                       'ieeg': allowed_extensions_ieeg}
 
+# accepted BIDS formats, which may be subject to change
+# depending on the specification
+# XXX: Add support for MEG and EEG
 BIDS_IEEG_COORDINATE_FRAMES = ['acpc', 'pixels', 'other']
+BIDS_IEEG_COORDINATE_UNITS = ['m', 'cm', 'mm']
 
 # these coordinate frames in mne-python are related to scalp/meg
 # 'meg', 'ctf_head', 'ctf_meg', 'head', 'unknown'
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -1,4 +1,6 @@
 """Configuration values for MNE-BIDS."""
+from mne.io.constants import FIFF
+
 
 BIDS_VERSION = "1.2.2"
 
@@ -53,6 +55,18 @@ ALLOWED_EXTENSIONS = {'meg': allowed_extensions_meg,
                       'eeg': allowed_extensions_eeg,
                       'ieeg': allowed_extensions_ieeg}
 
+# these coordinate frames in mne-python are related to scalp/meg
+# 'meg', 'ctf_head', 'ctf_meg', 'head', 'unknown'
+# copied from "mne.transforms._str_to_frame"
+_IEEG_COORDINATE_FRAME_DICT = dict(
+    mri=FIFF.FIFFV_COORD_MRI,
+    mri_voxel=FIFF.FIFFV_MNE_COORD_MRI_VOXEL,
+    mni_tal=FIFF.FIFFV_MNE_COORD_MNI_TAL,
+    ras=FIFF.FIFFV_MNE_COORD_RAS,
+    fs_tal=FIFF.FIFFV_MNE_COORD_FS_TAL,
+)
+_VERBOSE_IEEG_COORDINATE_FRAME = {val: key for key, val
+                                  in _IEEG_COORDINATE_FRAME_DICT.items()}
 
 # mapping subject information back to mne-python
 # XXX: MNE currently only handles R/L,
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -55,10 +55,12 @@ ALLOWED_EXTENSIONS = {'meg': allowed_extensions_meg,
                       'eeg': allowed_extensions_eeg,
                       'ieeg': allowed_extensions_ieeg}
 
+BIDS_IEEG_COORDINATE_FRAMES = ['acpc', 'pixels', 'other']
+
 # these coordinate frames in mne-python are related to scalp/meg
 # 'meg', 'ctf_head', 'ctf_meg', 'head', 'unknown'
 # copied from "mne.transforms._str_to_frame"
-_IEEG_COORDINATE_FRAME_DICT = dict(
+MNE_IEEG_COORD_FRAME_DICT = dict(
     mri=FIFF.FIFFV_COORD_MRI,
     mri_voxel=FIFF.FIFFV_MNE_COORD_MRI_VOXEL,
     mni_tal=FIFF.FIFFV_MNE_COORD_MNI_TAL,
@@ -67,8 +69,8 @@ _IEEG_COORDINATE_FRAME_DICT = dict(
     head=FIFF.FIFFV_COORD_HEAD,
     unknown=FIFF.FIFFV_COORD_UNKNOWN
 )
-_VERBOSE_IEEG_COORDINATE_FRAME = {val: key for key, val
-                                  in _IEEG_COORDINATE_FRAME_DICT.items()}
+MNE_VERBOSE_IEEG_COORD_FRAME = {val: key for key, val
+                                in MNE_IEEG_COORD_FRAME_DICT.items()}
 
 
 # mapping subject information back to mne-python
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -443,8 +443,10 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             if coord_frame.lower() not in ['captrak', 'besa']:
                 warn("Not setting EEG montage because BIDS/mne-bids does not "
                      "support {} coordinate frame yet. "
-                     "Please use 'captrack', or 'besa'.".format(coord_frame))
+                     "Please use 'captrak', or 'besa'.".format(coord_frame))
                 coord_frame = None
+            else:
+                coord_frame = 'head'
         elif kind == "ieeg":
             coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
             coord_unit = coordsystem_json['iEEGCoordinateUnits']
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -438,7 +438,7 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             coord_frame = coordsystem_json['EEGCoordinateSystem']
             coord_unit = coordsystem_json['EEGCoordinateUnits']
 
-            if coord_frame not in ['captrak', 'besa']:
+            if coord_frame.lower() not in ['captrak', 'besa']:
                 warn("Defaulting to 'head' coordinate frame for "
                      "EEG montage because mne-bids does not "
                      "support {} coordinate frame yet.".format(coord_frame))
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -21,7 +21,8 @@ from mne.transforms import apply_trans
 from mne_bids.tsv_handler import _from_tsv, _drop
 from mne_bids.config import (ALLOWED_EXTENSIONS, _convert_hand_options,
                              _convert_sex_options,
-                             _IEEG_COORDINATE_FRAME_DICT)
+                             BIDS_IEEG_COORDINATE_FRAMES,
+                             MNE_IEEG_COORD_FRAME_DICT)
 from mne_bids.utils import (_parse_bids_filename, _extract_landmarks,
                             _find_matching_sidecar, _parse_ext,
                             _get_ch_type_mapping, make_bids_folders,
@@ -436,26 +437,40 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             coord_frame = coordsystem_json['MEGCoordinateSystem']
             coord_unit = coordsystem_json['MEGCoordinateUnits']
         elif kind == "ieeg":
-            coord_frame = coordsystem_json['iEEGCoordinateSystem']
+            space = coordsystem_json['iEEGCoordinateSystem']
             coord_unit = coordsystem_json['iEEGCoordinateUnits']
+            space = space.lower() # only compare lower-case
+
+            # XXX: improve reading from 'other' systems
+            if space not in BIDS_IEEG_COORDINATE_FRAMES:
+                warn("Coordinate frame is not accepted BIDS keyword for {}. "
+                     "Use CoordinateSystem with one of these keywords: "
+                     "{}".format(bids_fname, BIDS_IEEG_COORDINATE_FRAMES))
+                coord_frame = None
+            elif space == 'acpc':
+                coord_frame = 'ras'
+            elif space == 'other':
+                elec_params = _parse_bids_filename(electrodes_fname, verbose)
+                coord_frame = elec_params['space']
+
+                # default coordinate frames to available ones in mne-python
+                if coord_frame not in MNE_IEEG_COORD_FRAME_DICT:
+                    warn("Coordinate frame from coordinate system input {} "
+                         "is still not supported. Reading in coordinate frame "
+                         "as 'unknown'.".format(coord_frame))
+                    coord_frame = 'unknown'
         else:  # noqa
             # XXX should add support of coordsystem.json for EEG
             raise RuntimeError("Kind {} not supported yet for "
                                "coordsystem.json and "
                                "electrodes.tsv.".format(kind))
 
-        coord_frame = coord_frame.lower()  # MNE uses lower case
-        if kind == 'ieeg':
-            # XXX: improve reading with a `_write_dig_bids`
-            # default coordinate frames to available ones in mne-python
-            if coord_frame not in _IEEG_COORDINATE_FRAME_DICT:
-                warn("Coordinate frame from coordinate system input {} "
-                     "is still not supported. Reading in coordinate frame "
-                     "as 'unknown'.".format(coord_frame))
-                coord_frame = 'unknown'
-        # read in electrode coordinates and attach to raw
-        raw = _handle_electrodes_reading(electrodes_fname, coord_frame,
-                                         coord_unit, raw, verbose)
+        if coord_frame is not None:
+            coord_frame = coord_frame.lower()  # MNE uses lower case
+
+            # read in electrode coordinates and attach to raw
+            raw = _handle_electrodes_reading(electrodes_fname, coord_frame,
+                                             coord_unit, raw, verbose)
 
     # Try to find an associated sidecar.json to get information about the
     # recording snapshot
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -421,8 +421,6 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
     coordsystem_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                'coordsystem.json',
                                                allow_fail=True)
-    print("FOUND ELECTRODES FNAME: ", electrodes_fname)
-
     if electrodes_fname is not None:
         if coordsystem_fname is None:
             raise RuntimeError("BIDS mandates that the coordsystem.json "
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -248,9 +248,9 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame,
     electrodes_dict['z'] = [_float_or_nan(x) for x in electrodes_dict['z']]
     ch_names_raw = [x for i, x in enumerate(ch_names_raw)
                     if electrodes_dict['x'][i] != "n/a"]
-    ch_locs = list(zip(electrodes_dict['x'],
-                       electrodes_dict['y'],
-                       electrodes_dict['z']))
+    ch_locs = np.c_[electrodes_dict['x'],
+                    electrodes_dict['y'],
+                    electrodes_dict['z']]
 
     # determine if there are problematic channels
     nan_chs = []
@@ -263,13 +263,10 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame,
 
     # convert coordinates to meters if necessary
     if coord_unit in ['m', 'cm', 'mm']:
-        ch_locs = [_scale_coord_to_meters(coord, coord_unit)
-                   if np.nan not in coord
-                   else coord
-                   for coord in ch_locs]
+        ch_locs = _scale_coord_to_meters(ch_locs, coord_unit)
 
     # create mne.DigMontage
-    ch_pos = dict(zip(ch_names_raw, np.array(ch_locs)))
+    ch_pos = dict(zip(ch_names_raw, ch_locs))
     montage = mne.channels.make_dig_montage(ch_pos=ch_pos,
                                             coord_frame=coord_frame)
     raw.set_montage(montage)
@@ -436,10 +433,10 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
         # Get coordinate frames that electrode coordinates are in
         # Note: all coordinate frames should be lower-case
         if kind == "meg":
-            coord_frame = coordsystem_json['MEGCoordinateSystem'].lower()
+            coord_frame = coordsystem_json['MEGCoordinateSystem']
             coord_unit = coordsystem_json['MEGCoordinateUnits']
         elif kind == "ieeg":
-            coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
+            coord_frame = coordsystem_json['iEEGCoordinateSystem']
             coord_unit = coordsystem_json['iEEGCoordinateUnits']
 
             # XXX: improve reading with a `_write_dig_bids`
@@ -447,17 +444,18 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             if coord_frame not in _IEEG_COORDINATE_FRAME_DICT:
                 warn("Coordinate frame from coordinate system input {} "
                      "is still not supported.".format(coord_frame))
-                coord_frame = None
+                coord_frame = 'unknown'
         else:  # noqa
+            # XXX should add support of coordsystem.json for EEG
             raise RuntimeError("Kind {} not supported yet for "
                                "coordsystem.json and "
                                "electrodes.tsv.".format(kind))
 
-        if coord_frame is not None:
-            # read in electrode coordinates and attach to raw
-            raw = _handle_electrodes_reading(electrodes_fname, coord_frame,
-                                             coord_unit, raw,
-                                             verbose)
+        coord_frame = coord_frame.lower()  # MNE uses lower case
+
+        # read in electrode coordinates and attach to raw
+        raw = _handle_electrodes_reading(electrodes_fname, coord_frame,
+                                         coord_unit, raw, verbose)
 
     # Try to find an associated sidecar.json to get information about the
     # recording snapshot
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -231,7 +231,10 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame,
             raise RuntimeError(msg)
 
     if verbose:
-        print("The read in electrodes file is: \n", electrodes_dict)
+        summary_str = [(ch, coord) for idx, (ch, coord)
+                       in enumerate(electrodes_dict.items())
+                       if idx < 5]
+        print("The read in electrodes file is: \n", summary_str)
 
     def _float_or_nan(val):
         if val == "n/a":
@@ -261,7 +264,7 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame,
     # convert coordinates to meters if necessary
     if coord_unit in ['m', 'cm', 'mm']:
         ch_locs = [_scale_coord_to_meters(coord, coord_unit)
-                   if "n/a" not in coord
+                   if np.nan not in coord
                    else coord
                    for coord in ch_locs]
 
@@ -440,10 +443,10 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             coord_unit = coordsystem_json['EEGCoordinateUnits']
 
             # only accept captrak, or besa
-            if coord_frame.lower() not in ['captrak', 'besa']:
+            if coord_frame.lower() not in ['captrak']:
                 warn("Not setting EEG montage because BIDS/mne-bids does not "
                      "support {} coordinate frame yet. "
-                     "Please use 'captrak', or 'besa'.".format(coord_frame))
+                     "Please use 'captrak'.".format(coord_frame))
                 coord_frame = None
             else:
                 coord_frame = 'head'
@@ -453,7 +456,7 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
 
             # default coordinate frames to available ones in mne-python
             # noqa: see https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html
-            if coord_frame not in _ieeg_coordinate_frames_dict.keys():
+            if coord_frame not in _ieeg_coordinate_frames_dict:
                 warn("Defaulting coordinate frame to MRI "
                      "from coordinate system input {}".format(coord_frame))
                 coord_frame = 'mri'
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -19,13 +19,13 @@ from mne.coreg import fit_matched_points
 from mne.transforms import apply_trans
 
 from mne_bids.tsv_handler import _from_tsv, _drop
-from mne_bids.config import ALLOWED_EXTENSIONS, \
-    _convert_hand_options, _convert_sex_options
+from mne_bids.config import (ALLOWED_EXTENSIONS, _convert_hand_options,
+                             _convert_sex_options,
+                             _IEEG_COORDINATE_FRAME_DICT)
 from mne_bids.utils import (_parse_bids_filename, _extract_landmarks,
                             _find_matching_sidecar, _parse_ext,
                             _get_ch_type_mapping, make_bids_folders,
-                            _estimate_line_freq, _scale_coord_to_meters,
-                            _ieeg_coordinate_frames_dict)
+                            _estimate_line_freq, _scale_coord_to_meters)
 
 reader = {'.con': io.read_raw_kit, '.sqd': io.read_raw_kit,
           '.fif': io.read_raw_fif, '.pdf': io.read_raw_bti,
@@ -442,12 +442,12 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
             coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
             coord_unit = coordsystem_json['iEEGCoordinateUnits']
 
+            # XXX: improve reading with a `_write_dig_bids`
             # default coordinate frames to available ones in mne-python
-            # noqa: see https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html
-            if coord_frame not in _ieeg_coordinate_frames_dict:
-                warn("Defaulting coordinate frame to MRI "
-                     "from coordinate system input {}".format(coord_frame))
-                coord_frame = 'mri'
+            if coord_frame not in _IEEG_COORDINATE_FRAME_DICT:
+                warn("Coordinate frame from coordinate system input {} "
+                     "is still not supported.".format(coord_frame))
+                coord_frame = None
         else:  # noqa
             raise RuntimeError("Kind {} not supported yet for "
                                "coordsystem.json and "
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -24,7 +24,7 @@ from mne_bids.config import ALLOWED_EXTENSIONS, \
 from mne_bids.utils import (_parse_bids_filename, _extract_landmarks,
                             _find_matching_sidecar, _parse_ext,
                             _get_ch_type_mapping, make_bids_folders,
-                            _estimate_line_freq)
+                            _estimate_line_freq, _scale_coord_to_meters)
 
 reader = {'.con': io.read_raw_kit, '.sqd': io.read_raw_kit,
           '.fif': io.read_raw_fif, '.pdf': io.read_raw_bti,
@@ -195,7 +195,8 @@ def _handle_events_reading(events_fname, raw):
     return raw
 
 
-def _handle_electrodes_reading(electrodes_fname, coord_frame, raw, verbose):
+def _handle_electrodes_reading(electrodes_fname, coord_frame,
+                               coord_unit, raw, verbose):
     """Read associated electrodes.tsv and populate raw.
 
     Handle xyz coordinates and coordinate frame of each channel.
@@ -241,6 +242,8 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame, raw, verbose):
     electrodes_dict['x'] = [_float_or_nan(x) for x in electrodes_dict['x']]
     electrodes_dict['y'] = [_float_or_nan(x) for x in electrodes_dict['y']]
     electrodes_dict['z'] = [_float_or_nan(x) for x in electrodes_dict['z']]
+    ch_names_raw = [x for i, x in enumerate(ch_names_raw)
+                    if electrodes_dict['x'][i] != "n/a"]
     ch_locs = list(zip(electrodes_dict['x'],
                        electrodes_dict['y'],
                        electrodes_dict['z']))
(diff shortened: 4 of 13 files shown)


mne_bids/read.py
```diff
@@ -210,9 +211,18 @@ def _handle_electrodes_reading(electrodes_fname, coord_frame, raw, verbose):
...
-                                electrodes_dict['z'])))
+    ch_locs = list(zip(electrodes_dict['x'],
+                       electrodes_dict['y'],
+                       electrodes_dict['z']))
+
+    # convert coordinates to meters if necessary
+    if coord_unit in ['m', 'cm', 'mm']:
+        ch_locs = [_scale_coord_to_meters(coord, coord_unit)
+                   if "n/a" not in coord
+                   else coord
```
jasmainak, line 222: this feels dangerous to me ... you could have stuff in 2 coordinate systems

jasmainak: actually never mind. It's fine

mne_bids/tests/test_read.py
```diff
@@ -295,19 +296,65 @@ def test_handle_coords_reading():
...
+        for axis in ['x', 'y', 'z']:
+            electrodes_dict[axis] = np.multiply(orig_electrodes_dict[axis],
+                                                scaling_number[i])
+        _to_tsv(electrodes_dict, electrodes_fname)
+
+        # read in raw file w/ updated montage
+        raw_test = read_raw_bids(bids_fname, bids_root)
+
+        # obtain the sensor positions and make sure they're the same
+        for j in range(len(raw.info['dig'])):
```
jasmainak, line 324: avoid using single letter variable names, they are hard to Ctrl + F

mne_bids/tests/test_read.py
```diff
@@ -23,8 +23,9 @@
     import mne
 
 from mne.io import anonymize_info
-from mne.utils import _TempDir, requires_nibabel, check_version, object_diff
+from mne.utils import _TempDir, requires_nibabel, check_version
 from mne.datasets import testing, somato
+from mne.transforms import _str_to_frame
```
jasmainak, line 28: if it's just a tiny dictionary, please just copy it over. I think it's more legible that way + you don't risk breaking `mne-bids` when there is a new release of `mne-python`.

mne_bids/tests/test_read.py
```diff
@@ -295,19 +296,65 @@ def test_handle_coords_reading():
...
+        raw_test = read_raw_bids(bids_fname, bids_root)
+        for digpoint in raw_test.info['dig']:
+            assert digpoint['coord_frame'] == mri_coord_frame_int
+
+    # coordinate frames in mne-python should all map correctly
+    coordinate_frames = _str_to_frame.keys()
+    for coord_frame in coordinate_frames:
+        # these coordinate frames in mne-python are related to scalp/meg
+        if coord_frame in ['meg', 'ctf_head', 'ctf_meg', 'head', 'unknown']:
+            continue
```
jasmainak, line 344: this is a very weird code construction. First we import a private dictionary then we skip half of the keys. I would suggest just explicitly writing the coordinate frames like you do above.

mne_bids/tests/test_read.py
```diff
@@ -295,19 +296,65 @@ def test_handle_coords_reading():
...
+    coordinate_units = ['m', 'cm', 'mm']
+    scaling_number = [1., 100., 1000.]
+    coordsystem_fname = _find_matching_sidecar(bids_fname, bids_root,
+                                               suffix='coordsystem.json')
     electrodes_fname = _find_matching_sidecar(bids_fname, bids_root,
                                               "electrodes.tsv",
                                               allow_fail=True)
+    orig_electrodes_dict = _from_tsv(electrodes_fname,
+                                     [str, float, float, float, str])
+    for i, coord_unit in enumerate(coordinate_units):
```
jasmainak, line 310: avoid single letter variable names

jasmainak:
Is this only for iEEG? Could we also do it for EEG and MEG so that we're sure everything is handled in a unified way? Thanks a lot for your efforts @adam2392 !!

adam2392:
> Is this only for iEEG? Could we also do it for EEG and MEG so that we're sure everything is handled in a unified way? Thanks a lot for your efforts @adam2392 !!

Yeah no problem! I like contributing code into the mne-packages. It makes me a better programmer, and reduces LOC in my research related to data preprocessing :).

Re MEG: So I'm not familiar w/ MEG as much, but I think the coordinate unit handles it there as well, automatically. Not sure about coordinateframes?

Re EEG: What happened to the scalp EEG PR: https://github.com/mne-tools/mne-bids/pull/263?

I think for scalp EEG it actually is a bit simpler as well, cuz coordinate frames are relatively limited. It's just a head. But also as long as coordinate unit is passed in then the `_handle_electrodes_reading` will perform as expected.

adam2392:
Re scalp EEG:

Can just add an addition if/else statement here:

```
# Get coordinate frames that electrode coordinates are in
        if kind == "meg":
            coord_frame = coordsystem_json['MEGCoordinateSystem']
            coord_unit = coordsystem_json['MEGCoordinateUnits']
        elif kind == "ieeg":
            coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
            coord_unit = coordsystem_json['iEEGCoordinateUnits']

            # default coordinate frames to available ones in mne-python
            if coord_frame not in ['mri', 'ras', 'mni_tal',
                                   'mri_voxel', 'fs_tal']:
                warn("Defaulting coordinate frame to MRI "
                     "from coordinate system input {}".format(coord_frame))
                coord_frame = 'mri'
```

adam2392:
@jasmainak got nothing better to do... so I added the code back in for scalp EEG. I work w/ it only maybe 5% of the time, so not as well-versed in some of the intricacies there. The nice thing is that it seems that mne-python for scalp EEG can just assume a `head` coordinate frame.

jasmainak:
I think @sappelhoff might be able to help and review the scalp EEG code!

mne_bids/write.py
```diff
@@ -1279,6 +1301,20 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
                                   overwrite, verbose)
+        elif kind == 'eeg':
+            # We only write EEG electrodes.tsv and coordsystem.json
+            # if we have LPA, RPA, and NAS available to rescale to a known
+            # coordinate system frame
+            coords = _extract_landmarks(raw.info['dig'])
+            if set(['RPA', 'NAS', 'LPA']) == set(list(coords.keys())):
+                # Rescale to MNE-Python "head" coord system, which is the
+                # "ElektaNeuromag" system, and equivalent to "CapTrak"
+                pass
```
sappelhoff, line 1312: are you going to put the conversion code here?

what happens if NOT `set(['RPA', 'NAS', 'LPA']) == set(list(coords.keys()))`?

adam2392: idk... :p I copied this from your PR. Any recommendation here?

sappelhoff: okay, so when RPA, NAS and LPA are NOT present, do nothing but write a warning that we can only write EEG coords if we have those.

now for the case that we DO have these coords --> convert them using this handy MNE function:

```
In [1]: import mne                                                              

In [2]: ? mne.channels.montage.transform_to_head                                
Signature:  mne.channels.montage.transform_to_head(montage)
Docstring:
Transform a DigMontage object into head coordinate.

It requires that the LPA, RPA and Nasion fiducial
point are available. It requires that all fiducial
points are in the same coordinate e.g. 'unknown'
and it will convert all the point in this coordinate
system to Neuromag head coordinate system.

Parameters
----------
montage : instance of DigMontage
    The montage.

Returns
-------
montage : instance of DigMontage
    The montage after transforming the points to head
    coordinate system.
File:      ~/miniconda3/envs/sp/lib/python3.6/site-packages/mne/channels/montage.py
Type:      function

```

and then write them ... HOWEVER; please still write a log message that electrode positions are being written and that the user should double check that these are NOT template positions.

Template positions should be removed before writing to BIDS using some kind of `raw.montage.remove()` (don't know the actual method right now)

and then in code above that log message you make a comment `XXX: to be improved, see https://github.com/mne-tools/mne-bids/issues/264`

sappelhoff: and in any case also add a log message that coords are converted to CapTrak system (i.e., `head`)

adam2392:
@sappelhoff lmk what you think...

mne_bids/write.py
```diff
@@ -187,12 +190,60 @@ def _electrodes_tsv(raw, fname, kind, overwrite=False, verbose=True):
...
                             ])
-    else:
+    elif kind == 'eeg':
+        picks = pick_types(raw.info, eeg=True)
+        ch_locs = []
+        for ind, (name, _x, _y, _z) in enumerate(zip(names, x, y, z)):
+            if ind not in picks:
+                ch_locs.append([np.nan, np.nan, np.nan])
+            elif "n/a" in [_x, _y, _z]:
+                warn("{} electrode has n/a coordinate. "
```
sappelhoff, line 200: 1. did you fortget a `.format` for this warning?
1. can we use the logger to emit a warning?

mne_bids/write.py
```diff
@@ -187,12 +190,60 @@ def _electrodes_tsv(raw, fname, kind, overwrite=False, verbose=True):
...
+                z.append('n/a')
         data = OrderedDict([('name', names),
                             ('x', x),
                             ('y', y),
                             ('z', z),
                             ])
+
+        # log messages to warn user of pitfalls of this writing
+        logger.info("EEG coords are converted to "
+                    "CapTrak system (i.e., head).")
```
sappelhoff, line 238: I recommend this string: `Using anatomical landmarks (NAS, LPA, RPA) to transform EEG electrode coordinates to CapTrak coordinate system.`

mne_bids/write.py
```diff
@@ -187,12 +190,60 @@ def _electrodes_tsv(raw, fname, kind, overwrite=False, verbose=True):
...
+        logger.info("EEG Electrode positions are being written and "
+                    "the user should double check that "
+                    "these are NOT template positions. "
+                    "To remove the montage, one can use: "
+                    "`raw.set_montage(None)` ")
```
sappelhoff, line 243: I recommend this string:
```
Writing EEG electrode coordinates. Please ensure that these are NOT template positions, but truly measured data. To remove template positions, call `raw.set_montage(None)` prior to BIDS conversion.
```

mne_bids/write.py
sappelhoff: you can remove the `pass` from line 1364 @adam2392

mne_bids/write.py
```diff
@@ -1249,37 +1308,66 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
+        elif kind == 'eeg':
+            # We only write EEG electrodes.tsv and coordsystem.json
+            # if we have LPA, RPA, and NAS available to rescale to a known
+            # coordinate system frame
+            coords = _extract_landmarks(raw.info['dig'])
+            if set(['RPA', 'NAS', 'LPA']) != set(list(coords.keys())):
+                warn("Writing of EEG electrodes.tsv is skipped here... "
+                     "It only occurs if "
+                     "we have 'RPA', 'NAS', and 'LPA'. Please "
+                     "set these and refer to MNE-Python for more info.")
```
sappelhoff, line 1363: Again: Can we use `logger` to emit a warning?

Would recommend this string: `Writing EEG electrodes.tsv is only possible if anatomical landmarks (NAS, LPA, RPA) are present. Skipping ...`

mne_bids/write.py
```diff
@@ -1249,39 +1310,75 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
+        elif kind == 'eeg':
+            # We only write EEG electrodes.tsv and coordsystem.json
+            # if we have LPA, RPA, and NAS available to rescale to a known
+            # coordinate system frame
+            coords = _extract_landmarks(raw.info['dig'])
+            if set(['RPA', 'NAS', 'LPA']) != set(list(coords.keys())):  # noqa
+                logger.warning("Writing EEG electrodes.tsv "
+                               "is only possible if "
+                               "anatomical landmarks (NAS, LPA, RPA) "
+                               "are present. Skipping ...")
```
agramfort, line 1370: I don't know why this is necessary. For me EEG or iEEG requires the same logic. If you can write iEEG in MRI you should be able to do the same with EEG. You should only require fiducials to write in head coordinate system I think

adam2392: If that is the case, then I could wrap this in a check of the coord_frame for the raw.info['dig']?

@sappelhoff ?

sappelhoff: The issue is that in MNE we cannot be sure what kind of a digitization we get, because MNE does not save the coordinate system and units.

So the only way to be sure with what we work is to transform to head (and thus, automate without guessing) ... and for that, we need fiducials.

The alternative is to allow a user to manually input the coordsystem and units, but I prefer the solution here.

mne_bids/read.py
```diff
@@ -383,14 +394,32 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+            coord_unit = coordsystem_json['EEGCoordinateUnits']
+
+            if coord_frame.lower() not in ['captrak', 'besa']:
+                warn("Defaulting to 'head' coordinate frame for "
+                     "EEG montage because mne-bids does not "
+                     "support {} coordinate frame yet.".format(coord_frame))
+            coord_frame = 'head'
         elif kind == "ieeg":
-            coord_frame = coordsystem_json['iEEGCoordinateSystem']
+            coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
```
agramfort, line 408: why the .lower()? it's not done for EEG or MEG

mne_bids/read.py
```diff
@@ -383,14 +394,32 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+            coord_unit = coordsystem_json['iEEGCoordinateUnits']
+
+            # default coordinate frames to available ones in mne-python
+            if coord_frame not in _ieeg_coordinate_frames_dict.keys():
```
agramfort, line 412:
```suggestion
            if coord_frame not in _ieeg_coordinate_frames_dict:
```

agramfort: isn't it dangerous to convert to MRI when you don't know? To me this is bug prone and should rather raise an error

mne_bids/tests/test_read.py
```diff
@@ -295,19 +295,60 @@ def test_handle_coords_reading():
...
     # read in the data and assert montage is the same
+    # regardless of 'm', 'cm', 'mm', or 'pixel'
     bids_fname = bids_basename + "_ieeg.edf"
-    raw_test = read_raw_bids(bids_fname, bids_root)
-
-    # obtain the sensor positions
-    orig_locs = raw.info['dig'][1]
-    test_locs = raw_test.info['dig'][1]
-    assert orig_locs == test_locs
-    assert not object_diff(raw.info['chs'], raw_test.info['chs'])
```
agramfort, line 305: why killing an existing test? this suggests it was working before but you had to remove the test.

mne_bids/tests/test_read.py
```diff
@@ -295,19 +295,60 @@ def test_handle_coords_reading():
...
+            electrodes_dict[axis] = np.multiply(orig_electrodes_dict[axis],
+                                                scaling)
+        _to_tsv(electrodes_dict, electrodes_fname)
+
+        # read in raw file w/ updated montage
+        raw_test = read_raw_bids(bids_fname, bids_root)
+
+        # obtain the sensor positions and make sure they're the same
+        for origdig, newdig in zip(raw.info['dig'], raw_test.info['dig']):
+            np.testing.assert_array_equal(origdig, newdig)
```
agramfort, line 324: use object_diff you removed. A digpoint is not an array

agramfort:
MNE stores dig points locations in meters and every dig point can have it's
own coordinate system definition.

So I don't see why MNE data structures are limiting.

Now it's true that the set_montage is maybe too strict by forcing you to
have enough info to be in head coordinate

sappelhoff:
> MNE stores dig points locations in meters and every dig point can have it's
own coordinate system definition.
> So I don't see why MNE data structures are limiting.

But BIDS expects the electrodes to all be in a single coordinate system.

MNE is limiting only in terms of how we can extract metadata. With an MNE montage object, you cannot tell me for each digpoint, which coordinate system it is based on (apart from `head` and `unknown`)

agramfort:
>
> But BIDS expects the electrodes to all be in a single coordinate system.
>
> MNE is limiting only in terms of how we can extract metadata. With an MNE
> montage object, you cannot tell me for each digpoint, which coordinate
> system it is based on (apart from head and unknown)
>

technically the info['dig']  can contain anything. It's more a current
limitation of DigMontage object that was written to deal with EEG electrode
locations.

so MNE DigMontage can be improved / extended to deal with your use cases
rather than doing hacks in mne-bids with some guessings and warnings

we can talk IRL if this can help

my 2c

sappelhoff:
> technically the info['dig']  can contain anything. It's more a current
limitation of DigMontage object that was written to deal with EEG electrode
locations.

Information that I would like to have in info['dig']:

1. a flag `"template_posititon" : bool` (if True, the positions are simulated/template positions (i.e., NOT truly measured) ... we can set this to True, when the montage was created using [`make_standard_montage`](https://mne.tools/stable/generated/mne.channels.make_standard_montage.html?highlight=montage#mne.channels.make_standard_montage) ... else this is FALSE

--> then in MNE-BIDS, we only write positions, if the `template_positions` flag is `False`

2. a variable `"coord_system" : str`, which is `mixed` as soon as the dig points have DIFFERENT coord_systems ... but if it is not `mixed`, it should be set according to the function that was used to read the dig points:

- CapTrak --> [`read_dig_captrack`](https://mne.tools/stable/generated/mne.channels.read_dig_captrack.html#mne.channels.read_dig_captrack)
- EGI --> [`read_dig_egi`](https://mne.tools/stable/generated/mne.channels.read_dig_egi.html#mne.channels.read_dig_egi)
- ... and so on

---

For the sake of this PR, I think it's an improvement that users can write electrode positions if they have LPA, RPA, and NAS ... as opposed to not being able to write them at all.

I don't have time currently to dive into this apart from giving advice of what I think is needed.

agramfort:
what seems possible to keep track of what is template location and what is
not is to use
the coil_type in chs data struct. It's 1 for EEG but we could define a new
one for EEG that
is a template location.

wrt to storing this info in dig we can use the "kind" key. in DigPoint to
support template location.

@larsoner suggested to get rid of coord_system parameter in DigMontage as
it's
an info present in the dig points now.

I can find some time but to help me I need code snippet / tests that I can
use to see if I get the use case correctly.

mne_bids/read.py
adam2392: I think the mne-python coordinate frames are all lower case, so I'll adjust for EEG and MEG.

mne_bids/read.py
adam2392: So while that is true, iEEG only makes sense really if it is in the MRI coordinate frame. All derivatives of this, such as acpc, ras, etc. are still from the MRI image of the subject brain. I think a warning could suffice here?

If not, I would say it might be better to defer to 'unknown'? Since electrode coordinates can be present, and maybe the coordinatesystem is something like `LIA`, which isn't supported in mne-python yet, so the user would still at least have their montage set.

sappelhoff:
Okay, does this help?

```Python
import os

import mne

# Get an actual dig montage (measured)
data_path = mne.datasets.testing.data_path()
fname_montage = os.path.join(data_path, 'montage', 'captrak_coords.bvct')
true_montage = mne.channels.read_dig_captrack(fname_montage)

# Make a template montage
template_montage = mne.channels.make_standard_montage(kind='standard_1020') 

# would be nice to have
assert true_montage.is_template == False
assert template_montage.is_template == True  # because used make_standard_montage

# would be nice to have
for dig in true_montage.dig:
    assert dig['coord_fname'] == 'captrak'  # because read with `read_dig_captrack`

for dig in template_montage.dig:
    assert dig['coord_fname'] == 'unknown'  # or 'head' ...

```

while doing this, I actually also found a bug.

Try: `true_montage.plot()`

sappelhoff:
btw, related: https://github.com/mne-tools/mne-bids/issues/264

mne_bids/write.py
```diff
@@ -1249,39 +1263,111 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
-            # Rescale to MNE-Python "head" coord system, which is the
-            # "ElektaNeuromag" system (equivalent to "CapTrak" system)
+            # RPA, NAS and LPA shouldn't be set in iEEG
             if set(['RPA', 'NAS', 'LPA']) != set(list(coords.keys())):
                 # Now write the data to the elec coords and the coordsystem
                 _electrodes_tsv(raw, electrodes_fname,
                                 kind, overwrite, verbose)
                 _coordsystem_json(raw, unit, orient,
                                   coord_frame, coordsystem_fname, kind,
                                   overwrite, verbose)
```
hoechenberger, line 1308: I don't understand this. So we _only_ write if not all fiducials are present in the data? But we _do_ write if, say, only `LPA` and `RPA` are present (and `NAS` is missing)? Why can we not write regardless of whether fiducials are present in the data or not? If they don't make sense or are not required for iEEG data, they can be just omitted. In either case we should raise if we're not writing data, no?

mne_bids/write.py
```diff
@@ -1249,39 +1263,111 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
-        elif kind != "meg":
-            warn('Writing of electrodes.tsv is not supported for kind "{}". '
-                 'Skipping ...'.format(kind))
+            else:  # noqa
+                logger.warning("Scalp EEG anatomical landmarks "
+                               "(NAS, LPA, RPA) "
+                               "shouldn't be set in iEEG. "
+                               "Skipping montage setting...")
```
hoechenberger, line 1313: See my comment above

mne_bids/read.py
```diff
@@ -383,14 +394,32 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+            coord_unit = coordsystem_json['MEGCoordinateUnits']
+        elif kind == 'eeg':
+            coord_frame = coordsystem_json['EEGCoordinateSystem']
+            coord_unit = coordsystem_json['EEGCoordinateUnits']
+
+            if coord_frame.lower() not in ['captrak', 'besa']:
+                warn("Defaulting to 'head' coordinate frame for "
+                     "EEG montage because mne-bids does not "
+                     "support {} coordinate frame yet.".format(coord_frame))
+            coord_frame = 'head'
```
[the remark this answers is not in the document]
jasmainak: @adam2392 no guessing anywhere. A hard-fail would be problematic since it would make the entire file structure unreadable. Perhaps skip setting the montage with a warning message for the user to set the coordinate system correctly.

mne_bids/tests/test_read.py
```diff
@@ -295,19 +295,60 @@ def test_handle_coords_reading():
...
+
+    # check that coordinate systems can be used and defaults to mri
+    coordinate_frames = ['lia', 'ria', 'lip', 'rip', 'las']
```
jasmainak, line 327: can you add a link to the documentation of these coordinate systems (from BIDS specification?) in the comments where you define them?

jasmainak:
@agramfort @hoechenberger @sappelhoff reading all the discussion makes me feel that we need a dedicated function to write the dig points. Something like `write_dig_bids` which exposes parameters like `coordinate_system` and `units`. There seem to be too many exceptions to deal with otherwise. Thoughts?

sappelhoff:
> we need a dedicated function to write the dig points. Something like write_dig_bids which exposes parameters like coordinate_system and units. There seem to be too many exceptions to deal with otherwise. Thoughts?

I agree that writing dig points is very hard to get right ... and we have the additional goal to make this as automatic as possible.

For EEG, the only way I (personally) would feel safe about automatically writing dig points, is if we internally rescale the coordsystem using NAS, LPA, RPA, so that we know **exactly** what we have.

The alternative is what you (@jasmainak ) suggest: exposing parameters --> but then it wouldn't be automatic anymore :man_shrugging:

hoechenberger:
> Something like `write_dig_bids` which exposes parameters like `coordinate_system` and `units`. There seem to be too many exceptions to deal with otherwise. Thoughts?

+1 on this one :)

mne_bids/read.py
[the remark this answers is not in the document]
adam2392: Okay will skip by setting coord_frame to None then.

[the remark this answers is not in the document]
adam2392: I was using `head` because @sappelhoff initially suggested it, but it seems based on discussion that EEG might be a lot more complicated.

mne_bids/write.py
```diff
@@ -186,12 +190,15 @@ def _electrodes_tsv(raw, fname, kind, overwrite=False, verbose=True):
...
                             ('size', sizes),
                             ])
-    else:
+    elif kind == 'eeg':
         data = OrderedDict([('name', names),
                             ('x', x),
                             ('y', y),
                             ('z', z),
                             ])
+    else:  # noqa
```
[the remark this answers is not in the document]
adam2392: Oh yeah oops, put it to `pragma: no cover` since idk if its necessary to test... just there to make sure an informative message gets passed back if a 'kind' is incorrectly passed.

mne_bids/write.py
adam2392: Sure that's fair, I left this in from awhile back in a different PR. I agree it really shouldn't matter if LPA, or RPA, NAS are there or not for iEEG.

mne_bids/write.py
```diff
@@ -1249,39 +1263,111 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
     if raw.info['dig'] is not None:
         if kind == "ieeg":
+            # get coordinate frame from digMontage
+            digpoint = raw.info['dig'][0]
+            if any(digpoint['coord_frame'] != _digpoint['coord_frame']
+                   for _digpoint in raw.info['dig']):
+                warn("Defaulting coordinate frame of iEEG coords to "
+                     "'mri' since not all digpoints "
+                     "have the same coordinate frame.")
+                coord_frame = "mri"  # defaults to MRI coordinates
```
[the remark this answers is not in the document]
adam2392: Instead of hard-failing, I produce warnings now. Lmk what you think. @hoechenberger

adam2392:
@agramfort @jasmainak @sappelhoff 

**Re Scalp EEG:**
So based on the discussion, I'll admit I'm a bit confused what to do w/ the EEG `write` and `read` functions. Is it safe to assume captrak/besa can be read in as `head` coordinate frame? Not sure what to do w/ write.

**Re iEEG:**
@hoechenberger to answer some of your questions above: For iEEG, the bids specification allows coordinate frames such as 'LIA', etc. See: https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html. However, mne-python doesn't accept these types yet, but these are all essentially derived from an 'mri' image because that is the only accepted way to get the iEEG coordinates initially for a patient.  

So I suppose from your comments, we can either:
1. default to 'mri' in the case of iEEG, or
2. also warning message and don't set, unless they first convert all their coordinates to something accepted by mne-python. 

For me I am leaning towards 1 because it is a huge hassle for me to have to do this transformation when things like "LIA" are acceptable in BIDS, but tbh I can also see the possible issues that arise w/ 1 compared to 2.

sappelhoff:
**EEG**

> Is it safe to assume captrak/besa can be read in as head coordinate frame? Not sure what to do w/ write.

READING:

captrak is the same coord system as `head`, but the units may differ, so there needs to be a check of the units field and a potential scaling to `meters`. For `besa`, I am not sure ... it's probably an error to have it in the spec at all: https://github.com/bids-standard/bids-specification/issues/420

WRITING:

I like where we currently stand. Write to Captrak if we have NAS, RPA, LPA ... or do not write at all! (future PRs can improve this with a function where users have to explicitly pass params ... although that would not be automatic anymore)

jasmainak:
@sappelhoff can you explaint to me why you need the landmark locations to write the electrode coordinates? The landmarks are in what coordinate system?

sappelhoff:
> @sappelhoff can you explaint to me why you need the landmark locations to write the electrode coordinates? The landmarks are in what coordinate system?

1. the landmark coordinates and all EEG electrode coordinates MUST be in the same coordinate system
1. then we can use the landmarks (NAS, RPA, LPA) to rescale all coordinates to the captrak (same as `head`) coordinate system
1. doing this allows us to know exactly what we are dealing with

that said, we could also trust MNE when it says that coordinates are in `head` system, and then we can write without double checking using landmarks + rescaling

hoechenberger:
> that said, we could also trust MNE when it says that coordinates are in `head` system, and then we can write without double checking using landmarks + rescaling

That sounds like the sane thing to do IMHO!

sappelhoff:
> That sounds like the sane thing to do IMHO!

right, now that I have written it all out, I could agree with this policy:

1. when coords are in `head` we are very happy and write as `captrak`
1. when coords are not in `head`, we try to find LPA, RPA, NAS and scale to `head`
    1. if this works out, we send a log message that we did a rescaling to `head` and we are happy
1. when coords are not in `head` AND we cannot rescale, we write electrodes as they are, and enter as `EEGCoordinateSystem` "other" ... then we send a warning that the user MUST specify `EEGCoordinateSystemDescription` manually, because otherwise the positions cannot be properly interpreted

If we implement it this way, only one issue remains for me:

--> how do we make sure that users don't write template electrode positions to BIDS?

I would find it bad if users did something with `mne.channels.make_standard_montage(kind='standard_1020') ` in their code and then write that data to BIDS, see #264

agramfort:
there should be some overlap beyond "head' between the coordinate systems MNE understands
(really it's freesurfer) and what is part of BIDS standard. I would really imagine that you have locations
in MRI coords without fiducials for sEEG or ECoG of even EEG/fMRI and that it should just
work with MNE-bids without having to define fiducials (at least for certain operations).

adam2392:
@agramfort is `object_diff` the best way to compare dig dictionaries here? I think it's not flexible for machine precision when I multiply the SI units in the test.

I keep on getting this error:
```
    
            # obtain the sensor positions and make sure they're the same
            print(coord_unit, scaling, len(raw.info['dig']), len(raw_test.info['dig']))
            for origdig, newdig in zip(raw.info['dig'], raw_test.info['dig']):
                # print(origdig, newdig)
                # print(origdig['r'], newdig['r'])
                for key in origdig:
                    print(type(origdig[key]), type(newdig[key]))
                    if key == 'r':
                        print("R: ", [type(x) for x in origdig[key]], [type(x) for x in newdig[key]])
                    print("Actual values: ", origdig[key], newdig[key])
>                   assert not object_diff(origdig[key], newdig[key])
E                   AssertionError: assert not ' array mismatch\n'
E                    +  where ' array mismatch\n' = object_diff(array([0.91347764, 0.03469434, 0.74704786]), array([0.91347764, 0.03469434, 0.74704786]))
```
But I checked type, and everything is exactly the same up to machine precision.

```
>                   np.testing.assert_array_equal(origdig[key], newdig[key])
E                   AssertionError: 
E                   Arrays are not equal
E                   
E                   Mismatch: 33.3%
E                   Max absolute difference: 1.38777878e-17
E                   Max relative difference: 1.4971148e-16
E                    x: array([0.597744, 0.680037, 0.092697])
E                    y: array([0.597744, 0.680037, 0.092697])
```

I updated the unit test to do a work-around: mne_bids/tests/test_read.py:363

mne_bids/tests/test_read.py
adam2392: See https://github.com/mne-tools/mne-bids/pull/390#issuecomment-619102739

mne_bids/read.py
adam2392: For EEG, I believe we've prelim addressed it w/ the discussion down below.

For iEEG: https://github.com/mne-tools/mne-bids/pull/390#issuecomment-618462528

adam2392:
For some reason mne-python converts any montage it has into head if unknown?

mne.channels.montage._set_montage

E.g this fails:
```
    # # create montage in 'unknown' coordinate frame
    # # and assert coordsystem/electrodes sidecar tsv don't exist
    montage = mne.channels.make_dig_montage(ch_pos=ch_pos,
                                            coord_frame="unknown")
    raw.set_montage(montage)
    with pytest.warns(UserWarning, match="Skipping EEG electrodes.tsv"):
        coordsystem_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                   suffix='coordsystem.json',
                                                   allow_fail=True)
        electrodes_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                  suffix="electrodes.tsv",
                                                  allow_fail=True)
        assert coordsystem_fname is None
        assert electrodes_fname is None
```

but the current one does not. I believe this is because there is this line inside `set_montage`: 
`mnt_head = _get_montage_in_head(montage)`

agramfort:
yes mne-python is certainly too limited at this stage for the use case of channels in non-head coord frame. It should be usable when you have channes in MRI for example.

cc @hoechenberger 

we do have data that pass the BIDS validator that is not properly handled by mne-bids. It's easier for me to start from actual data to understand the limitation. If not please do like we did in the past. Give me on dropbox a mock dataset.

adam2392:
> yes mne-python is certainly too limited at this stage for the use case of channels in non-head coord frame. It should be usable when you have channes in MRI for example.
> 
> cc @hoechenberger
> 
> we do have data that pass the BIDS validator that is not properly handled by mne-bids. It's easier for me to start from actual data to understand the limitation. If not please do like we did in the past. Give me on dropbox a mock dataset.

```
bids_root = _TempDir()

    data_path = op.join(testing.data_path(), 'EDF')
    raw_fname = op.join(data_path, 'test_reduced.edf')
    bids_fname = bids_basename + "_eeg.edf"
    raw = mne.io.read_raw_edf(raw_fname)

    # ensure we are writing 'eeg' data
    raw.set_channel_types({ch: 'eeg'
                           for ch in raw.ch_names})

    # set a `random` montage
    ch_names = raw.ch_names
    elec_locs = np.random.random((len(ch_names), 3)).astype(float)
    ch_pos = dict(zip(ch_names, elec_locs))

    # # create montage in 'unknown' coordinate frame
    # # and assert coordsystem/electrodes sidecar tsv don't exist
    montage = mne.channels.make_dig_montage(ch_pos=ch_pos,
                                            coord_frame="unknown")
    raw.set_montage(montage)
    with pytest.warns(UserWarning, match="Skipping EEG electrodes.tsv"):
        coordsystem_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                   suffix='coordsystem.json',
                                                   allow_fail=True)
        electrodes_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                  suffix="electrodes.tsv",
                                                  allow_fail=True)
        assert coordsystem_fname is None
        assert electrodes_fname is None
```

This will fail explicitly, and so there is no way to make this work as we want rn. As a result, I think the EEG framework that @sappelhoff mentioned makes sense, but I don't think mne-python is quite there yet to handle the cases:

- `set_montage` shouldn't transform unknown to head(?) or if it should for a lot of use cases, perhaps there should be some metadata attached to make sure there is some way of knowing that that was done w/, or w/o NAS/LPA/RPA. 

Whereas for iEEG I feel like the coordinates can be relatively constrained into 'mri' and the corresponding derivatives of such. I took out the EEG sections for write, read and test for now. Should we table that? I've attached the files here related to the code segments.

[files.zip](https://github.com/mne-tools/mne-bids/files/4531728/files.zip)

mne_bids/tests/test_read.py
```diff
@@ -327,20 +328,74 @@ def test_handle_coords_reading():
...
+        for axis in ['x', 'y', 'z']:
+            electrodes_dict[axis] = np.multiply(orig_electrodes_dict[axis],
+                                                scaling)
+        _to_tsv(electrodes_dict, electrodes_fname)
+
+        # read in raw file w/ updated montage
+        raw_test = read_raw_bids(bids_fname, bids_root)
+
+        # obtain the sensor positions and make sure they're the same
+        for origdig, newdig in zip(raw.info['dig'], raw_test.info['dig']):
```
jasmainak, line 363:
```suggestion
        for orig_dig, new_dig in zip(raw.info['dig'], raw_test.info['dig']):
```

it's a bit more readable with the underscore. Less chance of confusion in the future

mne_bids/tests/test_read.py
```diff
@@ -327,20 +328,74 @@ def test_handle_coords_reading():
...
+            for key in origdig:
+                if key == 'r':
+                    assert_almost_equal(origdig[key], newdig[key])
+                else:
+                    assert not object_diff(origdig[key], newdig[key])
```
jasmainak, line 368: I don't understand what's happening here. Can you explain?

mne_bids/tests/test_read.py
```diff
@@ -327,20 +328,74 @@ def test_handle_coords_reading():
...
+                                     [str, float, float, float, str])
+    for coord_unit, scaling in zip(coordinate_units, scaling_numbers):
+        # update coordinate SI units
+        _update_sidecar(coordsystem_fname, 'iEEGCoordinateUnits', coord_unit)
+        electrodes_dict = _from_tsv(electrodes_fname,
+                                    [str, float, float, float, str])
+        for axis in ['x', 'y', 'z']:
+            electrodes_dict[axis] = np.multiply(orig_electrodes_dict[axis],
+                                                scaling)
+        _to_tsv(electrodes_dict, electrodes_fname)
```
jasmainak, line 357: I would question this test. If there is not a public interface for this to happen, why should we test this?

mne_bids/utils.py
```diff
@@ -29,6 +29,18 @@
 
 from mne_bids.tsv_handler import _to_tsv, _tsv_to_str
 
+# these coordinate frames in mne-python are related to scalp/meg
+# 'meg', 'ctf_head', 'ctf_meg', 'head', 'unknown'
+_ieeg_coordinate_frames_dict = dict(
```
jasmainak, line 34: shouldn't this be in caps because it's a global?

mne_bids/read.py
```diff
@@ -415,17 +429,30 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+            coord_unit = coordsystem_json['MEGCoordinateUnits']
         elif kind == "ieeg":
-            coord_frame = coordsystem_json['iEEGCoordinateSystem']
-        else:
+            coord_frame = coordsystem_json['iEEGCoordinateSystem'].lower()
+            coord_unit = coordsystem_json['iEEGCoordinateUnits']
+
+            # default coordinate frames to available ones in mne-python
+            # noqa: see https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html
+            if coord_frame not in _ieeg_coordinate_frames_dict:
```
jasmainak, line 442: I'm a bit confused. There are only two coordinate systems allowed in the BIDS specification (in the link above) -- Pixels and ACPC. Why does your dictionary not contain them? Furthermore, I see that there are some coordinate systems that are common to MEG/EEG/iEEG and some that are are separate for each. Finally, you should put `_ieeg_coordinate_frames_dict` in the `config.py` file like the rest of these variables.

mne_bids/read.py
```diff
@@ -415,17 +429,30 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+                coord_frame = 'mri'
+        else:  # noqa
             raise RuntimeError("Kind {} not supported yet for "
                                "coordsystem.json and "
                                "electrodes.tsv.".format(kind))
-        # read in electrode coordinates and attach to raw
-        raw = _handle_electrodes_reading(electrodes_fname, coord_frame, raw,
-                                         verbose)
+
+        if coord_frame is not None:
```
jasmainak, line 451: what about MEG? Should you pass this down also for MEG data?

mne_bids/tests/test_write.py
```diff
@@ -664,7 +664,7 @@ def test_vhdr(_bids_validate):
 
 
 @pytest.mark.filterwarnings(warning_str['nasion_not_found'])
-def test_edf(_bids_validate):
+def test_edf(_bids_validate, caplog):
```
jasmainak, line 667: what's caplog?

agramfort:
@adam2392 can we do like last time. You put the self contained script with data in our share dropbox folder? I've put there the files you sent me. It will help me to look into it. thanks

mne_bids/tests/test_read.py
adam2392: See: https://github.com/mne-tools/mne-bids/pull/390#issuecomment-619102739

when doing the test by modifying the "CoordinateUnits" of the electrodes.tsv data, then the data is definitely not the same up to machine precision.

mne_bids/tests/test_read.py
adam2392: When you write an electrodes.tsv file outside the scope of MNE-bids, don't we want to support accepted BIDS behavior? Especially if the accepted units are 'mm', 'm', and 'cm', and not just 'm', which is the default in mne-python?

mne_bids/read.py
adam2392: MEG doesn't have an electrodes.tsv 

See: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html

mne_bids/tests/test_write.py
adam2392: a pytest fixture for capturing log output, but this can be removed now. Sorry forgot to take this out.

adam2392:
> @adam2392 can we do like last time. You put the self contained script with data in our share dropbox folder? I've put there the files you sent me. It will help me to look into it. thanks

@agramfort sorry maybe the purpose of this PR got lost in the convos. The original problems I had were:

1. **Reading in Correct Scale of Electrode Coordinates** that data (see [example](https://www.dropbox.com/sh/lkz1q6do80w53sv/AAAoPAKQJi3xCgJ2y5GGeREsa?dl=0)) I have for iEEG was not able to be read in correctly. Specifically, the `electrodes.tsv` coordinates were in 'mm', whereas mne-bids assumes 'm' by default. However, [BIDS specification](https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html) specifies we don't need to have 'm' necessarily. 

This is currently solved by:
-> the modification to `_handle_electrodes_reading` with `coord_unit` and corresponding unit test

2. **Reading Arbitrary coordinate frames for iEEG**
I had electrode coordinates not in ACPC, or a standard template format, which resulted in an error in reading in the file. For example, my "CoordinateSystem" was "LIA", as opposed to say... "RAS", or "ACPC", or even "MRI". From discussion above, and reading the [specification](https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html), these should be encapsulated in the keyword "Other" actually and a corresponding "CoordinateSystemDescription" keyword, and "space" BIDS entity. For example:

The files would be:
- sub-01_ses-test_acq-seeg_space-lia_electrodes.tsv
- sub-01_ses-test_acq-seeg_space-lia_coordsystem.json
 
The coordsystem.json would have:

```
'iEEGCoordinateSystem': coordsystem_name,  # MRI, Pixels, or ACPC
 'iEEGCoordinateUnits': unit,  # m (MNE), mm, cm , or pixels
 'iEEGCoordinateSystemDescription': coordsystem_description, # some description of the coordinate system
```

The descriptions can be taken from the specification itself (note that for us, we'll want to map BIDS coordinate systems back and forth from mne-python coordinate system keywords).

-> a temporary fix is to set `coord_frame = None` when we can't read in the coordinate frame

**In summary for this PR:**
1. I believe is the simplest addition, and it would be helpful to have this merged in since it's an easy fix on the `read_raw_bids` side with corresponding test.
2. temporary fix of just not reading when coordinate frame is not matching

**A longer term fix**
I'm seeing discussion of a `_write_dig_bids`/`_read_dig_bids` and I think this actually makes a lot of sense for both reading and writing since there are so many incongruities in BIDS coordinate systems vs what it is named in mne-python + a lot of nuanced cases. I am in favor of moving this into a different PR. I can make an issue to start this discussion.

mne_bids/tests/test_read.py
```diff
@@ -327,20 +330,64 @@ def test_handle_coords_reading():
...
+                np.multiply(orig_electrodes_dict[axis], scaling)
+        _to_tsv(electrodes_dict, electrodes_fname)
+
+        # read in raw file w/ updated montage
+        raw_test = read_raw_bids(bids_fname, bids_root)
+
+        # obtain the sensor positions and make sure they're the same
+        assert_dig_allclose(raw.info, raw_test.info)
+
+    # check that coordinate systems can be used and defaults to mri
```
agramfort, line 366: this part of the test now fails after my last commit. Please have a look. Than I think it's good enough

adam2392: In terms of setting the coordinate frame, this line in mne-python prevents this from setting it as 'unknown':

mne-python/mne/channels/montage.py:664
>`mnt_head = _get_montage_in_head(montage)`

Inside function: `mne.channels.montage._set_montage`. As a result, `raw.set_montage(montage)` now has all coordinate frames in `head` frame. Idk what's the proper fix in mne-python, but I put up a fix to the unit test for mne-bids for now. Lmk if this is okay.

mne_bids/read.py
```diff
@@ -415,17 +426,31 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
...
+            coord_unit = coordsystem_json['iEEGCoordinateUnits']
+        else:  # noqa
+            # XXX should add support of coordsystem.json for EEG
             raise RuntimeError("Kind {} not supported yet for "
                                "coordsystem.json and "
                                "electrodes.tsv.".format(kind))
+
+        coord_frame = coord_frame.lower()  # MNE uses lower case
+        if kind == 'ieeg':
+            # XXX: improve reading with a `_write_dig_bids`
```
jasmainak, line 444: reading with a `_write_dig_bids`?

mne_bids/write.py
```diff
@@ -187,14 +189,17 @@ def _electrodes_tsv(raw, fname, kind, overwrite=False, verbose=True):
...
                             ('x', x),
                             ('y', y),
                             ('z', z),
                             ])
+    else:  # pragma: no cover
+        raise RuntimeError("kind {} not supported.".format(kind))
+
     _write_tsv(fname, data, overwrite=overwrite, verbose=verbose)
-    return fname
+    return fname  # XXX why return parameter???
```
jasmainak, line 202: I think this should be deleted @agramfort

mne_bids/tests/test_read.py
```diff
@@ -327,20 +330,69 @@ def test_handle_coords_reading():
...
+        # and make sure all digpoints are MRI coordinate frame
+        with pytest.warns(UserWarning, match="Coordinate frame"):
+            raw_test = read_raw_bids(bids_fname, bids_root)
+            # XXX: add test for 'unknown' possibly
+            for digpoint in raw_test.info['dig']:
+                assert digpoint['coord_frame'] in [unknown_coord_frame,
+                                                   head_coord_frame]
+
+    # coordinate frames in mne-python should all map correctly
+    coordinate_frames = ['mri', 'mri_voxel', 'mni_tal', 'ras', 'fs_tal']
```
jasmainak, line 383: do you write the MNE-Python coordinate frame to BIDS?

They seem to have a different set of names: https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html no?

mne_bids/read.py
agramfort: means that this code block should be moved into a dedicated private function _write_dig_bids

mne_bids/tests/test_read.py
agramfort: we should

mne_bids/read.py
```diff
@@ -249,8 +256,12 @@ def _float_or_nan(val):
         warn("There are channels without locations "
              "(n/a) that are not marked as bad: {}".format(nan_chs))
 
+    # convert coordinates to meters if necessary
+    if coord_unit in ['m', 'cm', 'mm']:
+        ch_locs = _scale_coord_to_meters(ch_locs, coord_unit)
```
agramfort, line 261: we should raise an error otherwise. Do we check that units are among these always?

mne_bids/write.py
agramfort: you agree that there is no point in returning fname?
_electrodes_tsv does the same and I have no clue why

doc/whats_new.rst
```diff
@@ -29,6 +29,7 @@ Bug
 
 - fix bug in :func:`write_raw_bids` where raw.info['subject_info'] can be None, by `Adam Li`_ (`#392 <https://github.com/mne-tools/mne-bids/pull/392>`_)
 - :func:`read_raw_bids` will now read all channels from electrodes.tsv. Channels with coordinates 'n/a' will also be included but their location set to `np.nan` in the `raw` object, by `Adam Li`_ (`#393 <https://github.com/mne-tools/mne-bids/pull/393>`_)
+- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinateframes by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
```
agramfort, line 32:
```suggestion
- fix :func:`mne_bids.read_raw_bids` to correctly scale coordinate to meters in electrodes.tsv and also read in possible iEEG coordinate frames by `Adam Li`_ (`#390 <https://github.com/mne-tools/mne-bids/pull/390>`_)
```

mne_bids/write.py
jasmainak: yes agreed

adam2392: I think that's the same for all the _sidecar writing functions then?

agramfort: @sappelhoff any idea why this is like that?

mne_bids/read.py
adam2392: I now do it outside the private function: 

https://github.com/mne-tools/mne-bids/blob/306e072dc4625427e2a1a65bc75443118f31eb76/mne_bids/read.py#L463-L468

mne_bids/tests/test_read.py
adam2392: This now done via the `space` entity and a combo w/ `XXXCoordinateSystem = 'Other'`, which imo... is not the best way to do it, but it's what I'm getting from the BIDS specification.

mne_bids/read.py
```diff
@@ -415,6 +415,8 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
     coordsystem_fname = _find_matching_sidecar(bids_fname, bids_root,
                                                'coordsystem.json',
                                                allow_fail=True)
+    print("FOUND ELECTRODES FNAME: ", electrodes_fname)
```
agramfort, line 418: debug to remove?

mne_bids/write.py
adam2392: Deleted in _electrodes and _coordsystem

mne_bids/read.py
adam2392: > reading with a `_write_dig_bids`?

I'm happy to try to help out w/ this later (maybe next week), but for the purposes of this PR, I just want to solve the "scaling" in units and the arbitrary coordinate systems for iEEG.

adam2392:
Couple of notes and comments that don't seem trivial to currently address:

1. https://github.com/mne-tools/mne-bids/pull/390/files/6711bf1484fd638b9b78f5f551d3b962b4b100d7#r416899201
@agramfort wanted a unit-test to pass that allows 'unknown' coordframe. Currently this is impossible since these montages are automatically transformed to 'head'. Pending solving this issue: https://github.com/mne-tools/mne-python/issues/7691

we can allow 'unknown' coordinate frames (for EEG and iEEG!)

2. The way to handle arbitrary coordinate frames for iEEG at least according to the specification is using the `space` entity, which seems buggy. So probably the write/read is not fully robust yet, but this LGTM for now.

3. Removed all the return of `fname` for sidecar file writing functions in `write.py`

mne_bids/write.py
sappelhoff: I think it was convenient at some point? But I also don't remember :shrug:

adam2392:
I see the doc build failed.

I rebuilt the example locally and it works. Is this possibly due to the datasets not being updated correctly.

agramfort:
@adam2392 can you look at circleci issue?

agramfort:
any idea what made it fail?

the error seems related

>

adam2392:
> @adam2392 can you look at circleci issue?

This is  what I got locally.

```
(mne) adam2392@Adams-MBP-3 doc % git status
gOn branch electrodes
Your branch is up to date with 'dev/electrodes'.

nothing to commit, working tree clean
(mne) adam2392@Adams-MBP-3 doc % make html
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v2.3.1
loading pickled environment... done
[autosummary] generating autosummary for: api.rst, auto_examples/convert_eeg_to_bids.rst, auto_examples/convert_empty_room.rst, auto_examples/convert_group_studies.rst, auto_examples/convert_ieeg_to_bids.rst, auto_examples/convert_mne_sample.rst, auto_examples/convert_mri_and_trans.rst, auto_examples/create_bids_folder.rst, auto_examples/index.rst, auto_examples/read_bids_datasets.rst, ..., generated/mne_bids.make_bids_folders.rst, generated/mne_bids.make_dataset_description.rst, generated/mne_bids.read_raw_bids.rst, generated/mne_bids.utils.get_entity_vals.rst, generated/mne_bids.utils.get_kinds.rst, generated/mne_bids.utils.print_dir_tree.rst, generated/mne_bids.write_anat.rst, generated/mne_bids.write_raw_bids.rst, index.rst, whats_new.rst
generating gallery...
generating gallery for auto_examples... [100%] convert_empty_room.py                                             
computation time summary:
    - ../examples/convert_eeg_to_bids.py:        0.00 sec   0.0 MB
    - ../examples/convert_empty_room.py:         0.00 sec   0.0 MB
    - ../examples/convert_group_studies.py:      0.00 sec   0.0 MB
    - ../examples/convert_ieeg_to_bids.py:       0.00 sec   0.0 MB
    - ../examples/convert_mne_sample.py:         0.00 sec   0.0 MB
    - ../examples/convert_mri_and_trans.py:      0.00 sec   0.0 MB
    - ../examples/create_bids_folder.py:         0.00 sec   0.0 MB
    - ../examples/read_bids_datasets.py:         0.00 sec   0.0 MB
    - ../examples/rename_brainvision_files.py:   0.00 sec   0.0 MB
generating MNE-BIDS cli help ... [100%] mne_bids_raw_to_bids.py                                                  
[Done]
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: [config changed ('numpydoc_xref_aliases')] 34 added, 0 changed, 0 removed
reading sources... [100%] whats_new                                                                              
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] whats_new                                                                               
generating indices...  genindexdone
highlighting module code... [100%] mne_bids.write                                                                
writing additional pages...  searchdone
copying images... [100%] auto_examples/images/thumb/sphx_glr_convert_empty_room_thumb.png                        
copying downloadable files... [100%] auto_examples/rename_brainvision_files.ipynb                                
copying static files... ... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.
copying binder requirements...
copying binder notebooks...[100%] auto_examples                                                                  

Sphinx-gallery successfully executed 0 out of 0 files subselected by:

    gallery_conf["filename_pattern"] = '^((?!sgskip).)*$'
    gallery_conf["ignore_pattern"]   = '__init__\\.py'

after excluding 9 files that had previously been run (based on MD5).

embedding documentation hyperlinks...
embedding documentation hyperlinks for auto_examples... [100%] convert_eeg_to_bids.html                          

Build finished. The HTML pages are in _build/html.
```

jasmainak:
@adam2392 do you have the same MNE version as CircleCI?

adam2392:
> @adam2392 do you have the same MNE version as CircleCI?

```
>>> import mne
>>> mne.sys_info()
Platform:      Darwin-19.4.0-x86_64-i386-64bit
Python:        3.6.7 | packaged by conda-forge | (default, Nov  6 2019, 16:03:31)  [GCC Clang 9.0.0 (tags/RELEASE_900/final)]
Executable:    /Users/adam2392/miniconda3/envs/mne/bin/python
CPU:           i386: 8 cores
Memory:        Unavailable (requires "psutil" package)
mne:           0.21.dev0
numpy:         1.17.3 {blas=blas, lapack=blas}
scipy:         1.3.3
matplotlib:    3.1.2 {backend=TkAgg}

sklearn:       0.22
numba:         Not found
nibabel:       2.5.1
cupy:          Not found
pandas:        1.0.3
dipy:          Not found
mayavi:        4.7.1
pyvista:       Not found
vtk:           8.2.0
PyQt5:         5.12.3
```

agramfort:
I cannot replicate locally but we definitely broke something ....

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
               conda update --yes --quiet conda
               conda env create -f environment.yml --name=testenv
               source activate testenv
-              pip install mne==0.20
+              pip install mne
```
adam2392, line 42:
```suggestion
              pip install mne
```
@agramfort ?

adam2392:
@jasmainak rebased, lmk if there's any other changes you want.

I can work on the refactoring of the read/write of electrodes coords in another PR to fix the issue: #398

mne_bids/read.py
```diff
@@ -344,6 +356,11 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
     extra_params : None | dict
         Extra parameters to be passed to MNE read_raw_* functions.
         If a dict, for example: ``extra_params=dict(allow_maxshield=True)``.
+    space : None | str
+        The space of electrode coordinates to read in. None (default) assumes
+        there is only one electrodes.tsv file and reads that one in. If a
+        string is passed in, then only electrodes.tsv with specified space
+        BIDs-entity will be read in.
```
jasmainak, line 363:
```suggestion
        BIDS-entity will be read in.
```

mne_bids/read.py
```diff
@@ -344,6 +356,11 @@ def read_raw_bids(bids_fname, bids_root, extra_params=None,
     extra_params : None | dict
         Extra parameters to be passed to MNE read_raw_* functions.
         If a dict, for example: ``extra_params=dict(allow_maxshield=True)``.
+    space : None | str
```
jasmainak, line 359: why do you need this argument? You should get the `space` from `coordsystem.json`, no?

mne_bids/config.py
```diff
@@ -53,6 +55,27 @@
                       'eeg': allowed_extensions_eeg,
                       'ieeg': allowed_extensions_ieeg}
 
+# accepted BIDS formats, which may be subject to change
+# depending on the specification
+# XXX: Add support for MEG and EEG
+BIDS_IEEG_COORDINATE_FRAMES = ['acpc', 'pixels', 'other']
+BIDS_IEEG_COORDINATE_UNITS = ['m', 'cm', 'mm']
```
jasmainak, line 62: I like this!

mne_bids/tests/test_read.py
```diff
@@ -324,42 +326,122 @@ def test_handle_info_reading():
...
     ch_names = raw.ch_names
     elec_locs = np.random.random((len(ch_names), 3)).astype(float)
     ch_pos = dict(zip(ch_names, elec_locs))
-    montage = mne.channels.make_dig_montage(ch_pos=ch_pos,
-                                            coord_frame="mri")
-    raw.set_montage(montage)
-    write_raw_bids(raw, bids_basename, bids_root, overwrite=True)
+    coordinate_frames = ['mri', 'ras']
+    for coord_frame in coordinate_frames:
+        # XXX: Currently mne-bids can't handle two electrodes files
```
jasmainak, line 350: there can be two electrode files associated with a single raw file? BIDS allows this? for what purpose?

jasmainak:
anyone else wants to take a final look @sappelhoff or @hoechenberger ?

mne_bids/tests/test_read.py
adam2392: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#multiple-coordinate-systems

mne_bids/read.py
adam2392: If <MEG/EEG/IEEG>CoordinateSystem key in the coordsystem json is one of the keywords allowed per modality, then not needed. (I am pretty sure?) 

However, if it is "Other", then the `space` entity is needed in order to discern what the coordinate system is. E.g. MNI, etc. This is the only way I could see that would still follow the BIDS specification, and allowing some flexibility in not using their recommended coordinate system formats (imo too stringent rn since there's no easy way to automate ACPC reorientation).

jasmainak: Adding more arguments to this function can be a bad precedent. We should try to read as much possible and throw warnings to encourage users to have a more complete dataset that can be read into MNE. This just corresponds to filling up one entry in one file right?

agramfort: I agree with @jasmainak 

don't try to do too many things in one PR @adam2392 

this one was pretty close to merge and we need to rediscuss again and maybe revert/adjust

adam2392: Okay that's a fair point. I had included the space parameter to get around having multiple electrodes.tsv files, but if we want to proceed w/ only allowing one per dataset when reading, then that is fine, else warn(?). 

Let me know what you think and I can change accordingly.

agramfort:
aim for the minimal stuff to get the PR in.

>

mne_bids/read.py
jasmainak: yes that sounds fair to me. BIDS has too many bells and whistles, let's make it work in stages incrementally.

adam2392:
> yes that sounds fair to me. BIDS has too many bells and whistles, let's make it work in stages incrementally.

Okay! @jasmainak took out the ability to read in electrodes.tsv if there are multiple for that subject/session. 

@agramfort sorry for that. I think I got over-eager to make things work in one go :p.

agramfort:
CIs are not happy @adam2392

adam2392:
I updated the ieeg example to use the updated EDF datasets in misc PR rn, which should fix the issues w/ the CI and also the readability issues raised about the example in #404.

It "should" work once this PR (https://github.com/mne-tools/mne-python/pull/7768) is merged in + https://github.com/mne-tools/mne-misc-data/issues/2.

jasmainak:
phew, what a saga. Let us know when you're ready :)

agramfort:
@jasmainak merge if ok for you and CIs turn out green

adam2392:
@agramfort 

Took a look at the config.yml for circleci, and it looks like it works off the pip install mne, which won't work until the latest mne is released?

jasmainak:
I don't think it's a version issue. You need to force an update to the dataset and clear the cache on Circle CI

adam2392:
> I don't think it's a version issue. You need to force an update to the dataset and clear the cache on Circle CI

But the current version on circleci is 0.20.4 and not the dev version? Whereas its the dev version that has the updated mne-misc-dataset version 0.6: 

https://github.com/mne-tools/mne-python/blob/65b0c6ed590dc88a135ce3f7e78946bf5bbcf305/mne/datasets/utils.py#L241

Sorry not sure what to change.

jasmainak:
interesting, now the datasets versions are tied to the software version. Could you then update CircleCI to use the MNE dev version? I think it's fair to assume that users who use dev version of mne-bids can also figure out how to use the dev version of mne. This will be fixed once mne makes a release.

agramfort:
+1 to update circleci to mne dev

I am thinking that ideally we should expose version numbers in fetchers...

the datasets module could seriously be improved...

>

adam2392:
> interesting, now the datasets versions are tied to the software version. Could you then update CircleCI to use the MNE dev version? I think it's fair to assume that users who use dev version of mne-bids can also figure out how to use the dev version of mne. This will be fixed once mne makes a release.

Finally fixed!

agramfort:
thx @adam2392 !

PR #411, opened 2020-05-16
TESTS: Support non-global bids-validator in test_write.py; fix one test to work with mne-python:master

hoechenberger:
PR Description
--------------
- I cannot install `bids-validator` (or node) globally, so I'm now leveraging the `VALIDATOR_EXECUTABLE` environment variable from `test_write.py` to allow the user to pass a custom executable path to the test
- After was able run the tests thanks to this adjustment, I realized one of them kept failing <strike>because it expected an exception (from MNE) that simply has a different type and phrasing upstream. So I fixed this. Makes me wonder why our CI never caught this, and</strike> with `mne-python:master`. Wonder if we shouldn't remove this particular test altogether anyway, because it just checks if MNE raises a certain exception. That should be ensured through unit tests upstream…

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -81,17 +81,18 @@ warning_str = dict(
 @pytest.fixture(scope="session")
 def _bids_validate():
     """Fixture to run BIDS validator."""
-    shell = False
-    bids_validator_exe = ['bids-validator', '--config.error=41',
-                          '--config.error=41']
+    vadlidator_args = ['--config.error=41']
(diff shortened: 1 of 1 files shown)


hoechenberger:
Wow what is this sorcery!

hoechenberger:
Ok so 95db915 fixes `mne-bids` tests for working with `mne-python:master`, but breaks it for `mne-python:0.20` ugh

hoechenberger:
Closing in favor of #412, #413

PR #420, opened 2020-05-20
fix example that did not make sense

sappelhoff:
I was reading through the discussion on https://groups.google.com/forum/#!topic/bids-discussion/deTNEc-lEZQ

> "TL;DR: let's create tools that generate a large chunk of the BIDS (folder hierarchy and filenaming) directly when the data is acquired."

... and thought "hey, we can do that with MNE-BIDS". When checking our example on that I realized that there are several things that could be clarified.

### DIFF
--- a/examples/create_bids_folder.py
+++ b/examples/create_bids_folder.py
@@ -54,6 +54,6 @@ print(path_folder)
 
 # Note that passing `make_dir=True` will create the folder hierarchy. If that
 # path already exists, you can make use of the `overwrite` parameter.
-# If overwrite == False then no existing folders will be removed, however
-# if overwrite == True then any existing folders at the session level
+# If `overwrite=False` then no existing folders will be removed, however
+# if `overwrite=True` then any existing folders at the session level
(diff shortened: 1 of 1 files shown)


examples/create_bids_folder.py
```diff
@@ -47,10 +47,13 @@
...
                                 kind='meg', bids_root='path/to/project',
                                 make_dir=False)
 print(path_folder)
 
-# Note that passing `make_dir=True` will create the folder hierarchy, ignoring
-# errors if the folder already exists.
+# Note that passing `make_dir=True` will create the folder hierarchy. If that
+# path already exists, you can make use of the `overwrite` parameter.
+# If overwrite == False then no existing folders will be removed, however
+# if overwrite == True then any existing folders at the session level
```
hoechenberger, line 58:
```suggestion
# If `overwrite=False` then no existing folders will be removed, however
# if `overwrite=True` then any existing folders at the session level
```

hoechenberger:
Thanks @sappelhoff!

PR #446, opened 2020-06-08
[MRG] Updating make_bids_basename to return BIDSPath, allowing users to dynamically alter the bids file name

adam2392:
PR Description
--------------

Closes: #445 
Closes: #257

This is a draft to let maintainers see what I mean. Once comments are suggested and made, will iterate. The goal of this PR is to enable a bids filename created via `make_bids_basename` to be dynamically altered based on its BIDS entity without calling `make_bids_basename` again and again. 

In order to do so,  `make_bids_basename` now returns a `BIDSPath` object started by @jasmainak. In order to make the PR work w/o changing a lot of code, it was ideal to give `BIDSPath` as many string functionality as possible, because at the end of the day it is just a structured string, stored as a dictionary.

I override various magic methods. The `__str__` and  `__repr__` of `BIDSPath` is slightly borrowed from  pathlib.Path. Some "magic" method behavior is listed here:

1. comparison `__eq__`: makes BIDSPath as a str
2. `__fspath__`:  same as pathlib.Path
3. ~~`__bytes__`: same as pathlib.Path~~
4. `copy()`: returns deepcopy of self
6. `update`: updates multiple  entities at once, allowing aliases for subject, session, recording, acquisition, processing
7. property `entities`: returns an ordered dictionary of all the BIDS entitites.

TODO:
- [x] figure out how to integrate this change given that almost the entire codebase assumes `make_bids_basename` outputs a string. ~~Currently, I have a flag `as_str=True` to make tests pass,  but idk if that's the best solution.~~

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


jasmainak:
> I guess my concern is right now, BIDSPath if printed is a string, but when passed into functions gets seen as a dict.

isn't this the desired behavior?

adam2392:
> > I guess my concern is right now, BIDSPath if printed is a string, but when passed into functions gets seen as a dict.
> 
> isn't this the desired behavior?

Yes true, I was just getting worried about what happens when downstream functions do things tho that assume it's a string, but turns out this was mitigated by just overriding some of the magic methods for BIDSPath.

mne_bids/utils.py
```diff
@@ -29,6 +30,128 @@
...
+        self.as_str()
+
+    def __setitem__(self, key, value):
+        """Set item as a dictionary, and perform validation checks."""
+        if key not in ('sub', 'ses', 'task', 'acq',
+                       'proc', 'acq', 'run', 'rec',
+                       'space', 'suffix', 'prefix'):
+            raise ValueError('Key must be one of blah, got %s' % key)
+
+        return dict.__setitem__(self, key, value)
```
agramfort, line 95: just set attributes to the instance

mne_bids/utils.py
```diff
@@ -29,6 +30,128 @@
...
+    Examples
+    --------
+    >>> bids_basename = make_bids_basename(subject='test', session='two', task='mytask', suffix='data.csv')  # noqa: E501
+    >>> print(bids_basename)
+    sub-test_ses-two_task-mytask_data.csv
+    >>> print(bids_basename.__repr__())
+    BIDSPath (sub-test_ses-two_task-mytask_data.csv)
+    """
+
+    def __init__(self, *args, **kwargs):
```
agramfort, line 83: args and kwargs are bad practice in general as it's very hard to document

agramfort: just explicit all parameters and it will help you documenting the defaults

mne_bids/utils.py
```diff
@@ -29,6 +30,128 @@
...
+        """Return the string representation of the path."""
+        return self._get_name()
+
+    def __repr__(self):
+        """Representation in the style of `pathlib.Path`."""
+        return "{}({!r})".format(self.__class__.__name__, self._get_name())
+
+    def __add__(self, value):
+        """Return self+value."""
+        return self.as_str() + value
```
agramfort, line 107: if you add 2 objects of type T you should expect T + T to be still of type T. It's another reason to move away from dict. I am not sure why add is necessary here also.

mne_bids/utils.py
```diff
@@ -29,6 +30,128 @@
...
+
+    def replace(self, old, new, *args) -> str:
+        """Provide replace."""
+        return self.as_str().replace(old, new, *args)
+
+    def _get_name(self):
+        basename = _gen_bids_basename(**self)
+        return basename
+
+    def as_str(self) -> str:
```
agramfort, line 150: this should just be __str__ method.

mne_bids/utils.py
jasmainak: I don't get this either. Agreed about moving away from `dict` and simply setting attributes.

adam2392: Okay, so if people want to say do something like:

```
bids_fpath = bids_basename + '_ieeg.vhdr'
```

what would be the expectation (asking because this came up in the codebase)? For them to explicitly convert to a string? E.g.

```
bids_fpath = str(bids_basename) + '_ieeg.vhdr'
```

jasmainak: I like option 2 better

adam2392:
> I am not against an object BIDSPath but not with this design. Also why a method make_bids_basename that just calls the constructor of BIDSPath. If we do with the object that we should adopt its API and use the constructor.

Okay I am refactoring to make  it operate as an object instead of dict. 

Regarding how it is constructed, do we want to keep the function `make_bids_basename`? Do we just keep that, but allow people to invoke BIDSPath directly?

jasmainak:
I would keep `make_bids_basename` for now. It can return an instance of `BIDSPath`

mne_bids/write.py
```diff
@@ -976,16 +978,18 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
+        sub=subject_id, ses=session_id,
+        acq=acquisition, prefix=data_path,
+        suffix='coordsystem.json',
         on_invalid_er_task=on_invalid_er_task)
     electrodes_fname = _gen_bids_basename(
-        subject=subject_id, session=session_id, acquisition=acquisition,
-        suffix='electrodes.tsv', prefix=data_path,
+        sub=subject_id, ses=session_id,
+        acq=acquisition, prefix=data_path,
+        suffix='electrodes.tsv',
```
jasmainak, line 992: all this code could benefit from the `BIDSPath` api. And also below. Can you do that? Like in my original PR

mne_bids/utils.py
```diff
@@ -29,6 +30,142 @@
...
+        """
+        return os.fsencode(str(self))
+
+    def __eq__(self, other):
+        """Compare str representations."""
+        return str(self) == str(other)
+
+    def startswith(self, prefix, start=None, end=None):
+        """Return True/False for the str startswith."""
+        return str(self).startswith(prefix, start=start, end=end)
```
jasmainak, line 148:
```suggestion
```

I would say don't mix `BIDSPath` with `str` or other kinds of objects. The class should be as thin as possible

mne_bids/utils.py
```diff
@@ -29,6 +30,142 @@
...
+
+    def endswith(self, suffix, start=None, end=None):
+        """Return True/False for the str endswith."""
+        return str(self).endswith(suffix, start=start, end=end)
```
jasmainak, line 152:
```suggestion
```

mne_bids/utils.py
```diff
@@ -29,6 +30,142 @@
...
+        """Set item as a dictionary, and perform validation checks."""
+        if key not in ('sub', 'ses', 'task', 'acq',
+                       'proc', 'acq', 'run', 'rec',
+                       'space', 'suffix', 'prefix'):
+            raise ValueError('Key must be one of blah, got %s' % key)
+        setattr(self, key, value)
+
+    def __str__(self):
+        """Return the string representation of the path."""
+        return self._get_name()
```
jasmainak, line 125: why make another function for just 2 lines? This is not good. You are sending the person reading the code on a treasure hunt!

mne_bids/utils.py
```diff
@@ -29,6 +30,142 @@
...
+    Examples
+    --------
+    >>> bids_basename = make_bids_basename(subject='test', session='two', task='mytask', suffix='data.csv')  # noqa: E501
+    >>> print(bids_basename)
+    sub-test_ses-two_task-mytask_data.csv
+    >>> print(bids_basename.__repr__())
+    BIDSPath (sub-test_ses-two_task-mytask_data.csv)
+    """
+
+    def __init__(self, sub=None, ses=None, task=None,
```
jasmainak, line 83: sub -> subject

as documented?

mne_bids/write.py
adam2392: How should wee handle the `on_valid_er_task`? I'm not too familiar w/ what's going on there. Can we just stick that kwarg into `BIDSPath`?

jasmainak: no it should be handled outside `BIDSPath`

adam2392: Okay, I kept these here that use the `_gen_bids_basename` for `on_valid_er_task`, but handled some of the below code.

mne_bids/utils.py
```diff
@@ -814,14 +952,14 @@ def _gen_bids_basename(*, subject=None, session=None, task=None,
                f'but received: {on_invalid_er_task}')
         raise ValueError(msg)
 
-    order = OrderedDict([('sub', subject),
-                         ('ses', session),
+    order = OrderedDict([('sub', sub),
```
jasmainak, line 955: why do you change this throughout the codebase?

jasmainak:
@adam2392 your challenge is to get more red lines than green lines. Can it be done? :)

You can also update/simplify code in tests I believe (don't change logic though)

mne_bids/tests/test_write.py
```diff
@@ -498,20 +499,24 @@ def test_kit(_bids_validate):
...
     _bids_validate(bids_root)
     # ensure the marker files are renamed correctly
-    marker_fname = make_bids_basename(
-        subject=subject_id2, session=session_id, task=task, run=run,
-        suffix='markers.sqd', acquisition='pre',
-        prefix=os.path.join(bids_root, 'sub-02', 'ses-01', 'meg'))
+    marker_fname = marker_fname.copy()
+    marker_fname.acq = 'pre'
+    marker_fname.sub = subject_id2
+    marker_fname.prefix = op.join(bids_root, 'sub-02', 'ses-01', 'meg')
```
jasmainak, line 511: just a thought. If you had a method called `update` then you could do something like:

```py
marker_fname.copy().update(acq=acq, sub=sub, prefix=prefix)
```

mne_bids/tests/test_write.py
```diff
@@ -498,20 +499,24 @@ def test_kit(_bids_validate):
...
+    # marker_fname = make_bids_basename(
+    #     subject=subject_id2, session=session_id, task=task, run=run,
+    #     suffix='markers.sqd', acquisition='pre',
+    #     prefix=os.path.join(bids_root, 'sub-02', 'ses-01', 'meg'))
```
jasmainak, line 515: what is this for???

mne_bids/utils.py
```diff
@@ -29,6 +30,143 @@
...
+                entities[key] = value
+        return entities
+
+    def __setitem__(self, key, value):
+        """Set item as a dictionary, and perform validation checks."""
+        if key not in ('sub', 'ses', 'task', 'acq',
+                       'proc', 'acq', 'run', 'rec',
+                       'space', 'suffix', 'prefix'):
+            raise ValueError('Key must be one of blah, got %s' % key)
+        setattr(self, key, value)
```
jasmainak, line 133: do you use this in the code anywhere? I would remove any magic methods that are not used in the code anywhere and make the class minimal

mne_bids/write.py
jasmainak: you might want @hoechenberger 's opinion here.

mne_bids/utils.py
adam2392: Sorry this was leftover from when we were using it as a dict.

mne_bids/tests/test_write.py
adam2392: Sorry commented out, but forgot to remove.

mne_bids/tests/test_write.py
adam2392: good idea! Implemented.

mne_bids/utils.py
adam2392: Was playing around w/ refactoring and think this got changed by accident. Sorry!

adam2392:
> @adam2392 your challenge is to get more red lines than green lines. Can it be done? :)
> 
> You can also update/simplify code in tests I believe (don't change logic though)

This turned out to be quite a challenge imo! So the added class turns out to be ~170 lines, so... I think I did neutral :p. I would hope that the changes in a few of the lines are convincing enough for you and @agramfort that it simplifies user code(?)

mne_bids/tests/test_write.py
```diff
@@ -288,7 +290,7 @@ def test_fif(_bids_validate):
     raw_fname2 = op.join(data_path2, 'sample_audvis_raw.fif')
     raw.save(raw_fname2)
 
-    bids_basename2 = bids_basename.replace(subject_id, subject_id2)
+    bids_basename2 = bids_basename.copy().update(sub=subject_id2)
```
agramfort, line 293: sub -> subject?

mne_bids/write.py
```diff
@@ -977,37 +981,40 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
+    participants_json_fname.suffix = 'participants.json'
+
+    sidecar_fname = bids_fname.copy()
+    sidecar_fname.prefix = data_path
+    sidecar_fname.suffix = f'{kind}.json'
+
+    events_fname = sidecar_fname.copy()
+    events_fname.suffix = 'events.tsv'
+    channels_fname = sidecar_fname.copy()
+    channels_fname.suffix = 'channels.tsv'
```
agramfort, line 1012: why not using update method here?

mne_bids/utils.py
```diff
@@ -26,9 +27,171 @@
...
+    >>> print(bids_basename)
+    sub-test_ses-two_task-mytask_data.csv
+    >>> print(bids_basename.__repr__())
+    BIDSPath (sub-test_ses-two_task-mytask_data.csv)
+    """
+
+    def __init__(self, subject=None, session=None,
+                 task=None, acquisition=None, run=None, processing=None,
+                 recording=None, space=None, prefix=None, suffix=None,
+                 **kwargs):
```
agramfort, line 87: why the kwargs? I think the update logic is reminiscent of the dict we had before and should be reconsidered now that we have our own object

adam2392: I was wondering if it was possible to allow the 'aliases' for `sub: subject`, `ses: session`, etc.?

The reason is that, `_parse_bids_filename` currently uses those shorthands, so then it gets unwieldy in some areas.

agramfort:
$ python -c "import this" | grep "only"

don't make API a mess to save 4 key strokes

>

mne_bids/tests/test_write.py
```diff
@@ -497,21 +493,19 @@ def test_kit(_bids_validate):
...
 
     _bids_validate(bids_root)
     # ensure the marker files are renamed correctly
-    marker_fname = make_bids_basename(
-        subject=subject_id2, session=session_id, task=task, run=run,
-        suffix='markers.sqd', acquisition='pre',
-        prefix=os.path.join(bids_root, 'sub-02', 'ses-01', 'meg'))
+    prefix = op.join(bids_root, 'sub-02', 'ses-01', 'meg')
+    marker_fname = marker_fname.copy().update(acq='pre', sub=subject_id2,
+                                              prefix=prefix)
```
jasmainak, line 504: You can't do this in place? Like:

```py
marker_fname.update(acq='pre', sub=subject_id2, prefix=prefix)
```

mne_bids/tests/test_write.py
```diff
@@ -711,7 +706,8 @@ def test_edf(_bids_validate):
         read_raw_bids(bids_basename=bids_basename, bids_root=bids_root,
                       kind='eeg', extra_params=dict(foo='bar'))
 
-    bids_fname = bids_basename.replace('run-01', 'run-%s' % run2)
+    bids_fname = bids_basename.copy()
+    bids_fname.run = run2
```
jasmainak, line 710:
```suggestion
    bids_fname = bids_basename.copy().update(run=run2)
```

I would consistently use `update` everywhere

adam2392: Done.

adam2392:
> I would keep `make_bids_basename` for now. It can return an instance of `BIDSPath`

Agreed since this is the API ppl are used to for a long time. Perhaps down the line, a deprecation cycle is added? Unless, any other issues arise, I think the CIs should pass this time around and the comments were addressed to simplify as much of the tests as possible. 

I think BIDSPath and related config is around 170-180 linesish, so in total, maybe got rid of 20 or so lines? I think this PR could also be a good launching  ground for potentially a `DerivativesPath` in the future, since the derivatives PR is merged into the specification recently(?)

**Regarding what's left to do:** In order to document this class, what should I do? If there are any other major comments, let me know and I can address them later.

mne_bids/read.py
```diff
@@ -421,6 +421,10 @@ def read_raw_bids(bids_basename, bids_root, kind=None, extra_params=None,
         If the specified ``kind`` cannot be found in the dataset.
 
     """
+    # convert to string representation for downstream usage
+    if isinstance(bids_basename, BIDSPath):
+        bids_basename = str(bids_basename)
```
agramfort, line 426: I would just do this systematically and avoid the if

mne_bids/read.py
```diff
@@ -524,6 +528,9 @@ def get_matched_empty_room(bids_basename, bids_root):
         The basename corresponding to the best-matching empty-room measurement.
         Returns None if none was found.
     """
+    if isinstance(bids_basename, BIDSPath):
+        bids_basename = str(bids_basename)
```
agramfort, line 532: idem

mne_bids/tests/test_write.py
```diff
@@ -497,21 +495,18 @@ def test_kit(_bids_validate):
...
-        subject=subject_id2, session=session_id, task=task, run=run,
-        suffix='markers.sqd', acquisition='pre',
-        prefix=os.path.join(bids_root, 'sub-02', 'ses-01', 'meg'))
+    prefix = op.join(bids_root, 'sub-02', 'ses-01', 'meg')
+    marker_fname.update(acquisition='pre', subject=subject_id2, prefix=prefix)
     info = get_kit_info(marker_fname, False)[0]
     assert info['meas_date'] == get_kit_info(hpi_pre_fname,
                                              False)[0]['meas_date']
-    marker_fname = marker_fname.replace('acq-pre', 'acq-post')
+    marker_fname.acquisition = 'post'
```
agramfort, line 509: use update?

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+    def __init__(self, subject=None, session=None,
+                 task=None, acquisition=None, run=None, processing=None,
+                 recording=None, space=None, prefix=None, suffix=None):
+        self.update(subject=subject, session=session, task=task,
+                    acquisition=acquisition, run=run, processing=processing,
+                    recording=recording, space=space, prefix=prefix,
+                    suffix=suffix)
+
+    @property
+    def entities(self):
```
agramfort, line 92: this does not seem to be used publicly. I would make this a private property provided you actually need it

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+        """
+        return os.fsencode(str(self))
+
+    def __eq__(self, other):
+        """Compare str representations."""
+        return str(self) == str(other)
+
+    def __ne__(self, other):
+        """Compare str representations."""
+        return str(self) == str(other)
```
agramfort, line 128:
```suggestion
        return str(self) != str(other)
```
?
is this tested?

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+
+        Parameters
+        ----------
+        kwargs :
+            Allowed BIDS entities: see `BIDS_ENTITIES`.
+
+        Returns
+        -------
+        bidspath : instance of BIDSPath
+            The copied bidspath.
```
agramfort, line 151: this operates inplace

mne_bids/write.py
```diff
@@ -932,12 +932,24 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
                        "raw.times should not have changed since reading"
                        " in from the file. It may have been cropped.")
 
-    params = _parse_bids_filename(bids_basename, verbose)
-    subject_id, session_id = params['sub'], params['ses']
-    acquisition, task, run = params['acq'], params['task'], params['run']
+    # convert to BIDS Path
+    if isinstance(bids_basename, str):
+        params = _parse_bids_filename(bids_basename, verbose)
```
adam2392, line 937: (https://github.com/mne-tools/mne-bids/pull/446#discussion_r439810883)

@agramfort kept this 'if' here since it doesn't seem desirable to do the _parse_bids_filename every time, unless necessary, but if you think  it's better to also get rid of, then I can change.

mne_bids/utils.py
adam2392: Ah good point... Will add in tests for BIDSPath and all necessary func.

agramfort:
tell us when ready from your end

>

mne_bids/config.py
```diff
@@ -55,6 +55,11 @@
                       'eeg': allowed_extensions_eeg,
                       'ieeg': allowed_extensions_ieeg}
 
+# allowed BIDS entities
+BIDS_ENTITIES = ('subject', 'session', 'task', 'acquisition',
+                 'processing', 'run', 'recording',
+                 'space', 'suffix', 'prefix')
```
hoechenberger, line 61: Are `suffix` and `prefix` actually BIDS **entities** in the strict sense?

mne_bids/tests/test_write.py
```diff
@@ -636,15 +630,17 @@ def test_vhdr(_bids_validate):
 
     # Test that correct channel units are written ... and that bad channel
     # is in channels.tsv
-    channels_tsv_name = op.join(bids_root, 'sub-{}'.format(subject_id),
-                                'eeg', bids_basename_minimal + '_channels.tsv')
+    prefix = op.join(bids_root, 'sub-{}'.format(subject_id), 'eeg')
```
hoechenberger, line 633:
```suggestion
    prefix = op.join(bids_root, f'sub-{subject_id}', 'eeg')
```

mne_bids/tests/test_write.py
```diff
@@ -765,7 +761,7 @@ def test_edf(_bids_validate):
                        anonymize=dict(daysback=33000),
                        overwrite=True)
         data = _from_tsv(scans_tsv)
-        bids_fname = bids_basename + "_eeg.vhdr"
+        bids_fname = str(bids_basename) + "_eeg.vhdr"
```
hoechenberger, line 764:
```suggestion
        bids_fname = str(bids_basename) + '_eeg.vhdr'
```

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+    """Create a partial/full BIDS filename from its component parts.
+
+    BIDS filename prefixes have one or more pieces of metadata in them. They
+    must follow a particular order, which is followed by this function. This
+    will generate the *prefix* for a BIDS filename that can be used with many
+    subsequent files, or you may also give a suffix that will then complete
+    the file name.
+
+    Note that all parameters are not applicable to each kind of data. For
+    example, electrode location TSV files do not need a task field.
```
hoechenberger, line 44:
```suggestion
    Note that not all parameters are applicable to each kind of data. For
    example, electrode location TSV files do not need a "task" field.
```

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+        The coordinate space for an anatomical file. Corresponds to "space".
+    prefix : str | None
+        The prefix for the filename to be created. E.g., a path to the folder
+        in which you wish to create a file with this name.
+    suffix : str | None
+        The suffix for the filename to be created. E.g., 'audio.wav'.
+
+    Examples
+    --------
+    >>> bids_basename = make_bids_basename(subject='test', session='two', task='mytask', suffix='data.csv')  # noqa: E501
```
hoechenberger, line 72:
```suggestion
    >>> bids_basename = make_bids_basename(subject='test', session='two',
                                           task='mytask', suffix='data.csv')
```

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+            The copied bidspath.
+        """
+        return deepcopy(self)
+
+    def update(self, **kwargs):
+        """Update inplace BIDS entity key/value pairs in object.
+
+        Parameters
+        ----------
+        kwargs :
```
hoechenberger, line 145: dtype is missing

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+
+        Returns
+        -------
+        bidspath : instance of BIDSPath
+            The copied bidspath.
+        """
+        return deepcopy(self)
+
+    def update(self, **kwargs):
+        """Update inplace BIDS entity key/value pairs in object.
```
hoechenberger, line 141: This functions lacks documentation, I could only figure what it does by looking at the code. The `**kwargs` param should either be renamed to sth like `**vals` or `**data` or something. Also, wondering if accepting a `dict` instead would make sense?

mne_bids/utils.py
```diff
@@ -26,9 +27,144 @@
...
+            # error check allowed BIDS entity keywords
+            if key not in BIDS_ENTITIES:
+                raise ValueError('Key must be one of {BIDS_ENTITIES}, '
+                                 'got %s' % key)
+
+            # set entity value
+            setattr(self, key, value)
+
+        # run string representation to check validity of arguments
+        str(self)
```
hoechenberger, line 164: This seems like a bit of a hack, is there maybe a cleaner way to do it using a private function?

mne_bids/config.py
sappelhoff: > Are suffix and prefix actually BIDS entities in the strict sense?

nope :)

Cool relevant PR which will make our BIDS-related life easier in the future: https://github.com/bids-standard/bids-specification/pull/475

hoechenberger: > nope :)

So… wondering if `BIDS_ENTITIES` is the best choice of a name here, then… how about `BIDS_PATH_ENTITIES`? ;)

mne_bids/utils.py
```diff
@@ -26,9 +27,145 @@
...
+        The suffix for the filename to be created. E.g., 'audio.wav'.
+
+    Examples
+    --------
+    >>> bids_basename = make_bids_basename(subject='test', session='two',
+                                           task='mytask', suffix='data.csv')
+    >>> print(bids_basename)
+    sub-test_ses-two_task-mytask_data.csv
+    >>> print(bids_basename.__repr__())
+    BIDSPath (sub-test_ses-two_task-mytask_data.csv)
```
jasmainak, line 77: I'm wondering if the repr should actually list the different entities, something like:

```py
<BIDSPath |
prefix=None,
task='blah',
subject='blah'
>
```
Thoughts?

mne_bids/utils.py
```diff
@@ -26,9 +27,145 @@
...
+        """
+        return deepcopy(self)
+
+    def update(self, **kwargs):
+        """Update inplace BIDS entity key/value pairs in object.
+
+        Parameters
+        ----------
+        kwargs :
+            Allowed BIDS entities: see `BIDS_ENTITIES`.
```
jasmainak, line 147: Please just list all the entities, don't make the user chase documentation

mne_bids/write.py
```diff
@@ -977,37 +987,37 @@ def write_raw_bids(raw, bids_basename, bids_root, events_data=None,
...
+        subject=subject_id, session=session_id,
+        acquisition=acquisition,
+        prefix=data_path, suffix='coordsystem.json',
         on_invalid_er_task=on_invalid_er_task)
     electrodes_fname = _gen_bids_basename(
-        subject=subject_id, session=session_id, acquisition=acquisition,
-        suffix='electrodes.tsv', prefix=data_path,
+        subject=subject_id, session=session_id,
+        acquisition=acquisition,
+        prefix=data_path, suffix='electrodes.tsv',
```
jasmainak, line 1002: why don't you do:

```py
electrodes_fname = _gen_bids_basename(bids_path, on_valid_er_task=on_valid_er_task)
```
?

jasmainak:
PR is beginning to look promising, need more red lines though ;-) See my comment related to `_gen_bids_basename`

mne_bids/utils.py
adam2392: This is a stupid question... but these are of dtype 'dict' right? lol

adam2392: Also, renamed as `entities`

mne_bids/utils.py
adam2392: Renamed it as `entities`.

adam2392: Also added a few examples into the docstring.

mne_bids/utils.py
adam2392: I don't have a great way of integrating it off the top of my head because the `_gen_bids_basename` function does the construction of the basename and also error checking inside. 

Unless you think it should just get refactored too, but I was leaning towards the side w/o trying to do too much in this one  PR :P.

mne_bids/utils.py
adam2392: I like that. Perhaps in addition:

```Python
<BIDSPath | (sub-test_ses-two_task-mytask_data.csv)
prefix=None,
task='blah',
subject='blah'
...
>
```

?

mne_bids/write.py
adam2392: Hmm I tried and  then `test_find_matching_sidecar` broke. I think it is because the coordsystem and electrodes files don't need all the other task/run/recording/processing/etc. according to the current specification: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#channels-description-_channelstsv

However, that brings up another point: It isn't clear if the template needs to match across different modalities. Right now for coordsystem, subject/session/space/acquisition are generally needed depending on modality. So I added space in.

mne_bids/utils.py
adam2392: Is it possible to leave it as a public property? It allows someone to quickly determine which entity values are actually part of the `bids_basename`. 

It also allows one to quickly pass the entities as a dictionary into say `_gen_bids_basename`, like `**bids_basename.entities`

adam2392:
> PR is beginning to look promising, need more red lines though ;-) See my comment related to `_gen_bids_basename`

I am at a limit in terms of red lines :(, unless you see a good workaround for `_gen_bids_basename`? I added your comment on`__repr__` and also some unit tests for BIDSPath. 

Lmk if you think there are other areas worth refactoring to create red lines.

mne_bids/utils.py
```diff
@@ -26,9 +27,175 @@
...
+               f"prefix={self.prefix}, \n" \
+               f"subject={self.subject}, \n" \
+               f"session={self.session}, \n" \
+               f"task={self.task}, \n" \
+               f"run={self.run}, \n" \
+               f"acquisition={self.acquisition}, \n" \
+               f"processing={self.processing}, \n" \
+               f"recording={self.recording}, \n" \
+               f"space={self.space}, \n" \
+               f"suffix={self.suffix}"
```
jasmainak, line 121: How about:

```py
'\n'.join(repr(x).split(', '))[1:-1]
```

or something of that sort. Keeping the class name in there.

mne_bids/write.py
jasmainak: ummm ... not sure why `test_find_matching_sidecar` would fail. Can you perhaps interact with @hoechenberger to make this happen? I think [this](https://github.com/mne-tools/mne-bids/pull/446#discussion_r440344654) is the API we should use with the new object

adam2392: Well I can make it work w/ the following appendage to the unit test. However, the bids specification isn't really clear if say... task and run are needed, or even allowed in coordsystem/electrodes files.

mne_bids/utils.py
adam2392: Sorry I am confused with what you mean. This would give me an infinite recursion. Is x supposed to be the various BIDS Path entities?

jasmainak: sorry I was playing around with it in my terminal. Better now?

mne_bids/tests/test_utils.py
```diff
@@ -357,8 +355,9 @@ def test_find_matching_sidecar(return_bids_test_dir):
     # Now find a sidecar
     sidecar_fname = _find_matching_sidecar(bids_basename, bids_root,
                                            'coordsystem.json')
-    expected_file = op.join('sub-01', 'ses-01', 'meg',
-                            'sub-01_ses-01_coordsystem.json')
+    expected_file = op.join(
+        'sub-01', 'ses-01', 'meg',
+        'sub-01_ses-01_task-testing_run-01_coordsystem.json')
```
adam2392, line 360: Need this change if `_gen_bids_basename()` takes in bidspath instead of what it was originally in https://github.com/mne-tools/mne-bids/pull/446/files#r440344654

jasmainak: humm, can't you do:

```py
bids_path.update(task=None, run=None)
_gen_bids_basename(bids_path)
```
?

adam2392: Updated accordingly.

mne_bids/utils.py
adam2392: Let me know if the change is what you meant... It ends up looking like this in terminal:

```
bids_basename = BIDSPath(sub-01_ses-02_acq-03_run-2) | 
subject=01
session=02
acquisition=03
run=2
```

agramfort:
@adam2392 can you tell us when CIs are green and it's good to go on your end?

mne_bids/read.py
```diff
@@ -26,8 +26,8 @@
 from mne_bids.utils import (_parse_bids_filename, _extract_landmarks,
                             _find_matching_sidecar, _parse_ext,
                             _get_ch_type_mapping, make_bids_folders,
-                            _gen_bids_basename,
-                            _estimate_line_freq, _get_kinds_for_sub)
+                            _gen_bids_basename, _estimate_line_freq,
+                            _get_kinds_for_sub)
```
jasmainak, line 30: don't touch lines that are not relevant to the PR!

mne_bids/utils.py
```diff
@@ -26,9 +27,166 @@
...
+        return str(self)
+
+    def __bytes__(self):
+        """Return the bytes representation of the path.
+
+        This is only recommended to use under Unix.
+        """
+        return os.fsencode(str(self))
+
+    def __eq__(self, other):
```
jasmainak, line 125: is this tested?

mne_bids/utils.py
```diff
@@ -26,9 +27,166 @@
...
+            if value is not None:
+                entities[key] = value
+        return entities
+
+    def __str__(self):
+        """Return the string representation of the path."""
+        basename = _gen_bids_basename(**self.entities)
+        return basename
+
+    def __repr__(self):
```
jasmainak, line 109: is this tested?

mne_bids/utils.py
```diff
@@ -26,9 +27,166 @@
...
+        """Representation in the style of `pathlib.Path`."""
+        return f'{self.__class__.__name__}({str(self)}) | \n' + \
+               '\n'.join(f'{key}={val}' for key, val in self.entities.items())
+
+    def __fspath__(self):
+        """Return the string representation for any fs functions."""
+        return str(self)
+
+    def __bytes__(self):
```
jasmainak, line 118: tested? Please make sure all the methods are tested

mne_bids/utils.py
```diff
@@ -26,9 +27,166 @@
 from mne.io.constants import FIFF
 from mne.time_frequency import psd_array_welch
 
+from mne_bids.config import BIDS_PATH_ENTITIES
 from mne_bids.tsv_handler import _to_tsv, _tsv_to_str
 
 
+class BIDSPath(object):
```
jasmainak, line 34: have you updated `api.rst` ?

jasmainak:
@adam2392 I am on the code now, don't push any commit

mne_bids/utils.py
```diff
@@ -26,9 +27,166 @@
...
+
+    def __str__(self):
+        """Return the string representation of the path."""
+        basename = _gen_bids_basename(**self.entities)
+        return basename
+
+    def __repr__(self):
+        """Representation in the style of `pathlib.Path`."""
+        return f'{self.__class__.__name__}({str(self)}) | \n' + \
+               '\n'.join(f'{key}={val}' for key, val in self.entities.items())
```
jasmainak, line 112: ughghg pytest gives very ugly output if you do ```> p bids_fname```. Let's revert this to match the output of `__str__`

jasmainak:
@adam2392 pushed a tiny cosmetic commit. Can you make the CIs green?

Also:

- can you give me a high-level picture of what kind of paths are expected to be instances of `BIDSPath` and what paths are `str`?
- what's the plan for bids_fname? why is not an instance of `BIDSPath`?
- why do we still need to use `_parse_bids_filename` instead of using `bids_path.entities`?
- why does `get_head_mri_trans` still takes only `str` for `bids_basename`

I fear all these inconsistencies will create problems in the future. We need to make sure we have a clear consistent API. I also realized all this is not straightforward because of edge cases, so let us know how you'd like to proceed.

mne_bids/utils.py
adam2392: Should this be added to `Utils`, or create a separate section for `Classes`?

mne_bids/utils.py
adam2392: This is tested in `test_utils.py::test_bids_path`

mne_bids/utils.py
jasmainak: utils sounds reasonable to me.

adam2392:
> @adam2392 pushed a tiny cosmetic commit. Can you make the CIs green?
> 
> Also:
> 
> * can you give me a high-level picture of what kind of paths are expected to be instances of `BIDSPath` and what paths are `str`?

So in `nibabel`, there are some instances where they assume the input filepath to be a string, cuz they use things like `replace`, or `startswith`, or `endswith`. Since those methods were not desired, then those areas need to just make do w/ keeping BIDSPath as a str.

Otherwise, I wasn't sure if you wanted me to make drastic changes, but seeing as how this can replace more of the underlying API, then I went ahead and modified `_gen_bids_basename` and also added the rest of the error checks in.

> * what's the plan for bids_fname? why is not an instance of `BIDSPath`?



> * why do we still need to use `_parse_bids_filename` instead of using `bids_path.entities`?

If we allow users to pass in `str` to `read_raw_bids` and `write_raw_bids`, then we need this to setup the BIDSPath.

> * why does `get_head_mri_trans` still takes only `str` for `bids_basename`

See above on why str is needed.

>  Misc issues:
- `recording` should be `rec`: In [specification](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html), rec- is used instead of full recording. A lot of the unit tests broke when I took your commit in. I can make a separate PR if you think that's too unwieldy, but I just needed to change  `_parse_bids_filename` and the upstream calls to it.
- How to handle emptyroom stuff (I don't use this). Because this was handled in _gen_bids_basename before, and this was cleanly separated from `make_bids_basename` with the use of `get_matched_empty_room`. I'm not sure how to best incorporate the `on_invalid...` params now. I'll try some iterations and will lyk how it goes.

jasmainak:
> So in nibabel, there are some instances where they assume the input filepath to be a string

can nibabel handle Path objects? they are not strings so how does nibabel handle that?

> I think this still needs to be a separate function because it handles how kind is used right?

humm ... why can't `kind` be a suffix?

> In addition, anywhere that they use a str representation, this function is needed.

okay I see. Now it's not so clear to me the benefit of using a bids_path object. We would basically have to support both `str` and `bids_path`. And `bids_path` would not work everywhere. It seems more a burden. What do you think?

agramfort:
looking at https://2286-89170358-gh.circle-artifacts.com/0/dev/generated/mne_bids.utils.BIDSPath.html#mne_bids.utils.BIDSPath

I am a bit confused why you need parameters check_empty_room.

same remark for space parameter.

basically we should not overload BIDSPath for things which are not a pure Path thing.

adam2392:
> can nibabel handle Path objects? they are not strings so how does nibabel handle that?

Hmm... okay I am SO sorry... lol so I was using nibabel 2.5.1 cuz I've been using the same dev version for mne-bids/mne-python for awhile, but nibabel is now on 3.1.0, which supports Paths. `write_anat` can be refactored to use BIDSPath now instead of str.

> humm ... why can't `kind` be a suffix?

`kind` can be, and I am not against it, and then also `ext`? 

But my main issue then was how can we keep `suffix`, `kind` and `ext`? Since kind is only eeg, ieeg, or t1w, but suffix can be way more. 

> okay I see. Now it's not so clear to me the benefit of using a bids_path object. We would basically have to support both `str` and `bids_path`. And `bids_path` would not work everywhere. It seems more a burden. What do you think?

So we don't need to support `str` inside our private functions if you and @agramfort are okay with refactoring things like `_find_matching_sidecar` to use BIDSPath. 

However, public API, should still allow `str` input right? We just do the conversion to BIDSPath for them if they didn't use `make_bids_basename`.

adam2392:
> I am a bit confused why you need parameters check_empty_room.
> 
> same remark for space parameter.

The `space` parameter is part of the BIDS entities though, so shouldn't that be included?

For the `check_empty_room` parameter, this is an issue because whenever we have a `subject=emptyroom` bids basename, there are a set of issues that I'm having trouble dealing with:

1. The original emptyroom check occurs in `_gen_bids_basename` only IF `subject==emptyroom`.    See: https://github.com/mne-tools/mne-bids/blob/131c1aa434a11c51b061733fd2dc8de7ef1776b5/mne_bids/utils.py#L1066-L1068

This is totally fine when you're first creating the bids-basename, but this leads to problems whenever you use this again within functions. See 2.

2. In https://github.com/mne-tools/mne-bids/blob/131c1aa434a11c51b061733fd2dc8de7ef1776b5/mne_bids/write.py#L1024-L1040
`make_bids_basename` is used  to setup scans.tsv and other sidecar files. But because we are checking emptyroom parameters when `subject==emptyroom`, then we get an error that shouldn't occur here.

adam2392:
I've updated the comments above and left some notes to trace back to each decision.

Everything you have commented on, I tried addressing and summarizing above. **However**, I am not confident in the integration of 'emptyroom' subjects, so hoping to get your input there.

mne_bids/utils.py
adam2392: I don't think so. Do you know a good way to test __repr__?

mne_bids/utils.py
adam2392: Removed. I don't think __bytes__ is needed to act like a Path object.

mne_bids/config.py
```diff
@@ -24,6 +25,13 @@
                       '.edf': 'n/a', '.set': 'n/a', '.fdt': 'n/a',
                       '.mef': 'n/a', '.nwb': 'n/a'}
 
+# file-extension map to mne-python readers
+reader = {'.con': io.read_raw_kit, '.sqd': io.read_raw_kit,
+          '.fif': io.read_raw_fif, '.pdf': io.read_raw_bti,
+          '.ds': io.read_raw_ctf, '.vhdr': io.read_raw_brainvision,
+          '.edf': io.read_raw_edf, '.bdf': io.read_raw_bdf,
+          '.set': io.read_raw_eeglab}
```
adam2392, line 33: Needed to move this here to prevent circular imports.

mne_bids/utils.py
```diff
@@ -26,9 +27,249 @@
...
+        """Copy the instance.
+
+        Returns
+        -------
+        bidspath : instance of BIDSPath
+            The copied bidspath.
+        """
+        return deepcopy(self)
+
+    def get_bids_fname(self, kind=None, bids_root=None, extension=None):
```
adam2392, line 175: Essentially copied version of `_make_bids_fname`.

mne_bids/utils.py
```diff
@@ -915,18 +1146,79 @@ def make_bids_basename(subject=None, session=None, task=None,
...
-                  acquisition=acquisition, run=run, processing=processing,
-                  recording=recording, space=space, prefix=prefix,
-                  suffix=suffix)
+    return BIDSPath(subject=subject, session=session, task=task,
+                    acquisition=acquisition, run=run, processing=processing,
+                    recording=recording, space=space, prefix=prefix,
+                    suffix=suffix)
+
+
+def _get_bids_fname_from_filesystem(*, bids_basename, bids_root, sub, ses,
```
adam2392, line 1163: Needed to move this here to prevent circular imports.

mne_bids/utils.py
```diff
@@ -915,18 +1146,79 @@ def make_bids_basename(subject=None, session=None, task=None,
...
+                   'BIDS validator on your data.')
+            raise RuntimeError(msg)
+
+        matching_path = matching_paths[0]
+        bids_fname = op.basename(matching_path)
+
+    return bids_fname
+
+
+def _infer_kind(*, bids_basename, bids_root, sub, ses):
```
adam2392, line 1205: Needed to move here to prevent circular imports.

mne_bids/read.py
```diff
@@ -264,118 +258,6 @@ def _handle_channels_reading(channels_fname, bids_fname, raw):
...
-                   'BIDS validator on your data.')
-            raise RuntimeError(msg)
-
-        matching_path = matching_paths[0]
-        bids_fname = op.basename(matching_path)
-
-    return bids_fname
-
-
-def _make_bids_fname(bids_basename, bids_root=None, kind=None, extension=None,
```
adam2392, line 331: moved whole function into BIDSPath.

mne_bids/read.py
```diff
@@ -264,118 +258,6 @@ def _handle_channels_reading(channels_fname, bids_fname, raw):
...
-        msg = (f'Found data of more than one recording modality. Please '
-               f'pass the `kind` parameter to specify which data to load. '
-               f'Found the following kinds: {kinds}')
-        raise RuntimeError(msg)
-
-    assert len(kinds) == 1
-    return kinds[0]
-
-
-def _get_bids_fname_from_filesystem(*, bids_basename, bids_root, sub, ses,
```
adam2392, line 289: Copied to utils.py

mne_bids/read.py
```diff
@@ -264,118 +258,6 @@ def _handle_channels_reading(channels_fname, bids_fname, raw):
     return raw
 
 
-def _infer_kind(*, bids_basename, bids_root, sub, ses):
```
adam2392, line 267: Copied to utils.py

mne_bids/read.py
```diff
@@ -22,18 +22,12 @@
...
 from mne_bids.config import (ALLOWED_EXTENSIONS, _convert_hand_options,
-                             _convert_sex_options)
+                             _convert_sex_options, reader)
 from mne_bids.utils import (_parse_bids_filename, _extract_landmarks,
                             _find_matching_sidecar, _parse_ext,
                             _get_ch_type_mapping, make_bids_folders,
-                            _gen_bids_basename,
-                            _estimate_line_freq, _get_kinds_for_sub)
-
-reader = {'.con': io.read_raw_kit, '.sqd': io.read_raw_kit,
```
adam2392, line 32: Moved to config.py

agramfort:
this handling of emptyroom and the _gen_bids_basename was really a code smell. I pushed a changed here. Let me know what you think. Basically  I made so that BIDSPath does not do any check and you only get checks with make_bids_basename keeping BIDSPath as an internal tool

agramfort:
@adam2392 can you check what I did and update what's new? than good to go from my end

doc/whats_new.rst
```diff
@@ -47,7 +48,8 @@ API
...
 - BIDS conformity: The ``_part-%d`` entity is now called ``_split-`` throughout BIDS, MNE, and MNE-BIDS, by `Stefan Appelhoff`_ (`#417 <https://github.com/mne-tools/mne-bids/pull/417>`_)
-- The :code:`mne_bids.wite.make_bids_basename` function has been moved to :func:`mne_bids.utils.make_bids_basename`. Like before, it can also be accessed via `mne_bids.make_bids_basename`, by `Richard Höchenberger`_ (`#424 <https://github.com/mne-tools/mne-bids/pull/424>`_)
+- The :code:`mne_bids.write.make_bids_basename` function has been moved to :func:`mne_bids.utils.make_bids_basename`. Like before, it can also be accessed via `mne_bids.make_bids_basename`, by `Richard Höchenberger`_ (`#424 <https://github.com/mne-tools/mne-bids/pull/424>`_)
+- The :func:`mne_bids.make_bids_basename` function has been updated to create a :code:`mne_bids.utils.BIDSPath` object, which operates like a path object and allows dynamic updating of BIDs entities, by `Adam Li`_ (`#446 <https://github.com/mne-tools/mne-bids/pull/446>`_)
```
jasmainak, line 52: the function still exists, right?

adam2392: `make_bids_basename`? Yes, but now it solely returns a BIDSPath object.

mne_bids/utils.py
jasmainak: just call it :)

BIDSPath().__repr__

or 

assert 'blah' in BIDSPath().__repr__

adam2392: Ah okay that's what I ended up figuring heh. Added in now! Lmk if it doesn't work.

mne_bids/utils.py
jasmainak: So, is `BIDSPath` by default not the full path? Why is that?

doc/whats_new.rst
jasmainak: sorry my bad :)

mne_bids/utils.py
adam2392: If the `BIDSPath` object is the full path, then it's not really the `bids_basename` anymore, which doesn't seem required by the API right now though? 

The internals of `read_raw_bids` and `write_raw_bids` doesn't require the full fname. This function operates like `_make_bids_fname` and can "infer" the kind and the extension by searching down the bids_root. 

We can either:
1. Modify this to incorporate a full path at all times? Not sure what this should look like though.
2. Turn this into a private function?
3. Add some further edits to this function, but keeping the header as is?

jasmainak: @adam2392 let's discuss this in a new issue. We have spent too long on this PR. Let's merge this.

jasmainak:
Thanks heaps @adam2392 for the PR!

mne_bids/utils.py
adam2392: Not really sure how to articulate the issue you're thinking of... :| can you make it?

agramfort: reading this again I think @jasmainak has a point here. The fact that get_bids_fname returns BIDSPath and not a str is a code smell. It suggests that BIDSPath can be informed by bids_root and kind.

wdyt?

adam2392: Ah yeah I see; it does seem like it could be confusing. So would the suggested fix be to:

1. make it return a str?
or
2. add bids_root/kind to BIDSPath explicitly?

agramfort:
what would be the pros and cons of either options?

>

mne_bids/utils.py
jasmainak: started a separate issue @adam2392

PR #489, opened 2020-08-05
[MRG] Adding `status_description` to channels.tsv files defaulting to n/a for now

adam2392:
PR Description
--------------

Closes: #348 

Defaults to `n/a` for now to allow for easy grepping as @jasmainak suggested. Future improvements can be made that allow the status to change, pending an addition to mne-python, or implementation of #458 

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [x] All CIs are happy


hoechenberger:
I missed the discussion in #348, but since I've started working on #491 I'm wondering why you would want to fill it with `n/a` instead of empty strings?That's what I'm doing in #491 and I feel that that's the "better" way, because … well I feel that this string field should be empty if there's no explicit value set …

adam2392:
> I missed the discussion in #348, but since I've started working on #491 I'm wondering why you would want to fill it with `n/a` instead of empty strings?That's what I'm doing in #491 and I feel that that's the "better" way, because … well I feel that this string field should be empty if there's no explicit value set …

I think either value works, so can change it. That should pass the bids-validator right?

hoechenberger:
> I think either value works, so can change it.

Let's see what the others think :)

> That should pass the bids-validator right?

Hah, thanks for the reminder. I forgot to include BIDS validation in #491. Let's see…

hoechenberger:
> Hah, thanks for the reminder. I forgot to include BIDS validation in #491. Let's see…

Ok so this is resolved now… won't pass the validation with an empty cell, so I'm using `n/a` now in #491 too as a default! :)

hoechenberger:
@sappelhoff Feel free to merge if you're happy

agramfort:
+0 on my side. I did not follow discussion.

@sappelhoff @jasmainak ball is in your hands

sappelhoff:
thanks @adam2392

PR #508, opened 2020-08-20
only do a deep check of BIDSPath entities in make_bids_basename

agramfort:
required by study template that writes derivatives with original kinds etc.


hoechenberger:
Thx @agramfort!

PR #513, opened 2020-08-25
Remove BIDSPath.check attribute

hoechenberger:
PR Description
--------------
As pointed out by @sappelhoff in https://github.com/mne-tools/mne-bids/pull/511#discussion_r476601505,
the presence of the attribute could lead to unexpected behavior.

My solution is to just drop it and expect users to explicitly pass the `check` kwarg to `BIDSPath.update()` if they want to disable checking during an update.

Also requires the entities passed to `BIDSPatch.__init__()` to be keyword-only arguments. Thoughts?

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -67,15 +67,8 @@ class BIDSPath(object):
     extension : str | None
         The extension of the filename. E.g., ``'.json'``.
     check : bool
-        If True enforces the entities to be valid according to the
-        current BIDS standard. Defaults to True.
-
-    Attributes
-    ----------
-    check : bool
-        If True enforces the entities to be valid according to the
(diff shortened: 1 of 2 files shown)


hoechenberger:
@adam2392 Mind having a look?

agramfort:
You are reverting what I pushed this morning. It was @adam2392 original proposal. As is you don’t know if à bids path is valid or not without calling a private function and you need to pass check at every update 

What do others think ?

hoechenberger:
@agramfort You're right, I went for a walk in the park and came back to this PR, thinking: umm, let's not do this? ;) So… Closing!

PR #540, opened 2020-09-01
FIX : enforce str as param for BIDSPath

### LINKED ISSUE #538: mne_bids.BIDSPath throws cryptic error for non-string parameters
```
import mne_bids
bids_path = mne_bids.BIDSPath(subject=1)
```

Gives this stack trace

```
  File "/Users/alexrockhill/software/mne-bids/mne_bids/path.py", line 277, in __init__
    suffix=suffix, extension=extension)
  File "/Users/alexrockhill/software/mne-bids/mne_bids/path.py", line 514, in update
    _check_key_val(key, val)
  File "/Users/alexrockhill/software/mne-bids/mne_bids/utils.py", line 183, in _check_key_val
    if any(ii in val for ii in ['-', '_', '/']):
  File "/Users/alexrockhill/software/mne-bids/mne_bids/utils.py", line 183, in <genexpr>
    if any(ii in val for ii in ['-', '_', '/']):
TypeError: argument of type 'int' is not iterable
```

but should give something more informative or just cast to string.

agramfort:
closes #538

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -102,7 +102,7 @@ ALLOWED_PATH_ENTITIES_SHORT = {'sub': 'subject', 'ses': 'session',
                                'split': 'split', 'suffix': 'suffix'}
 
 # See: https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg  # noqa
-ENTITY_INDEX_OR_LABEL = {
+ENTITY_VALUE_TYPE = {
     'subject': 'label',
     'session': 'label',
     'task': 'label',
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -101,6 +101,21 @@ ALLOWED_PATH_ENTITIES_SHORT = {'sub': 'subject', 'ses': 'session',
                                'space': 'space', 'rec': 'recording',
                                'split': 'split', 'suffix': 'suffix'}
 
+# See: https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg  # noqa
+ENTITY_INDEX_OR_LABEL = {
+    'subject': 'label',
+    'session': 'label',
+    'task': 'label',
+    'run': 'index',
+    'processing': 'label',
+    'recording': 'label',
+    'space': 'label',
+    'acquisition': 'label',
+    'split': 'index',
+    'suffix': 'label',
+    'extension': 'label'
+}
+
 # accepted BIDS formats, which may be subject to change
 # depending on the specification
 BIDS_IEEG_COORDINATE_FRAMES = ['acpc', 'pixels', 'other']
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -19,7 +19,7 @@ from mne_bids.config import (
     ALLOWED_PATH_ENTITIES, ALLOWED_FILENAME_EXTENSIONS,
     ALLOWED_FILENAME_SUFFIX, ALLOWED_PATH_ENTITIES_SHORT,
     ALLOWED_DATATYPES, SUFFIX_TO_DATATYPE, ALLOWED_DATATYPE_EXTENSIONS,
-    reader, ENTITY_INDEX_OR_LABEL)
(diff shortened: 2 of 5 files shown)


sappelhoff (approved):

agramfort:
actually certain entities like split or maybe run or session could naturally be integers.

shall we accept for certain entities? is this standardized somewhere?

@sappelhoff @hoechenberger ?

sappelhoff:
> shall we accept for certain entities? is this standardized somewhere?

yes, it's standardized --> all *entitites* that are of format `entity_name-<index>` will have a nonnegative integer

however, the integers might be zero-padded, which would make it a string again.

see point 9 in https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#definitions

agramfort:
hum not sure I follow. sub or task can be arbitrary strings not just an
index.

are you suggesting that we should accept int everywhere?

>

hoechenberger:
> shall we accept for certain entities?

I would say for run and split we could cast to string, but not for the others.

sappelhoff:
> hum not sure I follow. sub or task can be arbitrary strings not just an
index.

some entities are specified as `entity_name-<label>`, others are `entity_name-<index>` ... there is a difference between `label` and `index`

point 9 and 10 in the list here: https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#definitions

> I would say for run and split we could cast to string, but not for the others.

I would say for all `index` type entities we can cast to str, but we'd have to check which ones they are (split and run for sure ...), I don't know that by heart.

agramfort:
actually I have my answer here

https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg

you see that run or split are <index> others are <label>

doing the changes

>

hoechenberger:
Super good catch, @agramfort! Eagle eye

mne_bids/config.py
```diff
@@ -101,6 +101,21 @@
                                'space': 'space', 'rec': 'recording',
                                'split': 'split', 'suffix': 'suffix'}
 
+# See: https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg  # noqa
+ENTITY_INDEX_OR_LABEL = {
```
hoechenberger, line 105:
```suggestion
ENTITY_VALUE_TYPE = {
```
maybe?

mne_bids/path.py
```diff
@@ -483,15 +483,33 @@ def update(self, *, check=None, **kwargs):
...
+            if ENTITY_INDEX_OR_LABEL[key] == 'label':
+                _validate_type(val, types=(None, str),
+                               item_name=key)
+            else:
+                assert ENTITY_INDEX_OR_LABEL[key] == 'index'
+                _validate_type(val, types=(int, str, None), item_name=key)
+                if isinstance(val, str) and not val.isdigit():
+                    raise ValueError(f'{key} is not an index (Got {val})')
+                elif isinstance(val, int):
+                    kwargs[key] = '{:02}'.format(val)
```
hoechenberger, line 512: Since this would only apply to `split` and `run`, if I'm not mistaken, I'm not sure we should zero-pad here at all, as but indices in most cases be <10, no?

sappelhoff: I thought the same, however *sometimes* they may be over 10 ... while they will be probably never > 100

so zero padding to length 2 covers *all* cases, at the expense that those with <10 will see zero-padded indices (I personally would not care about that)

so for me it's +1 to zero pad

hoechenberger: Ok, good point.

agramfort: this was already done this way. And it's also safe for numbers > 99

In [1]: '{:02}'.format(1)
Out[1]: '01'

In [2]: '{:02}'.format(100)
Out[2]: '100'

In [3]: '{:02}'.format(10)
Out[3]: '10'

hoechenberger (approved):
LGTM once CI goes green

hoechenberger:
… oh. it IS green already. @sappelhoff good to merge for you?

sappelhoff (approved):
nice, looks clean to me. Thx @agramfort

Issue #539, opened 2020-09-01
mne_bids.BIDSPath.fpath error


alexrockhill:
```
import mne
import mne_bids
import os.path as op
import numpy as np
out_dir = mne.utils._TempDir()
raw = mne.io.RawArray(np.random.random((1, 50000)), mne.create_info(['Ch1'], 500, ['eeg']))
raw.info['line_freq'] = 60
raw.save(op.join(out_dir, 'test-raw.fif'))
raw = mne.io.read_raw_fif(op.join(out_dir, 'test-raw.fif'))
base_path = mne_bids.BIDSPath(subject='1', root=op.join(out_dir, 'bids_dir'))
mne_bids.write_raw_bids(raw, base_path)
base_path.fdata
```

Gets the cryptic error 

```
RuntimeError: Found more than one matching data file for the requested recording. Cannot proceed due to the ambiguity. This is likely a problem with your BIDS dataset. Please run the BIDS validator on your data.
```

This does pass the validator and there was only one file saved so it probably shouldn't say more than one matching file found.

hoechenberger:
> This does pass the validator and there was only one file saved so it probably shouldn't say more than one matching file found.

Yes, the error message is misleading here.

May I ask what output you actually expected? It's not really clear to me from your MWE. Thanks!

adam2392:
Perhaps we could use `print_dir_tree` even in the error message here. I agree error message is noninformative here.

alexrockhill:
I was trying to get the directory for the BIDSPath. I'd like to use it to save a beh file as by joining `bids_dir`, `path.subsesdir` and `path.basename` + `beh.tsv|json`. I could make a PR when I have a bit of time later today

hoechenberger:
So what you would want is essentiall

> I was trying to get the directory for the BIDSPath. I'd like to use it to save a beh file as by joining `bids_dir`, `path.subsesdir` and `path.basename` + `beh.tsv|json`. I could make a PR when I have a bit of time later today

So you want the relative path **inside** the `bids_root`?

Also please note that `BIDSPath.basename` contains the extension already

alexrockhill:
Exactly!

hoechenberger:
I imagine you should be able to do:
```python
fpath = bids_path.copy().update(root='.').fpath
rel_path = fpath.parent if fpath.is_file() else fpath
```

hoechenberger:
[...]

PR #542, opened 2020-09-01
[MRG, BUG] bids_path updates

alexrockhill:
PR Description
--------------

The changes to `write` might be too much, but I thought I would put it as a suggestion. I started those because if you're putting in a `bids_path`, you should get it out (or modified in place) instead of the `bids_root` string. The other changes can be reversed if they're too much but I do like them because I think it will make debugging easier when there is only one instance of the a variable in the BIDSPath object field rather than a local instance and an object instance. This addresses https://github.com/mne-tools/mne-bids/issues/539 where the `bids_path` was not updated after writing out and so there was no suffix and thus no filepath returned. The small changes in `path` fixed that without all the other stuff in `write` though.

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -56,6 +56,7 @@ Bug
 - Ensure that ``Raw.info['bads']`` returned by :func:`mne_bids.read_raw_bids` is always a list, by `Richard Höchenberger`_ (`#501 <https://github.com/mne-tools/mne-bids/pull/501>`_)
 - :func:`mne_bids.write_raw_bids` now ensures that **all** parts of the :class:`mne.io.Raw` instance stay in sync when using anonymization to shift dates, e.g. ``raw.annotations``, by `Richard Höchenberger`_ (`#504 <https://github.com/mne-tools/mne-bids/pull/504>`_)
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
+- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned by `Alex Rockhill`_ (`#508 <https://github.com/mne-tools/mne-bids/pull/508>`_)
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -56,7 +56,7 @@ Bug
 - Ensure that ``Raw.info['bads']`` returned by :func:`mne_bids.read_raw_bids` is always a list, by `Richard Höchenberger`_ (`#501 <https://github.com/mne-tools/mne-bids/pull/501>`_)
 - :func:`mne_bids.write_raw_bids` now ensures that **all** parts of the :class:`mne.io.Raw` instance stay in sync when using anonymization to shift dates, e.g. ``raw.annotations``, by `Richard Höchenberger`_ (`#504 <https://github.com/mne-tools/mne-bids/pull/504>`_)
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
-- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned by `Alex Rockhill`_ (`#508 <https://github.com/mne-tools/mne-bids/pull/508>`_)
+- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -56,7 +56,7 @@ Bug
 - Ensure that ``Raw.info['bads']`` returned by :func:`mne_bids.read_raw_bids` is always a list, by `Richard Höchenberger`_ (`#501 <https://github.com/mne-tools/mne-bids/pull/501>`_)
 - :func:`mne_bids.write_raw_bids` now ensures that **all** parts of the :class:`mne.io.Raw` instance stay in sync when using anonymization to shift dates, e.g. ``raw.annotations``, by `Richard Höchenberger`_ (`#504 <https://github.com/mne-tools/mne-bids/pull/504>`_)
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
-- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
+- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned, by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
 
 API
 ~~~
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -387,7 +387,7 @@ class BIDSPath(object):
                     _get_matching_bidspaths_from_filesystem(self)
 
                 # FIXME This will break
-                # FIXME e.g. with FIFF data split across multiple FIXME files.
+                # FIXME e.g. with FIFF data split across multiple files.
                 # if extension is not specified and no unique file path
                 # return filepath of the actual dataset for MEG/EEG/iEEG data
                 if self.suffix is None or self.suffix in ALLOWED_DATATYPES:
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -360,10 +360,7 @@ class BIDSPath(object):
         """
         # create the data path based on entities available
         # bids_root, subject, session and suffix
-        if self.root is not None:
-            data_path = self.root
-        else:
-            data_path = ''
+        data_path = '' if self.root is None else self.root
         if self.subject is not None:
             data_path = op.join(data_path, f'sub-{self.subject}')
         if self.session is not None:
@@ -393,7 +390,7 @@ class BIDSPath(object):
                 # FIXME e.g. with FIFF data split across multiple FIXME files.
                 # if extension is not specified and no unique file path
                 # return filepath of the actual dataset for MEG/EEG/iEEG data
-                if self.suffix in ALLOWED_DATATYPES:
+                if self.suffix is None or self.suffix in ALLOWED_DATATYPES:
                     # now only use valid datatype extension
                     valid_exts = sum(ALLOWED_DATATYPE_EXTENSIONS.values(), [])
                     matching_paths = [p for p in matching_paths
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -1113,8 +1113,7 @@ def test_set(_bids_validate):
     # is broken for earlier versions
     events_tsv_fname = op.join(bids_root, 'sub-' + subject_id,
                                'ses-' + session_id, 'eeg',
-                               bids_path.basename.stem.replace(
-                                   'eeg', 'events.tsv'))
+                               bids_path.basename + '_events.tsv')
     if check_version('mne', '0.18'):
         assert op.exists(events_tsv_fname)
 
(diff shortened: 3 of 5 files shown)


adam2392:
+1 Yeah I think returning the path to the written file is sensible to me. Returning the bids_root seems redundant.

mne_bids/write.py
```diff
@@ -963,20 +964,15 @@ def write_raw_bids(raw, bids_path, events_data=None,
                        "raw.times should not have changed since reading"
                        " in from the file. It may have been cropped.")
 
-    bids_path = bids_path.copy()
```
adam2392, line 966: I would still make a copy and return a modified copy because this will cause undesired behavior if you have a bids_path to start, call `write_raw_bids` without needing/wanting to use the output fpath, and then expect to use the original bids_path again.

alexrockhill:
@adam2392 So one of the things that's crashing the tests is that `bids_path.basename` gives the extension and the suffix as well. I thought that was only supposed to have the base part (e.g. `sub-1_task-test`) excluding any extensions. To my memory that was how you used to pass in `basepath` before `bids_path` became a class and that's a common BIDS reference is that the basepath doesn't have the extension. cc @jasmainak

mne_bids/write.py
alexrockhill: A lot of mne functions modify in place without an option... I agree that this may be an issue but I wrote is that way to be consistent e.g. https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.set_annotations

adam2392:
> @adam2392 So one of the things that's crashing the tests is that `bids_path.basename` gives the extension and the suffix as well. I thought that was only supposed to have the base part (e.g. `sub-1_task-test`) excluding any extensions. To my memory that was how you used to pass in `basepath` before `bids_path` became a class and that's a common BIDS reference is that the basepath doesn't have the extension. cc @jasmainak

The `basename` property was done in the style with `pathlib.Path`, so that can be adapted as seen fit. I'm for including suffix/extension since it makes it more explicit and in line with `op.basename(...)`. 

https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-name-structure

Should we have a property for the entities without the suffix and extension though?

mne_bids/write.py
alexrockhill: Also https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.drop_channels, returns the modified instance but also acts in place as far as I remember. I always found that annoying because if you return the modified instance, to me you shouldn't modify in place as well.

My preference might be to return nothing an just modify the path instance and make that clear to the user that their path is updated and so to pass a copy if needed.

alexrockhill:
I'm not sure what the best way is but I remember in a previous version of mne_bids, you didn't have to pass the suffix and extension in basepath

mne_bids/write.py
adam2392: I suppose @agramfort and @hoechenberger can chime in here?

If we want to modify in place, then a lot of tests will need to pass in the copied version instead (that was the difficulty I found, so I went w/ making a copy :p).

mne_bids/write.py
```diff
@@ -851,7 +851,8 @@ def write_raw_bids(raw, bids_path, events_data=None,
             participants.tsv
             scans.tsv
 
-        Note that the datatype ``'meg'`` is automatically inferred from the raw
+        Note that ``bids_path`` will be modified in place as well as returned
+        and that the datatype ``'meg'`` is automatically inferred from the raw
```
agramfort, line 855: write_raw_bids does not only right fif files. it can also write EEG. I feel this needs to be rephrased.

alexrockhill:
You know if we modified `BIDSPath` in `write_raw_bids`, then `BIDSPath.fpath` could be a field instead of a function and it wouldn't have to parse anything or guess at files.

IMO it would be best to do a little bit bigger rewrite and have `BIDSPath` have fields for `electrodes_path`, `scans_path` and the other paths that are being copied in `write_raw_bids`. My conceptualization of `BIDSPath` is one object that stores all the path information for a particular data file but right now the sidecar files are not explicitly linked in the 'BIDSPath` object and I think that's a bit of a problem.

mne_bids/write.py
agramfort: @alexrockhill MNE function should not operate inplace. It's class method that do. Please avoid edge effects by modifying input parameters in functions if you can.

mne_bids/tests/test_write.py
```diff
@@ -1400,16 +1400,16 @@ def test_write_raw_pathlike():
     raw = _read_raw_fif(raw_fname)
 
     bids_root = Path(_TempDir())
-    events_fname = (Path(data_path) / 'MEG' / 'sample' /
-                    'sample_audvis_trunc_raw-eve.fif')
+    events_fname = Path(data_path) / 'MEG' / 'sample' /\
+        'sample_audvis_trunc_raw-eve.fif'
```
agramfort, line 1404: please avoid cosmetic changes

mne_bids/write.py
```diff
@@ -1001,157 +998,155 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
 
-    if datatype == 'meg' and convert and not anonymize:
+    if bids_path.datatype == 'meg' and convert and not anonymize:
         raise ValueError(f"Got file extension {convert} for MEG data, "
                          f"expected one of "
                          f"{ALLOWED_DATATYPE_EXTENSIONS['meg']}")
 
     if not convert and verbose:
-        print('Copying data files to %s' % op.splitext(bids_fname)[0])
+        print('Copying data files to %s' % op.splitext(bids_path.fpath)[0])
```
agramfort, line 1107: I feel the splitext is unnecessary here.

mne_bids/write.py
```diff
@@ -1001,157 +998,155 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
 
     # File saving branching logic
     if convert:
-        if datatype == 'meg':
-            if ext == '.pdf':
-                bids_fname = op.join(data_path, op.basename(bids_fname))
-            _write_raw_fif(raw, bids_fname)
+        if bids_path.datatype == 'meg':
+            _write_raw_fif(
+                raw, (op.join(data_path, op.basename(bids_path.fpath))
```
agramfort, line 1113: we should ideally adopt Path way of doing things and not rely on os.path 

maybe for later.

```suggestion
                raw, data_path / bids_path.fpath.name
```

agramfort (commented):
I like the new way of doing things !

just a few remarks

agramfort:
I am -1 on adding attributes for every thing. But we can image a method for
getting sidecar files of a BIDSPath.

so far I can live with copy().update(suffix='electrodes', extenstion='.tsv')

>

mne_bids/write.py
alexrockhill: Sounds good, thanks for the clarification and good call @adam2392

alexrockhill:
> I am -1 on adding attributes for every thing. But we can image a method for getting sidecar files of a BIDSPath. so far I can live with copy().update(suffix='electrodes', extenstion='.tsv')
> […](#)

What about `branch` and 'get_branch` functions to wrap the `copy().update(suffix='electrodes', extenstion='.tsv')` code? The dual purpose of that would be my issue from earlier that if I wanted to use `BIDSPath` to add a behavior file, I could just call branch.

agramfort:
I am sure you can live with copy().update

I prefer we have one way of doing things. The more you complexify the API the
more you need documentation and tests etc.

$ python -c "import this" | grep "only one"

alexrockhill:
Yeah now that I think about it the fields wouldn't work because the `bids_path` would have to be saved out, you're right @agramfort, I'll leave it be.

mne_bids/tests/test_write.py
alexrockhill: My pylinter said it was causing a pep error so I was trying to change it so that it's compatible with the latest version

agramfort:
arfff.... no big deal.

>

agramfort:
@alexrockhill set the PR to MRG when you think it's good to go.

doc/whats_new.rst
```diff
@@ -56,6 +56,7 @@ Bug
 - Ensure that ``Raw.info['bads']`` returned by :func:`mne_bids.read_raw_bids` is always a list, by `Richard Höchenberger`_ (`#501 <https://github.com/mne-tools/mne-bids/pull/501>`_)
 - :func:`mne_bids.write_raw_bids` now ensures that **all** parts of the :class:`mne.io.Raw` instance stay in sync when using anonymization to shift dates, e.g. ``raw.annotations``, by `Richard Höchenberger`_ (`#504 <https://github.com/mne-tools/mne-bids/pull/504>`_)
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
+- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
```
hoechenberger, line 59:
```suggestion
- Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned, by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
```

mne_bids/path.py
```diff
@@ -393,7 +390,7 @@ def fpath(self):
                 # FIXME e.g. with FIFF data split across multiple FIXME files.
```
hoechenberger, line 390:
```suggestion
                # FIXME e.g. with FIFF data split across multiple files.
```

mne_bids/tests/test_write.py
```diff
@@ -1400,16 +1401,16 @@ def test_write_raw_pathlike():
     raw = _read_raw_fif(raw_fname)
 
     bids_root = Path(_TempDir())
-    events_fname = (Path(data_path) / 'MEG' / 'sample' /
-                    'sample_audvis_trunc_raw-eve.fif')
+    events_fname = \
+        Path(data_path) / 'MEG' / 'sample' / 'sample_audvis_trunc_raw-eve.fif'
```
hoechenberger, line 1405: Uff, why the backslash :) Wasn't it prettier before?

mne_bids/write.py
```diff
@@ -851,8 +851,9 @@ def write_raw_bids(raw, bids_path, events_data=None,
             participants.tsv
             scans.tsv
 
-        Note that the datatype ``'meg'`` is automatically inferred from the raw
-        object and extension ``'.fif'`` is copied from ``raw.filenames``.
+        Note that the datatype is automatically inferred from the raw
+        object as well as the extension. Data with meg and other
+        electrophysiology data in the same file will be stored as meg.
```
hoechenberger, line 856:
```suggestion
        Note that the data type is automatically inferred from the raw
        object, as well as the extension. Data with MEG and other
        electrophysiology data in the same file will be stored as ``'meg'``.
```

mne_bids/write.py
```diff
@@ -899,8 +900,9 @@ def write_raw_bids(raw, bids_path, events_data=None,
 
     Returns
     -------
-    bids_root : str
-        The path of the root of the BIDS-compatible folder.
+    bids_path : BIDSPath
+        The modified path to the file written, including the root of the
+        BIDS-compatible folder under `bids_path.root`
```
hoechenberger, line 905:
```suggestion
        BIDS-compatible folder in ``bids_path.root``.
```

mne_bids/write.py
```diff
@@ -938,8 +940,7 @@ def write_raw_bids(raw, bids_path, events_data=None,
                            'instantiate using mne_bids.BIDSPath().')
 
     # check root available
```
hoechenberger, line 942:
```suggestion
    # Check if the root is available
```

mne_bids/write.py
```diff
@@ -1001,157 +999,155 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
-                            raw, datatype, overwrite, verbose)
+            _write_dig_bids(electrodes_path.fpath, coordsystem_path.fpath,
+                            bids_path.root, raw, bids_path.datatype,
+                            overwrite, verbose)
     else:
-        logger.warning(f'Writing of electrodes.tsv '
-                       f'is not supported for datatype "{datatype}". '
-                       f'Skipping ...')
+        logger.warning(f'Writing of electrodes.tsv is not supported '
+                       f'for datatype "{bids_path.datatype}". Skipping ...')
```
hoechenberger, line 1075:
```suggestion
                       f'for data type "{bids_path.datatype}". Skipping ...')
```

mne_bids/tests/test_write.py
alexrockhill: No one likes the change to this huh? :) It failed by pylinter because of a line break after a binary operator. I thought pep8 was getting rid of those so we should change it

agramfort:
use flake8 like all of us :)

>

mne_bids/tests/test_write.py
hoechenberger: Nope, ASAIK this is valid PEP8. Also the operator is overloaded here anyway, sooooo... ;)

alexrockhill:
Not sure what's going on with the bti paths because the directories are different, trying to figure it out one sec

hoechenberger:
> Not sure what's going on with the bti paths because the directories are different, trying to figure it out one sec

Yes it's annoying :( good luck!

alexrockhill:
Ok there was just a style error, once the CIs pass, I think it should be good to go

agramfort (approved):
@hoechenberger merge if happy and green

thx

hoechenberger (approved):

hoechenberger:
Thank you, @alexrockhill!

Issue #536, opened 2020-08-29
make_bids_folders() should become a method of BIDSPath


hoechenberger:
Current signature:
https://github.com/mne-tools/mne-bids/blob/1a52b18d87e09b18b821287180da4e235acb939e/mne_bids/path.py#L726-L727

The function could easily changed to accept a `BIDSPath` instead of `subject`, `session`, `datatype`, and `bids_root`. In the next step, it should become a method of `BIDSPath`. I was thinking something like `BIDSPath.mkdir()`.

Thoughts?

adam2392:
+1 for this. esp since a loto f the functionality comes from `Path.mkdir(overwrite=True, exist_ok=True, etc.)`.

hoechenberger:
The `make_dir=False` functionality should be removed, and instead we should expose sth like `BIDSPath.fpath.parent` as `BIDSPath.directory` or something like that

adam2392:
> The `make_dir=False` functionality should be removed, and instead we should expose sth like `BIDSPath.fpath.parent` as `BIDSPath.directory`

down for this, since `pathlib` is more general

Issue #527, opened 2020-08-27
Drop make_bids_folders()


hoechenberger:
`make_bids_folders()` can safely be dropped if we make it convenient to access the folder structure(s) via `BIDSPath`. I imagine we have an attribute `BIDSPath.folder` that essentially returns `BIDSPath.fpath.parent`

```python
bids_path = BIDSPath(subject='foo', session='bar', task='baz', modality='meg')
subject_folder = bids_path.folder  # returns a Path
if not subject_folder.exists():
    subject_folder.mkdir(parents=True)
```
(one could also omit the `if` line and call `subject_folder.mkdir()` without a prior check)

jasmainak:
how would this work for "incomplete" `BIDSPath` ?

hoechenberger:
We would need at least subject and modality. I believe this is the only restriction?

--

jasmainak:
hummm right, I think so!

adam2392:
> We would need at least subject and modality. I believe this is the only restriction?

I would say at least subject because if you want to go to the scans file, you might not need modality.

hoechenberger:
> I would say at least subject because if you want to go to the scans file, you might not need modality. 
￼￼
And if you only want to go to the root dir, and the root shall be a relative path -- you don't need anything at all 😅

agramfort:
we can try to see if this works but this is going the right direction I think

PR #543, opened 2020-09-01
[MRG, API] Add mkdir function and directory property to BIDSPath

adam2392:
PR Description
--------------

Closes: #536 
Closes: #527 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


adam2392:
I haven't implemented unit tests yet.

@hoechenberger is this sort of in line w/ where you were thinking for #536?

mne_bids/path.py
```diff
@@ -312,6 +312,23 @@ def basename(self):
         basename = '_'.join(basename)
         return basename
 
+    @property
+    def directory(self):
+        # create the data path based on entities available
+        # bids_root, subject, session and suffix
```
hoechenberger, line 318: Where do you use `suffix` here?

mne_bids/path.py
```diff
@@ -312,6 +312,23 @@ def basename(self):
...
+        if self.root is not None:
+            data_path = self.root
+        else:
+            data_path = ''
+        if self.subject is not None:
+            data_path = op.join(data_path, f'sub-{self.subject}')
+        if self.session is not None:
+            data_path = op.join(data_path, f'ses-{self.session}')
+        # file-suffix will allow 'meg', 'eeg', 'ieeg', 'anat'
```
hoechenberger, line 327: What does that mean here?

mne_bids/path.py
```diff
@@ -345,6 +362,17 @@ def copy(self):
         """
         return deepcopy(self)
 
+    def mkdir(self, exist_ok=True, parents=True):
```
hoechenberger, line 365:
```suggestion
    def mkdir(self, exist_ok=True, parents=True):
```
Do we even need `exist_ok`?

Also I would propose to rename `parents` to `create_parents`, WDYT?

hoechenberger:
> @hoechenberger is this sort of in line w/ where you were thinking for #536?

Yep! But I'm just about to go to bed, will look again after a good night's sleep :)

mne_bids/path.py
adam2392: > Do we even need `exist_ok`?

I suppose no. Is there any reason we shouldn't always pass exist_ok=True? I'm game to remove it.

> Also I would propose to rename `parents` to `create_parents`, WDYT?

Hmm I was keeping in line with `pathlib.Path`, but think `create_parents` is more explicit. I'm okay with either one, but I would prefer keeping in line w/ `pathlib` semantics to make it easier to remember.

adam2392: Lmk what you think.

doc/whats_new.rst
```diff
@@ -76,10 +76,11 @@ API
 - :code:`mne_bids.make_bids_basename` has been removed. Use :class:`mne_bids.BIDSPath` directly, by `Adam Li`_ (`#511 <https://github.com/mne-tools/mne-bids/pull/511>`_)
 - Add ``check`` parameter and attribute to :class:`mne_bids.BIDSPath` that allows users to turn off entity checks by `Adam Li`_ (`#511 <https://github.com/mne-tools/mne-bids/pull/511>`_)
 - Add ``datatype`` parameter and attribute to :class:`mne_bids.BIDSPath` that allows users to specify EEG, MEG, or iEEG datasets by `Adam Li`_ (`#514 <https://github.com/mne-tools/mne-bids/pull/514>`_)
-- Add ``datatype`` to replace ``kind`` parameter to :func:`mne_bids.make_bids_folders` and :func:`mne_bids.read_raw_bids` that allows users to specify EEG, MEG, or iEEG datasets by `Adam Li`_ (`#514 <https://github.com/mne-tools/mne-bids/pull/514>`_)
+- Add ``datatype`` to replace ``kind`` parameter to ``mne_bids.make_bids_folders`` and :func:`mne_bids.read_raw_bids` that allows users to specify EEG, MEG, or iEEG datasets by `Adam Li`_ (`#514 <https://github.com/mne-tools/mne-bids/pull/514>`_)
```
adam2392, line 79:
```suggestion
```

adam2392:
Codecov goes down because of `_mkdir_p`, which was used in `make_bids_folder`, but now I can't see any reason that we would want to overwrite? If so, then the `pathlib.Path.mkdir(exist_ok=True, parents=True)` seems easier to use, which would justify the codecov change.

mne_bids/path.py
```diff
@@ -215,6 +212,8 @@ class BIDSPath(object):
         The basename of the file path. Similar to `os.path.basename(fpath)`.
     root : str
         The root of the BIDS path.
+    directory : pathlib.Path
```
agramfort, line 215: why do we need this?

bids_path.fpath.parent ?

mne_bids/path.py
agramfort: please stick to pathlib names as is right now

agramfort: I would just do:

mkdir(parents=False, exist_ok=False)

for follow the param order from pathlib.Path

mne_bids/path.py
```diff
@@ -345,6 +378,27 @@ def copy(self):
...
+             (default is True).
+
+        Returns
+        -------
+        bids_path : pathlib.Path
+            The created directory.
+        """
+        bids_path = Path(self.directory)
+        bids_path.mkdir(exist_ok=exist_ok, parents=parents)
+        return bids_path
```
agramfort, line 400: I would return self. You can always get the directory with .fpath.parent

agramfort: i would allow you to chain operations

mne_bids/path.py
hoechenberger: @agramfort 
> please stick to pathlib names as is right now
> mkdir(parents=False, exist_ok=False)
> 
> for follow the param order from pathlib.Path

I don't think we'll need `parents`, no?

agramfort:
true we could even skip the parents param as it's like to be true all the
time

>

mne_bids/path.py
hoechenberger: Hmm is that really a common use case, chaining with `mkdir`?

`bids_path.mkdir()....` how would one proceed after that? Typically, one has "massaged" BIDSPath with all the params one needs before calling `mkdir`, no? I guess it could be more convenient to return the created dir:

```python
dir_ = bids_path.mkdir()
dir_.rglob(...)
```

Not saying that I have a very strong opinion here, just brainstorming.

hoechenberger: > ```python
> bids_path.mkdir().rglob(...)
> ```

Then again, who'd want to glob an empty dir *LOL*

hoechenberger:
> true we could even skip the parents param as it's like to be true all the time

Yep!

And regarding `exist_ok` I'm also inclined to just drop it. Idk? Is there any situation we would **want** to raise if the directory already exists? Doesn't this do the trick?
```python
if not bids_path.directory.exists():
    bids_path.mkdir()
```
Just trying to keep the API slim!

agramfort:
I would keep exist_ok param like pathlib. It's still simpler.

hoechenberger:
> I would keep exist_ok param like pathlib. It's still simpler.

Ok with me to keep this functionality.

I'd like to argue, though, that in MNE and in the study-template we would probably call this param something like `on_exists='raise'`.

mne_bids/path.py
adam2392: So since fpath sometimes tries to infer the full filepath, then getting its parent is simply the directory `<root>/sub-<sub>/ses-<ses>/<datatype>`, but if the fpath errors out, then you can't get its parent. This might happen if you don't add enough entities, and it is unparseable. 

I don't think we've settled on the best way to handle that case, but for now, if we expose directory, then this would solve the issue? 

WDYT @hoechenberger

mne_bids/path.py
adam2392: This outcome might  depend on how we resolve https://github.com/mne-tools/mne-bids/pull/543#discussion_r482297399

agramfort: I would return self anyhow. You can do .mkdir().directory

now would:

Path(str(bids_path)).parent do the job?

basically I want to keep the public API simple to start with and see
if certain things become recurrent requests.

mne_bids/path.py
hoechenberger: Shouldn't we prevent the construction of a `BIDSPath` that lacks all the info we require to construct a sensible path in the first place?

The following
```
BIDSPath(root='/foo/', subject='01')
```
should probably error out on instantiation / update (subject supplied, but data type missing -> one directory level remains unspecified), unless one explicitly requests the operation via a special kwarg (`force=True` or something).

mne_bids/path.py
adam2392: I guess we can always make the parents. I don't see a reason why not(?) I'll adjust it.

mne_bids/path.py
adam2392: Well, what if someone wants to create the subjects folder? Then it's sensible they would want to do:

```Python
bids_path = BIDSPath(root='/foo/', subject='01')
bids_path.mkdir()
subj_path = bids_path.directory

# errors out if you have multiple runs for subject '01'
bids_path.fpath()
```

adam2392:
So right now, I tentatively added `directory` and `mkdir` to `BIDSPath`, which would operate like this:

```Python
bids_path = BIDSPath(root='/foo/', subject='01')
bids_path.mkdir()
subj_path = bids_path.directory

# errors out if you have multiple runs for subject '01'
bids_path.fpath()
```

I think fpath is a bit troublesome because it tries to determine by itself when "to infer" the file path (e.g. if missing `extension`) vs when it just gives you the `op.join(directory, basename)` path. Should we add a `infer=True` kwarg to fpath to allow it to be explicit? If we do so, then the above code might be able to look like this:

```Python
bids_path = BIDSPath(root='/foo/', subject='01')
bids_path.mkdir()

# option 1 (currently)
subj_path = bids_path.directory

# option 2 
subj_path = bids_path.fpath(infer=False).parent
```

**A comment on "inferring" the full filename:**
I think even though the "inference" results in some level of difficulty for us, it's perfectly valid that we try to infer the extension considering we're operating in BIDS world and the entities should be enough to uniquely identify a file.

mne_bids/path.py
```diff
@@ -345,6 +375,27 @@ def copy(self):
...
+            already exists (default is True).
+
+        Returns
+        -------
+        bids_path : pathlib.Path
+            The created directory.
+        """
+        bids_path = Path(self.directory)
+        bids_path.mkdir(parents=parents, exist_ok=exist_ok)
+        return bids_path
```
agramfort, line 397:
```suggestion
        self.directory.mkdir(parents=parents, exist_ok=exist_ok)
        return self
```

mne_bids/read.py
```diff
@@ -341,8 +340,7 @@ def read_raw_bids(bids_path, extra_params=None, verbose=True):
                                    sub=sub, ses=ses)
     bids_path.update(datatype=datatype, suffix=datatype)
 
-    data_dir = make_bids_folders(subject=sub, session=ses, datatype=datatype,
-                                 make_dir=False)
+    data_dir = bids_path.mkdir()
```
agramfort, line 343:
```suggestion
    data_dir = bids_path.mkdir().directory
```

mne_bids/read.py
```diff
@@ -341,8 +340,7 @@ def read_raw_bids(bids_path, extra_params=None, verbose=True):
...
     bids_fname = op.basename(bids_path.fpath)
```
agramfort, line 344:
```suggestion
    bids_fname = bids_path.fpath.name
```

agramfort (commented):
ok I got used to directory attribute.

I clearly prefer it to solution 2

mne_bids/path.py
```diff
@@ -392,9 +392,8 @@ def mkdir(self, parents=True, exist_ok=True):
         bids_path : pathlib.Path
             The created directory.
```
agramfort, line 393: fix docstring :)

agramfort (approved):

mne_bids/path.py
```diff
@@ -312,6 +313,35 @@ def basename(self):
...
 
+    @property
+    def directory(self):
+        """Get the BIDS parent directory.
+
+        If ``subject``, ``session`` and ``datatype`` are
+        set then they will be used to construct the
+        directory. For example if ``subject='01'``,
+        ``session='02'`` and ``datatype='ieeg'``, then
+        directory would be::
```
hoechenberger, line 324:
```suggestion
        If ``subject``, ``session`` and ``datatype`` are set, then they will be
        used to construct the directory location. For example, if
        ``subject='01'``, ``session='02'`` and ``datatype='ieeg'``, then the
        directory would be::
```

mne_bids/path.py
```diff
@@ -312,6 +313,35 @@ def basename(self):
...
+
+            <root>/sub-01/ses-02/ieeg
+
+        Returns
+        -------
+        data_path : pathlib.Path
+            The directory path of a BIDS-compliant dataset.
```
hoechenberger, line 331:
```suggestion
            The path of the BIDS directory.
```

mne_bids/path.py
```diff
@@ -312,6 +313,35 @@ def basename(self):
...
+        """
+        # create the data path based on entities available
+        # bids_root, subject, session and datatype
```
hoechenberger, line 334:
```suggestion
        # Create the data path based on the available entities:
        # bids_root, subject, session, and datatype
```

mne_bids/path.py
```diff
@@ -345,6 +375,23 @@ def copy(self):
         """
         return deepcopy(self)
 
+    def mkdir(self, exist_ok=True):
+        """Create the BIDS path directory.
```
hoechenberger, line 379:
```suggestion
        """Create the directory structure of the BIDS path.
```

mne_bids/path.py
```diff
@@ -345,6 +375,23 @@ def copy(self):
...
+
+        Parameters
+        ----------
+        exist_ok : bool
+            Whether it is okay that the directory
+            already exists (default is True).
```
hoechenberger, line 385:
```suggestion
        exist_ok : bool
            If ``False``, raise an exception if the directory already exists.
            Otherwise, do nothing (default).
```

hoechenberger (commented):
Small nitpicks, will resolve that momentarily

hoechenberger (approved):

PR #548, opened 2020-09-06
[MRG] Fix raw reading

hoechenberger:
PR Description
--------------

I couldn't read files anymore using `read_raw_bids`. 

Reason was that we now construct the directory location via `bids_path.directory`, which includes the
BIDS root; yet we were still joining that path with the root, yielding invalid paths (i.e., the root would essentially appear twice at the beginning of the path).

Now, our CI didn't catch it for the simple reason that in our tests were essentially joining a string (root) and a `Path` (`BIDSPath.directory`) via `os.path.join`, exposing an interesting behavior of that function:

```python
import os
import pathlib

path_str = '/foo/bar'
path = pathlib.Path(path_str)

print(os.path.join(path_str, path))
```
What do you think would it produce?

**WRONG**, it would produce:
```
/foo/bar
```

So in this PR:

- We construct the path of the file to be read in `read_raw_bids` correctly by calling `bids_fpath = op.join(data_dir, bids_fname)` instead of `bids_fpath = op.join(bids_root, data_dir, bids_fname)`
- We remove a small bug (or at least an unnecessary call) in `read_raw_bids`, which was that we would try to create the folder structure when trying to retrieve the directory path (i.e., `bids_path.mkdir().directory` where `bids_path.directory` should suffice)
- We turn `BIDSPath.root` into a `pathlib.Path` which will avoid these kinds of silent bugs in the future; this also leads to a more consistent API

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -192,7 +192,7 @@ class BIDSPath(object):
         The "data type" of folder being created at the end of the folder
         hierarchy. E.g., ``'anat'``, ``'func'``, ``'eeg'``, ``'meg'``,
         ``'ieeg'``, etc.
-    root : str | None
+    root : str | pathlib.Path | None
         The root for the filename to be created. E.g., a path to the folder
         in which you wish to create a file with this name.
     check : bool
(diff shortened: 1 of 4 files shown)


agramfort (approved):

agramfort:
Thx @hoechenberger

PR #549, opened 2020-09-07
add support for CTF gradient compensated data

agramfort:
see https://github.com/mne-tools/mne-python/blob/master/mne/io/ctf/info.py#L235

cc @hoechenberger 

it fixes it but I don't understand why it's done this way in mne-python.

maybe @larsoner can help


hoechenberger (approved):
This fixes my problem with our CTF dataset, thanks!

hoechenberger:
Whoops. I didn't mean to merge this, actually. 💩

PR #559, opened 2020-09-24
MRG: BF: Use persyst reader func and not return value

hoechenberger:
PR Description
--------------

Use persyst reader function instead of the return value of that function to read persyst data.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


hoechenberger:
Obviously we don't have test coverage for this, otherwise we would've caught this… @adam2392 Do you think you could add a test?

hoechenberger:
I should drink my coffee before filing PRs. This is alright. Whoops.

PR #561, opened 2020-09-24
[MRG] Add MNE-BIDS Inspector

hoechenberger:
PR Description
--------------

This PR adds a first draft of an MNE-BIDS "Inspector", housed in `mne_bids/inspect.py`. The aim here is to create a way for users to conveniently browse and annotate BIDS raw data. A common theme is marking channels as bad based on notes you took during an electrophysiological recording. We (@agramfort  and I) decided to encourage our collaborators to convert their data to BIDS ASAP after the recording. That way, we can develop tools that will ideally work seamlessly across different labs.

So which functionality does this PR add, specifically?

- adds `mne_bids.inspect()` function and `mne_bids inspect` CLI command
- allows browsing of MEG and EEG raw data through `mne.Raw.plot()`; users may alter the channel status through clicking on the traces (mark / unmark as bad)
- when the browse window is being closed, users are asked whether they want to keep the changes; if they do, modify the `channels.tsv` file.

See this video demo:

https://www.youtube.com/watch?v=6L7jrcTUPGM

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


mne_bids/inspect.py
```diff
@@ -0,0 +1,128 @@
...
+    descriptions = []
+    for ch_name in bads:
+        idx = channels_tsv_data['name'].index(ch_name)
+        if channels_tsv_data['status'][idx] == 'bad':
+            # Channel was already marked as bad, retain existing description.
+            description = channels_tsv_data['status_description'][idx]
+        else:
+            # Channel has been manually marked as bad during inspection, assign
+            # default description.
+            description = 'Manual inspection via MNE-BIDS'
```
adam2392, line 121: I wonder if we can go a step further and say by who using

`os.getenv('USER')`, which would work for everyone except Windows users(?).

hoechenberger: How about
```python
In [5]: Path(os.path.expanduser('~')).name
Out[5]: 'hoechenberger'
```
And we could even add a timestamp.

hoechenberger:
To run tests, we would need to add matplotlib and Qt to the CI. Before I do that, I'd like to hear your general feedback on this.

mne_bids/inspect.py
hoechenberger: … but maybe that's difficult in terms of users' privacy. We should at least not do that without very explicitly letting them know about it.

adam2392: I like the idea of a timestamp! Hmm I suppose, but perhaps `user` can be a kwarg of the function then, so that way it is explicit, else it defaults to the behavior you have here?

jasmainak:
Video looked good, I like it so far!

sappelhoff:
> Video looked good, I like it so far!

agreed :-) way to go to charm reviewers ahead of time

mne_bids/inspect.py
```diff
@@ -0,0 +1,128 @@
+from pathlib import Path
+import matplotlib
+import matplotlib.pyplot as plt
+from matplotlib.widgets import Button
+
+from mne_bids import read_raw_bids, mark_bad_channels
+from mne_bids.read import _from_tsv
+
+matplotlib.use('Qt5Agg')  # XXX Discuss!
```
adam2392, line 9: Did you try this out on a mac? JW.

I find some of the qt apps were always buggy on my mac, but that might've changed with qt5.

hoechenberger: > Did you try this out on a mac? JW.

Yes, I'm using a Mac. `Qt5Agg` is the commended backend for MNE.

What's "JW"?

adam2392: Oh okay awesome! 

Oh.. JW=just wondering. Sorry I abbreviate a lot heh. :p

agramfort, line 9: I would not do this. First I don't think it's mandatory. @cbrnr will tell you that it will work with macos backend :) and mne-bids will be working without matplotlib installed. matplotlib is not a mandatory dep of mne-python and it should not be for mne-bids

hoechenberger: > it will work with macos backend :)

Ok I can drop this line that uses the PyQt backend.



> mne-bids will be working without matplotlib installed. matplotlib is not a mandatory dep of mne-python and it should not be for mne-bids

Well that's all nice and such, but the inspector simply will not work without Matplotlib, there's no way around it. Either you have it, or you'll get an `ImportError`…

hoechenberger: I've now changed the backend selection such that it will only request the PyQt backend if it's installed, and otherwise use the default. Good?

3bba9cb

hoechenberger: I will rework the runtime dependency stuff shortly.

cbrnr: So apparently I'm already the macOS backend guy 🤣 - but yes, it does work and there's no need to use PyQt5!

mne_bids/inspect.py
```diff
@@ -1,12 +1,10 @@
 from pathlib import Path
-import matplotlib.pyplot as plt
-from matplotlib.widgets import Button
 
 from mne_bids import read_raw_bids, mark_bad_channels
 from mne_bids.read import _from_tsv
 
 
-def inspect(bids_path, verbose=None):
+def inspect_bids(bids_path, block=True, verbose=None):
```
agramfort, line 7: shall we call this inspect raw? unless you foresee inspection of more types of files?

.github/workflows/unit_tests.yml
```diff
@@ -79,7 +82,25 @@ jobs:
       if: "matrix.os == 'ubuntu-latest'"
       run: |
         pip install -U pandas
-    - name: Run pytest
+    - name: Run pytest (Linux)
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        export BIDS_VALIDATOR_VERSION=`bids-validator --version`
+        echo Using bids-validator $BIDS_VALIDATOR_VERSION
+        xvfb-run python -m pytest . --cov=mne_bids mne_bids/tests/ mne_bids/commands/tests/ --cov-report=xml --cov-config=setup.cfg --verbose --ignore mne-python
```
jasmainak, line 90: why do you need `xvfb` for matplotlib stuff? or is it for the code overage report?

.github/workflows/unit_tests.yml
```diff
@@ -26,6 +26,9 @@ jobs:
       FORCE_COLOR: true
     steps:
     - uses: actions/checkout@v2
+    - name: Install X11 on macOS
+      if: "matrix.os == 'macos-latest'"
+      run: brew cask install xquartz
```
jasmainak, line 31: nothing to update in contributing/installation instructions?

jasmainak:
let me know what commands I should run to test it locally on my machine. Happy to break it for you. Have a lot of experience in being a beta tester ;-)

hoechenberger:
> let me know what commands I should run to test it locally on my machine. Happy to break it for you. Have a lot of experience in being a beta tester ;-)

Haha thanks ;) It's already broken, but I'll poke you once I need help tearing everything apart ;)

mne_bids/inspect.py
hoechenberger: Yeah I thought we could also add an MRI viewer and such… WDYT?

hoechenberger:
So I'm running into a weird issue here.

What I'm doing is I'm hooking into the MNE Raw Browser Figure's `close_event` callback such that when the figure is being closed, a new figure – a dialog box – will pop up, asking if the changed bad channels selection shall be written to disk.

If I open this dialog box in a non-blocking fashion via `fig.show()`, the buttons are inactive (don't respond to hovering or clicking); however other events (closing the dialog window, or pressing a key) are still processed correctly.

If I open the dialog box via `plt.show(block=True)` instead, the buttons **do** work as expected, but the MNE Raw Browser won't close anymore.

I'm at my wits' end …

The following video demonstrated the behavior:

https://www.youtube.com/watch?v=OESz7Rb3ans

@drammock Have you seen anything like this before and know what the problem might be?


This is the code I used to run the inspector:

```python
import os.path as op
import tempfile
import mne

from mne_bids import BIDSPath, write_raw_bids, inspect_bids

# Write BIDS test data.
bids_root = tempfile.mktemp('mnebids_utils_test_bids_ds')
bids_path = BIDSPath(
    subject='01', task='test', datatype='meg', root=bids_root)

raw_fname = op.join(data_path, 'MEG', 'sample',
                    'sample_audvis_trunc_raw.fif')
raw.info['line_freq'] = 60
write_raw_bids(raw, bids_path=bids_path, overwrite=True)

# Run the MNE-BIDS Inspector.
fig = inspect_bids(bids_path, block=True)
```

Implementation of the inspector: https://github.com/mne-tools/mne-bids/blob/69ff5d257a379f1f56461ff496cdf345312f300b/mne_bids/inspect.py

drammock:
> @drammock Have you seen anything like this before and know what the problem might be?

Screenshot_2020-11-24_15-43-01

hoechenberger:
@drammock **THANKS** Cannot believe I spent hours trying to figure this out when the solution was right there in the docs…  😳

.github/workflows/unit_tests.yml
```diff
@@ -34,9 +34,9 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies using pip
+    - name: Install Python dependencies using pip
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip wheel
```
agramfort, line 39: why do you need wheel?

mne_bids/inspect.py
agramfort: if you want to go generic maybe you can do:

inspect_dataset(bids_path, ..., datatype=('meg', 'eeg'))

so we can decide how to inspect differently datatypes

here it would inspect all the files ending with _eeg or _meg

wdyt?

.github/workflows/unit_tests.yml
hoechenberger: The Matplotlib dependencies wouldn't install properly otherwise, if I remember correctly. Can try to remove wheel again and see if the issue persists

hoechenberger: Oh, no, so the thing was without wheel obviously we cannot install wheels, which makes things difficult with matplotlib.

mne_bids/inspect.py
hoechenberger: How would we go about multiple matches? Would we just iterate over all of them?

.github/workflows/unit_tests.yml
hoechenberger: I removed it, we're testing with the Agg backend now.

mne_bids/inspect.py
agramfort: yes first do meg then eeg etc.

hoechenberger: Currently the doctoring says:

```
    bids_path : BIDSPath
        A :class:`mne_bids.BIDSPath` containing enough information to uniquely
        identify the raw data file. If this ``BIDSPath`` is accepted by
        :func:`mne_bids.read_raw_bids`, it will work here.
```
I think this is fair… WDYT?

hoechenberger:
Ok I think the only things that are missing are:

- [ ] changelog entry
- [ ] example for the gallery
- [ ] updated install instructions

mne_bids/inspect.py
hoechenberger: Seems I missed your previous response, but I also thought about this again and now I agree, we should iterate over all matches and not force the user to specify a BIDSPath for each file individually. I will implement a solution.

mne_bids/commands/tests/test_cli.py
```diff
@@ -255,4 +263,37 @@ def test_count_events(tmpdir):
...
+                         root=bids_root)
+    write_raw_bids(raw, bids_path, overwrite=True, verbose=False)
+
+    import matplotlib
+    matplotlib.use('agg')
+
+    with ArgvSetter(('--bids_root', bids_root, '--subject', subject,
+                     '--task', task, '--type', datatype)):
+        with pytest.warns(RuntimeWarning, match='The unit for chann*'):
+            mne_bids_inspect.run()
```
agramfort, line 296: running `mne_bids inspect --bids_root BIDS` should be enough. Ie there should be default values None so I can just loop over all files with just mne_bids inspect --bids_root BIDS

hoechenberger: Yes, this was an oversight. Fixed in 6d5cf3c

hoechenberger:
Updated example:
https://3641-89170358-gh.circle-artifacts.com/0/dev/auto_examples/mark_bad_channels.html#interactive-use

hoechenberger:
New example:
https://3644-89170358-gh.circle-artifacts.com/0/dev/auto_examples/mark_bad_channels.html#interactive-use

Demonstrated use without and with frequency filter

hoechenberger:
@agramfort Merge if you're happy!

mne_bids/inspect.py
```diff
@@ -0,0 +1,215 @@
...
+
+from mne_bids import read_raw_bids, mark_bad_channels
+from mne_bids.read import _from_tsv
+from mne_bids.config import ALLOWED_DATATYPE_EXTENSIONS
+
+
+def inspect_dataset(bids_path, l_freq=None, h_freq=None, verbose=None):
+    """Inspect and annotate BIDS raw data.
+
+    This function allows you to browse MEG and EEG raw data stored in a BIDS
```
agramfort, line 11:
```suggestion
    This function allows you to browse MEG, EEG and iEEG raw data stored in a BIDS
```

agramfort (approved):
besides LGTM. Worked great on some data locally.

hoechenberger:
Merging so we can iterate!

Issue #571, opened 2020-10-02
Roundtrip error from EDF -> BV for write_raw_bids and then read_raw_bids? Possibly linked to mark_bad_channels?


adam2392:
#### Describe the bug
There is a potential bug that might even extend to mne-python, but I am having a hard time understanding what is going on. I basically have this EDF file that I want to convert to BIDS and in so doing, would prefer doing BV write since that seems less error prone compared to an EDF write.

However, when I load back the data, I get a bunch of nans in places. I suspect this is due to actually the `units` of the channels being messed up somehow. So I noticed that my `raw._orig_units` changes from uV to nV after reading/writing.

#### Steps to reproduce
```Python
   entities = {'subject': '001', 'session': '001', 'task': 'interictal', 'run': '01', 'suffix': 'ieeg', 'datatype': 'ieeg', 'extension': '.vhdr'}

    # load in EDF file
    raw = read_raw_edf(edf_fpath)

    nan_indx = np.argwhere(np.isnan(raw.get_data()))
    if nan_indx.any():
        print(nan_indx)
        print(len(nan_indx))

    # set channel types
    raw.set_channel_types({ch: 'ecog' for ch in raw.ch_names})

    # set line freq
    raw.info['line_freq'] = 60

    # construct the BIDS path that the file will get written to
    bids_path = BIDSPath(**entities, root=bids_root)

    # Get acceptable range of days back and pick random one
    daysback_min, daysback_max = get_anonymization_daysback(raw)
    daysback = random.randrange(daysback_min, daysback_max)
    anonymize = dict(daysback=0, keep_his=False)

    # write to BIDS based on path
    print(F'HERE ARE THE RAW UNITS: {raw._orig_units}')
    raise Exception('hi')
    output_bids_path = write_raw_bids(raw, bids_path=bids_path,
                                      anonymize=anonymize,
                                      overwrite=True, verbose=False)
    
    # mark one bad channel
    mark_bad_channels(raw.ch_names[0], bids_path=bids_path,
                          descriptions=['Excel-labeled'],
                          overwrite=False)

    raw = read_raw_bids(bids_path)
    raw.load_data()
      print(raw._orig_units)
      nan_indx = np.argwhere(np.isnan(raw.get_data()))
      if nan_indx.any():
                print(nan_indx)
                print(len(nan_indx))
                raise Exception('WTF?')
```

#### Expected results
I'm not sure what happens here but I think the `mark_bad_channels` basically re-normalizes the data a second time?

[...]

PR #574, opened 2020-10-03
[MRG] Temporary fix for mark_bad_channels

adam2392:
PR Description
--------------

Addresses `.fif` files for: #571 

and prevents this from producing nan files when user does not have `.fif` files.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1502,6 +1502,10 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
         _write_raw_brainvision(raw, bids_path.fpath, events)
     elif isinstance(raw, mne.io.RawFIF):
         raw.save(raw.filenames[0], overwrite=True, verbose=False)
+    else:
+        raise RuntimeError('Can only mark bad channels for '
+                           'FIF and BV files right now. Please '
+                           'mark bad channels manually.')
 
 
 def write_meg_calibration(calibration, bids_path, verbose=None):
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1498,7 +1498,7 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
     if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
         events, event_id = _read_events(events_data=None, event_id=None,
                                         raw=raw,
-                                        ext='.vhdr', verbose=verbose)
+                                        ext='.vhdr', verbose=False)
         _write_raw_brainvision(raw, bids_path.fpath, events)
     elif isinstance(raw, mne.io.RawFIF):
         raw.save(raw.filenames[0], overwrite=True, verbose=False)
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1504,7 +1504,7 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
         raw.save(raw.filenames[0], overwrite=True, verbose=False)
     else:
         raise RuntimeError('Can only mark bad channels for '
-                           'FIF and BV files right now. Please '
+                           'FIFF and BrainVision files for now. Please '
                            'mark bad channels manually.')
 
 
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -17,6 +17,7 @@ from collections import defaultdict, OrderedDict
 import numpy as np
 from scipy import linalg
 from numpy.testing import assert_array_equal
+import mne
 from mne.transforms import (_get_trans, apply_trans, get_ras_to_neuromag_trans,
                             rotation, translation)
 from mne import Epochs
@@ -1494,7 +1495,12 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
     bads = _get_bads_from_tsv_data(tsv_data)
     raw.info['bads'] = bads
     # XXX (How) will this handle split files?
-    if raw.filenames[0].endswith('.fif'):
+    if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
+        events, event_id = _read_events(events_data=None, event_id=None,
+                                        raw=raw,
+                                        ext='.vhdr', verbose=verbose)
+        _write_raw_brainvision(raw, bids_path.fpath, events)
+    elif isinstance(raw, mne.io.RawFIF):
         raw.save(raw.filenames[0], overwrite=True, verbose=False)
 
 
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1494,7 +1494,8 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
     bads = _get_bads_from_tsv_data(tsv_data)
     raw.info['bads'] = bads
     # XXX (How) will this handle split files?
-    raw.save(raw.filenames[0], overwrite=True, verbose=False)
+    if raw.filenames[0].endswith('.fif'):
+        raw.save(raw.filenames[0], overwrite=True, verbose=False)
 
 
 def write_meg_calibration(calibration, bids_path, verbose=None):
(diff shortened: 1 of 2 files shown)


hoechenberger (changes requested):
I'd rather you checked for the instance type of `raw` instead. For BV, this would be `RawBrainVision`

mne_bids/write.py
```diff
@@ -1493,7 +1493,8 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
     bads = _get_bads_from_tsv_data(tsv_data)
     raw.info['bads'] = bads
     # XXX (How) will this handle split files?
-    raw.save(raw.filenames[0], overwrite=True, verbose=False)
+    if raw.filenames[0].endswith('.fif'):
+        raw.save(raw.filenames[0], overwrite=True, verbose=False)
```
agramfort, line 1497: indeed it does not seem to be the right fix.

I would write to fif if we have MEG data / are dealing if meg modality.
what if you have CTF, KIT meg data?

hoechenberger: > I would write to fif if we have MEG data / are dealing if meg modality.

We always want to overwrite the input data, so using the same filename is fine. The tricky part is choosing the correct writer; e.g. BV data we'd need to manually pass to `pybv` for writing instead of just calling `Raw.save`

mne_bids/write.py
```diff
@@ -1493,7 +1494,12 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
...
     raw.info['bads'] = bads
     # XXX (How) will this handle split files?
-    if raw.filenames[0].endswith('.fif'):
+    if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
+        events, event_id = _read_events(events_data=None, event_id=None,
+                                        raw=raw,
+                                        ext='.vhdr', verbose=verbose)
+        _write_raw_brainvision(raw, bids_path.fpath, events)
+    elif isinstance(raw, mne.io.RawFIF):
         raw.save(raw.filenames[0], overwrite=True, verbose=False)
```
hoechenberger, line 1503: else, raise, or something :)

adam2392:
So now I'm just using the same logic that is present in the `write_raw_bids`. There's a bunch of variables so it's not easy to just refactor that logic out, but is this correct fix for checking type of `Raw`?

I suppose a unit test should be added too...

mne_bids/write.py
```diff
@@ -1493,7 +1494,12 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
     bads = _get_bads_from_tsv_data(tsv_data)
     raw.info['bads'] = bads
     # XXX (How) will this handle split files?
-    if raw.filenames[0].endswith('.fif'):
+    if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
+        events, event_id = _read_events(events_data=None, event_id=None,
+                                        raw=raw,
+                                        ext='.vhdr', verbose=verbose)
```
hoechenberger, line 1500: I think verbose should be False here

adam2392:
@hoechenberger do you mind explaining how the units work? It seems that the units of the channels.tsv are changed when `mark_bad_channels` is called? I am having a hard time dissecting that.

hoechenberger:
> It seems that the units of the channels.tsv are changed when `mark_bad_channels` is called?

Shouldn't happen, `mark_bad_channels` in `master` doesn't touch the units. I bet it has to do with you using `write_raw_bids()` here to write the raw, which would also overwrite the sidecars, whereas in `master` we're using `raw.save` directly to avoid any side effects like this

hoechenberger:
Maybe set a breakpoint at `write._channels_tsv()` to see when it gets called?

adam2392:
I updated the test to work, but it seems that because `write_brain_vision` automatically converts all data to `uV`, you get this:

```
raw (read in via read_raw_brainvision):
{'FP1': 'µV', 'FP2': 'µV', 'F3': 'µV', 'F4': 'µV', 'C3': 'µV', 'C4': 'µV', 'P3': 'µV', 'P4': 'µV', 'O1': 'µV', 'O2': 'µV', 'F7': 'µV', 'F8': 'µV', 'P7': 'µV', 'P8': 'µV', 'Fz': 'µV', 'FCz': 'µV', 'Cz': 'µV', 'CPz': 'µV', 'Pz': 'µV', 'POz': 'µV', 'FC1': 'µV', 'FC2': 'µV', 'CP1': 'µV', 'CP2': 'µV', 'FC5': 'µV', 'FC6': 'µV', 'CP5': 'n/a', 'CP6': 'µS', 'HL': 'n/a', 'HR': 'n/a', 'Vb': 'S', 'ReRef': 'C'}

raw_2 (with read_raw_bids):
{'FP1': 'µV', 'FP2': 'µV', 'F3': 'µV', 'F4': 'µV', 'C3': 'µV', 'C4': 'µV', 'P3': 'µV', 'P4': 'µV', 'O1': 'µV', 'O2': 'µV', 'F7': 'µV', 'F8': 'µV', 'P7': 'µV', 'P8': 'µV', 'Fz': 'µV', 'FCz': 'µV', 'Cz': 'µV', 'CPz': 'µV', 'Pz': 'µV', 'POz': 'µV', 'FC1': 'µV', 'FC2': 'µV', 'CP1': 'µV', 'CP2': 'µV', 'FC5': 'µV', 'FC6': 'µV', 'CP5': 'µV', 'CP6': 'µV', 'HL': 'µV', 'HR': 'µV', 'Vb': 'µV', 'ReRef': 'µV'}
```

So in order to make the two raws match up after using mark_bad_channels, one needs to drop these non-EEG channels from the dataset, which is normal behavior anyways. These non-EEG channels tho present w/ an edge case where the raw data will not match up due to different units and scaling of those specific channels. `'CP5', 'CP6', 'HL', 'HR', 'Vb', 'ReRef'`

hoechenberger:
> So in order to make the two raws match up after using mark_bad_channels, one needs to drop these non-EEG channels from the dataset, which is normal behavior anyways.

It's normal behavior for us to simply drop channels..?

adam2392:
> It's normal behavior for us to simply drop channels..?

How should I interpret/handle these channels in the original dataset?

`'CP5': 'n/a', 'CP6': 'µS', 'HL': 'n/a', 'HR': 'n/a', 'Vb': 'S', 'ReRef': 'C'`

vs when we perform `write_raw_brainvision`, these become:

`'CP5': 'µV', 'CP6': 'µV', 'HL': 'µV', 'HR': 'µV', 'Vb': 'µV', 'ReRef': 'µV'`

If I compare the two raws in the test without these channels, it's 1-1 match.

hoechenberger:
> How should I interpret/handle these channels in the original dataset?

Oh, you mean dropping **during the test**, right?

mne_bids/write.py
```diff
@@ -1494,7 +1495,17 @@ def mark_bad_channels(ch_names, descriptions=None, *, bids_path,
...
+    if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
+        events, event_id = _read_events(events_data=None, event_id=None,
+                                        raw=raw,
+                                        ext='.vhdr', verbose=False)
+        _write_raw_brainvision(raw, bids_path.fpath, events)
+    elif isinstance(raw, mne.io.RawFIF):
+        raw.save(raw.filenames[0], overwrite=True, verbose=False)
+    else:
+        raise RuntimeError('Can only mark bad channels for '
+                           'FIF and BV files right now. Please '
```
hoechenberger, line 1507:
```suggestion
                           'FIFF and BrainVision files for now. Please '
```

adam2392:
> Oh, you mean dropping **during the test**, right?

Yep! How should we get around that if that's an issue?

hoechenberger (approved):
> > Oh, you mean dropping **during the test**, right?
> 
> Yep! How should we get around that if that's an issue?

Never mind, I was totally confused and always thought you're talking about dropping channels during our ordinary `write_raw_bids()` calls! Sorry

hoechenberger:
Good to merge once CI goes green!

hoechenberger:
Thank you, @adam2392!

Issue #581, opened 2020-10-14
Should write_raw_bids create an events data structure even if no events/eventid passed in?


adam2392:
#### Describe the bug
If you have a dataset (say in EDF format), that you pass to write_raw_bids that has events in the Annotations of the Raw data structure, it is not written to `events.tsv` unless you explicitly pass in the events/eventid kwarg to `write_raw_bids`.

#### Expected results
I would expect it to tell me it is not doing it unless the events are encoded into a Stim channel, and if you want to parse those events explicitly, then you need to use `mne.events_from_annotations(raw)` to pass in the events explicitly.

#### Additional information
This line in `write_raw_bids` only looks for stim channel and only sets events if they are `.vhdr`, or `.set` extension. 

```
    events, event_id = _read_events(events_data, event_id, raw, ext,
                                    verbose=verbose)
```

My question is: Is this intended behavior? If so, I can make a quick PR updating the documentation on `write_raw_bids` to remind myself and other users to use `mne.events_from_annotations` explicitly if they don't have stim, vhdr, or set data.

agramfort:
a discussion was started about this in the past. My opinion is that we
should write whatever is passed. So if you have annotations write them and
if you have events passed explicitly you write them too.

>

hoechenberger:
Yes we should merge things. So far Annotations are simply discarded.

sappelhoff:
+1 to write events as extensively as we can without guessing. If users don't want that, they can do `raw.annotations = None` before mne-bids

PR #582, opened 2020-10-15
[MRG] Fixing write_raw_bids to write events.tsv if they are encoded in annotations

adam2392:
PR Description
--------------

Closes: #581 

Lmk if you have any ways of "compressing" the `_find_events` function, which has a lot of if/else. I'm not too familiar w/ `stim`, so left that alone.

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -46,7 +46,7 @@ Changelog
 - Allow :func:`mne_bids.write_raw_bids` to write EEG/iEEG files from Nihon Kohden using ``mne.io.read_raw_nihon`` function, by `Adam Li`_ (`#567 <https://github.com/mne-tools/mne-bids/pull/567>`_)
 - Allow :func:`mne_bids.get_entity_vals` to ignore datatypes using ``ignore_datatypes`` kwarg, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
 - Add ``with_key`` keyword argument to :func:`mne_bids.get_entity_vals` to allow returning the full entity strings, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
-- :func:`mne_bids.write_raw_bids` now also writes :attr:`mne.io.Raw.annotations` to ``*_events.tsv``, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
+- :func:`mne_bids.write_raw_bids` now also writes :attr:`mne.io.Raw.annotations` to ``*_events.tsv``, by `Adam Li`_ and `Richard Höchenberger`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -67,6 +67,8 @@ Bug
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
 - Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned, by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
 - Ensure :func:`mne_bids.print_dir_tree` prints files and directories in alphabetical order, by `Richard Höchenberger`_ (`#563 <https://github.com/mne-tools/mne-bids/pull/563>`_)
+- Ensure :func:`mne_bids.write_raw_bids` writes events encoded in annotations of the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
+
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -46,6 +46,7 @@ Changelog
 - Allow :func:`mne_bids.write_raw_bids` to write EEG/iEEG files from Nihon Kohden using ``mne.io.read_raw_nihon`` function, by `Adam Li`_ (`#567 <https://github.com/mne-tools/mne-bids/pull/567>`_)
 - Allow :func:`mne_bids.get_entity_vals` to ignore datatypes using ``ignore_datatypes`` kwarg, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
 - Add ``with_key`` keyword argument to :func:`mne_bids.get_entity_vals` to allow returning the full entity strings, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
+- :func:`mne_bids.write_raw_bids` now also writes  :class:`mne.Annotations` if present in the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
 
 Bug
 ~~~
@@ -67,8 +68,6 @@ Bug
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
 - Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned, by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
 - Ensure :func:`mne_bids.print_dir_tree` prints files and directories in alphabetical order, by `Richard Höchenberger`_ (`#563 <https://github.com/mne-tools/mne-bids/pull/563>`_)
-- :func:`mne_bids.write_raw_bids` now also writes  :class:`mne.Annotations` if present in the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
-
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -46,7 +46,7 @@ Changelog
 - Allow :func:`mne_bids.write_raw_bids` to write EEG/iEEG files from Nihon Kohden using ``mne.io.read_raw_nihon`` function, by `Adam Li`_ (`#567 <https://github.com/mne-tools/mne-bids/pull/567>`_)
 - Allow :func:`mne_bids.get_entity_vals` to ignore datatypes using ``ignore_datatypes`` kwarg, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
 - Add ``with_key`` keyword argument to :func:`mne_bids.get_entity_vals` to allow returning the full entity strings, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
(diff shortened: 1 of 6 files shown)


doc/whats_new.rst
```diff
@@ -67,6 +67,8 @@ Bug
 - Fix :func:`mne_bids.write_raw_bids` failed BIDS validator for ``raw.info['dig'] = []`` by `Alex Rockhill`_ (`#505 <https://github.com/mne-tools/mne-bids/pull/505>`_)
 - Fix bug in :func:`mne_bids.BIDSPath.fpath` where no data file was returned when the suffix was not assigned, by `Alex Rockhill`_ (`#542 <https://github.com/mne-tools/mne-bids/pull/542>`_)
 - Ensure :func:`mne_bids.print_dir_tree` prints files and directories in alphabetical order, by `Richard Höchenberger`_ (`#563 <https://github.com/mne-tools/mne-bids/pull/563>`_)
+- Ensure :func:`mne_bids.write_raw_bids` writes events encoded in annotations of the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
```
hoechenberger, line 70:
```suggestion
- :func:`mne_bids.write_raw_bids` now also writes  :class:`mne.Annotations` if present in the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
```

hoechenberger: This is an enhancement / new feature, not a bug fix, as we never claimed to preserve Annotations. Can you move it to the respective section?

mne_bids/utils.py
```diff
@@ -228,12 +228,18 @@ def _read_events(events_data, event_id, raw, ext, verbose=None):
                              f'found {events_data.shape[1]}')
         events = events_data
     elif 'stim' in raw:
+        print('USING STIM FOR EVENTS!!!! \n\n\n')
```
hoechenberger, line 231: Cruft?

hoechenberger (changes requested):
My view is that we should preserve **all** events here, i.e. if there are Annotations, write them; if there's `events`, write them. If we have both, write the union of both.

I also believe we should drop automated event extraction from the STIM channels, as this has proven to be buggy / not reliable for quite a number of MEG datasets in the past (at least with the default values)

adam2392:
> My view is that we should preserve **all** events here, i.e. if there are Annotations, write them; if there's `events`, write them. If we have both, write the union of both.

What's the best way/practice for combining events? I presume we want them i) ordered and ii) have unique event IDs per different event description.

Would the best way be to simply grab events from filename/`events_data` kwarg, `stim` ch events and annotation events and then:

1. combine events_id from `events_id` kwarg and annotation events_id since IDs need to operate like a dictionary. Probably accomplish this by finding the max event_id so far and incrementing by 1 if a new event is found?
2. np.concatenate the events array and update the event IDS from step 1
3. np.argsort by onset time of event

or am I over-complicating things?

adam2392:
There's this weird error coming up in `_read_raw_brainvision` after the changes I just pushed up... Not sure why this would be occurring tho.

hoechenberger:
> or am I over-complicating things?

I would try to turn everything we have into Annotations, and then use `mne.events_from_annotations`. This should take care of everything, I suppose?

adam2392:
> I would try to turn everything we have into Annotations, and then use `mne.events_from_annotations`. This should take care of everything, I suppose?
> 
> https://mne.tools/dev/auto_tutorials/intro/plot_20_events_from_raw.html#converting-between-events-arrays-and-annotations-objects

Ah okay yeah that would work. We can just append `Annotations` objects and then call `mne.events_from_annotations`.

A few questions then there because `Annotations` requires onset, duration and description:

1. For `stim` channel: `find_events` does not return events_id, which encodes the description. How should I handle that.
2. For events_data being a file, how should I construct the descriptions?
3. If `events_data` is passed in, but `event_id` is not, how should I construct the descriptions?

agramfort:
for events we have a way as you need to pass event_id. For annotations you
have everything. So keep it simple.

>

hoechenberger:
> 1. For `stim` channel: `find_events` does not return events_id, which encodes the description. How should I handle that.

Honestly I would drop support for this auto-event-extraction sorcery and force users to do this manually first, and then pass `events` and `event_id`.

> 2. For events_data being a file, how should I construct the descriptions?

Force users to pass an `event_id` dict.

> 3. If `events_data` is passed in, but `event_id` is not, how should I construct the descriptions?

I would raise an exception; see 2.

mne_bids/utils.py
```diff
@@ -227,17 +226,25 @@ def _read_events(events_data, event_id, raw, ext, verbose=None):
...
-        events, event_id = events_from_annotations(raw, event_id,
-                                                   verbose=verbose)
     else:
+        events = np.empty(shape=(0, 3))
+
+    # get events from annotations
+    events_annot, event_id_annot = events_from_annotations(raw, event_id,
+                                                           verbose=verbose)
+    if event_id is not None:
+        events = np.concatenate((events, events_annot), axis=0)
```
hoechenberger, line 236: Probably unnecessary, but could you sort based on the first column too?

```suggestion
        events = np.concatenate((events, events_annot), axis=0)
        events.sort(axis=0)
```

mne_bids/utils.py
```diff
@@ -227,17 +226,25 @@ def _read_events(events_data, event_id, raw, ext, verbose=None):
...
+        event_id.update(event_id_annot)
+    else:
+        events = events_annot
+        event_id = event_id_annot
```
hoechenberger, line 240: Can we turn the `if/else` block around?

```
if event_id is None:
    events = events_annot
    ...
else:
    ...
```
I always find "if not None ... else" difficult to parse, mentally ;)

mne_bids/utils.py
```diff
@@ -14,12 +14,12 @@
 from os import path as op
 
 import numpy as np
-from mne import read_events, find_events, events_from_annotations
+from mne import (read_events, events_from_annotations)
```
hoechenberger, line 17:
```suggestion
from mne import read_events, events_from_annotations
```

mne_bids/utils.py
```diff
@@ -217,6 +215,7 @@ def _read_events(events_data, event_id, raw, ext, verbose=None):
         before the event or immediately after.
 
     """
+    # get events from the events_data passed in
```
hoechenberger, line 218:
```suggestion
    # get events from events_data
```

mne_bids/write.py
```diff
@@ -1072,7 +1087,7 @@ def write_raw_bids(raw, bids_path, events_data=None,
         logger.warning(f'Writing of electrodes.tsv is not supported '
                        f'for data type "{bids_path.datatype}". Skipping ...')
 
-    events, event_id = _read_events(events_data, event_id, raw, ext,
+    events, event_id = _read_events(events_data, event_id, raw,
                                     verbose=verbose)
```
hoechenberger, line 1091:
```suggestion
                                    verbose=False)
```

doc/whats_new.rst
```diff
@@ -46,6 +46,7 @@ Changelog
 - Allow :func:`mne_bids.write_raw_bids` to write EEG/iEEG files from Nihon Kohden using ``mne.io.read_raw_nihon`` function, by `Adam Li`_ (`#567 <https://github.com/mne-tools/mne-bids/pull/567>`_)
 - Allow :func:`mne_bids.get_entity_vals` to ignore datatypes using ``ignore_datatypes`` kwarg, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
 - Add ``with_key`` keyword argument to :func:`mne_bids.get_entity_vals` to allow returning the full entity strings, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
+- :func:`mne_bids.write_raw_bids` now also writes  :class:`mne.Annotations` if present in the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
```
agramfort, line 49:
```suggestion
- :func:`mne_bids.write_raw_bids` now also writes :class:`mne.Annotations` if present in the raw file, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
```

mne_bids/tests/test_write.py
```diff
@@ -883,13 +883,37 @@ def test_eegieeg(dir_name, fname, reader, _bids_validate):
...
+
+    # read in the source raw file
     raw = reader(raw_fname)
 
+    # if there were events in the original raw dataset that
+    # were not encoded in the stim channel,
+    # they should be written even if write_raw_bids does not
+    # pass in events
+    events, events_id = mne.events_from_annotations(raw, event_id=None)
+    if events_id != {} and 'stim' not in raw:
```
agramfort, line 897: I don't get why this `if events_id != {} and 'stim' not in raw:`

to me it should be simpler. If you have annotations present with raw it's written and you read
the written events as an annotation to make sure it's correct.

bids events.tsv is very close to an mne annotations written to disk as .txt or csv

agramfort:
my message from this morning that did not go through:

```when you read you have annotations added. I would expect that writing back would just
take the available annotations and just work```

File: mne_bids/tests/test_write.py (hunk 9 above)
Comment (adam2392) [reply mne_bids/tests/test_write.py]: Ah okay I got rid of this `if` statement. I don't think it's necessary anymore and I took it out and the tests worked.

**Review by hoechenberger — APPROVED** [review]

File: mne_bids/utils.py [hunk 10]
```diff
@@ -227,17 +226,28 @@ def _read_events(events_data, event_id, raw, ext, verbose=None):
...
                              f'found {events_data.shape[1]}')
         events = events_data
-    elif 'stim' in raw:
-        events = find_events(raw, min_duration=0.001, initial_event=True,
-                             verbose=verbose)
-    elif ext in ['.vhdr', '.set'] and check_version('mne', '0.18'):
-        events, event_id = events_from_annotations(raw, event_id,
-                                                   verbose=verbose)
     else:
+        events = np.empty(shape=(0, 3))
```
agramfort, line 230:
```suggestion
        events = np.empty(shape=(0, 3), dtype=int)
```

doc/whats_new.rst
```diff
@@ -46,6 +46,7 @@ Changelog
 - Allow :func:`mne_bids.write_raw_bids` to write EEG/iEEG files from Nihon Kohden using ``mne.io.read_raw_nihon`` function, by `Adam Li`_ (`#567 <https://github.com/mne-tools/mne-bids/pull/567>`_)
 - Allow :func:`mne_bids.get_entity_vals` to ignore datatypes using ``ignore_datatypes`` kwarg, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
 - Add ``with_key`` keyword argument to :func:`mne_bids.get_entity_vals` to allow returning the full entity strings, by `Adam Li`_ (`#578 <https://github.com/mne-tools/mne-bids/pull/578>`_)
+- :func:`mne_bids.write_raw_bids` now also writes :attr:`mne.io.Raw.annotations` to ``*_events.tsv``, by `Adam Li`_ (`#582 <https://github.com/mne-tools/mne-bids/pull/582>`_)
```
agramfort, line 49: @hoechenberger add your name here. You did a lot.

agramfort (approved):
besides nitpicks LGTM !

thx @adam2392 and @hoechenberger

hoechenberger:
Thanks @adam2392!

adam2392:
Thanks @hoechenberger for finishing this out :)

PR #607, opened 2020-10-28
[MRG] Edit 'no events found' warning for resting state data

ethanknights:
PR Description
--------------

When no events information is provided, a check is added if the raw filename string includes 'rest'. If so, this is acknowledged in a warning (rather than the warning about providing annotations).
Ideally, the string comparison would be for 'task-rest_', but the raw.filename is pre-BIDS. This means  some unintended cases might be picked up if 'rest' is part of a different task/filename, but this should be rare (and is why I didn't remove the warning entirely).

Note I tried running pytest, following the contributions.MD but I get 9 errors which I don't _think_ would have been caused by this very minor change (e.g. in pick.py). I can look into this if needed.

Feel free to reject/close if this seems unnecessary (I just found helpful when processing datasets with a combination of tasks with/out events).  

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


hoechenberger:
Hello @ethanknights and welcome! Thanks for your contribution. 👍👍👍 
We'll try to review it and get back to you soon.

mne_bids/read.py
```diff
@@ -148,8 +149,14 @@ def _read_events(events_data, event_id, raw, verbose=None):
     )
     all_dur = raw.annotations.duration
     if all_events.size == 0:
-        warn('No events found or provided. Please add annotations to the raw '
-             'data, or provide the events_data and event_id parameters.')
+        #Confirm from filename dataset is not likely resting state
+        fileparts = op.split(raw.filenames[0])
+        if not re.search('rest',fileparts[-1]):
```
agramfort, line 154:
```suggestion
        if 'rest' not in fileparts[-1]:
```

mne_bids/read.py
```diff
@@ -148,8 +149,14 @@ def _read_events(events_data, event_id, raw, verbose=None):
...
+            warn('No events found or provided. Please add annotations to the raw '
+                    'data, or provide the events_data and event_id parameters.')
```
agramfort, line 156:
```suggestion
            warn('No events found or provided. Please add annotations to the raw '
                 'data, or provide the events_data and event_id parameters.')
```

mne_bids/read.py
```diff
@@ -148,8 +149,14 @@ def _read_events(events_data, event_id, raw, verbose=None):
...
+        else:
+            warn('No events found or provided. Filename suggests data is resting state, '
+                    'so please ignore if deliberate.')
```
agramfort, line 159:
```suggestion
            warn('No events found or provided. Filename suggests data is resting state, '
                 'so please ignore if deliberate.')
```

mne_bids/read.py
agramfort: but @hoechenberger I would actually not warn at all. I would just log things when verbose=True

agramfort (commented):
@ethanknights thanks !

you should avoir opening PRs from your master branch but rather create a dedicate branch. Let us know if you need guidance.

mne_bids/read.py
sappelhoff: I think the problem may be that when no events are found, no events.tsv is written, and then the data may be invalid BIDS ... users may not realize their error if they have verbose = False. 

So I think this should be logged or warned about even when verbose=False.

However for the case of resting data (this PR), I am +1 to not log/warn

mne_bids/read.py
```diff
@@ -9,6 +9,7 @@
 import os.path as op
 import glob
 import json
+import re
```
sappelhoff, line 12: as Alex shows in his comment: https://github.com/mne-tools/mne-bids/pull/607/files#r513752558

the "re" module is not necessary for what you want to achieve

```suggestion
```

agramfort:
but events.tsv is not necessary right? also the word rest is not standard.
i have data now where rest condition is called "spont" ....

I would just remove the warning

>

sappelhoff:
> but events.tsv is not necessary right? 

True

"rest" is only RECOMMENDED, so yes - it can also be called differently (unfortunately):

> A RECOMMENDED convention is to name resting state task using labels beginning with rest

---

okay, fine with me to turn the warning into a logging info to be printed if  `verbose==True` ... and to not print at all if task contains "rest"

agramfort:
ok for me !

then I need to fix my other dataset :)

hoechenberger:
Sounds good. @ethanknights Want to give this a try?

- Do not warn
- Log info, but only if `task` is not `rest` and `verbose is True`

ethanknights:
> @ethanknights thanks !
> 
> you should avoir opening PRs from your master branch but rather create a dedicate branch. Let us know if you need guidance.

Noted thanks 

> Sounds good. @ethanknights Want to give this a try?

> Do not warn
> Log info, but only if task is not rest and verbose is True

Ok, yes happy with people's points and will sort this in a day or two (need to look at how logging works here) - cheers!

ethanknights:
Reopened here #631

PR #626, opened 2020-11-13
[MRG] BUMP pybv 0.3 -> 0.4

### LINKED ISSUE #621: Next release – Bump pybv requirement to 0.4
We've fixed some serious issues here, which demand a new release; and I think once pybv 0.4 has been released, we should bump our requirement to that version, because otherwise we might end up writing … problematic BV files. WDYT? 

cc @sappelhoff

sappelhoff:
closes #621 

Merge checklist
---------------

- [ ] All comments resolved
- [x] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -271,7 +271,7 @@ def test_create_fif(_bids_validate):
     _bids_validate(bids_root)
 
 
-@requires_version('pybv', '0.3')
+@requires_version('pybv', '0.4')
 @pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
 def test_fif(_bids_validate):
     """Test functionality of the write_raw_bids conversion for fif."""
@@ -866,7 +866,7 @@ def test_vhdr(_bids_validate):
     assert len([f for f in os.listdir(data_path) if op.isfile(f)]) == 0
 
     # test anonymize and convert
-    if check_version('pybv', '0.3'):
+    if check_version('pybv', '0.4'):
         raw = _read_raw_brainvision(raw_fname)
         output_path = _test_anonymize(raw, bids_path)
         _bids_validate(output_path)
@@ -1084,7 +1084,7 @@ def test_eegieeg(dir_name, fname, reader, _bids_validate):
     assert len(list(data.values())[0]) == 2
 
     # check that scans list is properly converted to brainvision
-    if check_version('pybv', '0.3'):
+    if check_version('pybv', '0.4'):
         daysback_min, daysback_max = _get_anonymization_daysback(raw)
         daysback = (daysback_min + daysback_max) // 2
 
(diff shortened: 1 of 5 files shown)


hoechenberger (approved):

mne_bids/write.py
```diff
@@ -681,8 +681,8 @@ def _write_raw_brainvision(raw, bids_fname, events):
         The events as MNE-Python format ndaray.
 
     """
-    if not check_version('pybv', '0.3'):
-        raise ImportError('pybv >=0.3 is required for converting '
+    if not check_version('pybv', '0.4'):
```
sappelhoff, line 684:
```suggestion
    if not check_version('pybv', '0.4'):  # pragma: no cover
```

mne_bids/write.py
```diff
@@ -709,17 +709,20 @@ def _write_raw_brainvision(raw, bids_fname, events):
                'mne-bids. Please contact the mne-bids team about this.')
         raise RuntimeError(msg)
 
-    # XXX Always write as float32 to avoid precision loss during I/O roundtrip.
-    # See https://github.com/bids-standard/pybv/issues/45
-    fmt = 'binary_float32'
+    # We enforce conversion to float32 format
+    # XXX: pybv can also write to int16, to do that, we need to get
+    # original units of data prior to conversion, and add a optimization
```
sappelhoff, line 714:
```suggestion
    # original units of data prior to conversion, and add an optimization
```

mne_bids/tests/test_write.py
```diff
@@ -271,7 +271,7 @@ def test_create_fif(_bids_validate):
     _bids_validate(bids_root)
 
 
-@requires_version('pybv', '0.3')
+@requires_version('pybv', '0.4')
```
hoechenberger, line 274: Why don't we simply remove these? Developers should have pybv installed to run the tests, and we don't want any tests to be silently skipped, no?

sappelhoff: mmh yes, I don't actually know why we have these in the test scripts.

agramfort: no we don't want to impose to users all dependencies. If they don't need pybv the tests should not break for them.

also for linux packaging eg on debian if we do this it means we need a debian packages for pybv otherwise mne will not pass the tests. Not sure you have the bandwidth to package pybv on debian :)

hoechenberger: But to run the tests, you also need bids-validator, and this requires node, … And which ordinary user grabs and runs the tests, anyway? The recommended installation is via pip or conda, and the sdist doesn't contain any tests.

sappelhoff: ahh I remember why we had this! If we run a test with "minimal dependencies", the tests involving optional dependencies need to be skipped.

I think this PR can be merged as is

agramfort:
thx @sappelhoff

PR #629, opened 2020-11-15
ENH stats event counts

agramfort:
new count_events function


mne_bids/stats.py
```diff
@@ -0,0 +1,78 @@
...
+# Authors: Alex Gramfort <alexandre.gramfort@inria.fr>
+#
+# License: BSD (3-clause)
+
+
+from mne_bids import BIDSPath
+
+
+def count_events(root, subject=None, session=None, task=None, run=None):
+    """Count events
```
hoechenberger, line 12: I feel this should accept a `BIDSPath` – or even be exposed as a `BIDSPath` method?

agramfort:
see https://3496-89170358-gh.circle-artifacts.com/0/dev/generated/mne_bids.stats.count_events.html#mne_bids.stats.count_events

agramfort:
Screenshot 2020-11-15 at 17 42 40

in https://3496-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_mne_sample.html#sphx-glr-auto-examples-convert-mne-sample-py

jasmainak:
good stuff. Do we plan to have other stat functions in the future? One thing that immediately comes to mind is that we could add a summary of events also in the report. Something like: "There are 63 +/- 3 trials for condition A and 72 +/- 5 trials for condition B"

agramfort:
> good stuff. Do we plan to have other stat functions in the future? One thing that immediately comes to mind is that we could add a summary of events also in the report. Something like: "There are 63 +/- 3 trials for condition A and 72 +/- 5 trials for condition B"

it's what the --describe option does in the CLI and otherwise use `count_events(...).describe()`

agramfort:
ok I think I am done here @hoechenberger @sappelhoff @jasmainak

mne_bids/stats.py
agramfort: done

.github/workflows/unit_tests.yml
```diff
@@ -75,6 +75,10 @@ jobs:
         mne sys_info
     - name: Install MNE-BIDS
       run: pip install --no-deps .
+    - name: Install pandas
+      if: "matrix.os == 'ubuntu-latest'"
+      run: |
+        pip install -U pandas
```
jasmainak, line 81: I would also list `pandas` in the installation page as an optional dependency

mne_bids/stats.py
```diff
@@ -0,0 +1,110 @@
...
+# Authors: Alex Gramfort <alexandre.gramfort@inria.fr>
+#
+# License: BSD (3-clause)
+
+
+from mne_bids import BIDSPath, get_datatypes
+from mne_bids.config import EPHY_ALLOWED_DATATYPES
+
+
+def count_events(root_or_path, datatype='auto'):
```
jasmainak, line 12: this feels confusing. Why not just `bids_path` instead of `root_or_path`? any use cases not covered by it?

mne_bids/stats.py
```diff
@@ -0,0 +1,110 @@
...
+        bids_path = BIDSPath(root=root_or_path)
+    else:
+        bids_path = root_or_path.copy()
+
+    bids_path.update(suffix='events', extension='tsv')
+
+    datatypes = get_datatypes(bids_path.root)
+    this_datatypes = list(set(datatypes).intersection(EPHY_ALLOWED_DATATYPES))
+
+    if (datatype == 'auto') and bids_path.datatype is not None:
```
jasmainak, line 47:
```suggestion
    if datatype == 'auto' and bids_path.datatype is not None:
```

jasmainak (approved):
Took a quick look, didn't check the tests. But LGTM :) I trust @agramfort 's code ;-)

mne_bids/stats.py
agramfort: I considered it and i find it easier to do

```
from mne_bids.stats import count_events
count_events('/path/...)
```

I consider it also the default usage. So let's not make the default more verbose
of genericity.

.github/workflows/unit_tests.yml
agramfort: done

mne_bids/stats.py
agramfort: I rather added some () :)

hoechenberger: Yes I like the parens ;)

hoechenberger: By why not put parens around the second term too :)

jasmainak: I won't question French aesthetics :)

hoechenberger (approved):
Just a few suggestions and a question, otherwise looking good!

mne_bids/stats.py
```diff
@@ -0,0 +1,110 @@
...
+
+    tasks = sorted(set([bp.task for bp in bids_path.match()]))
+
+    all_counts = []
+
+    for task in tasks:
+        bids_path.update(task=task)
+
+        all_df = []
+        for bp in bids_path.match():
```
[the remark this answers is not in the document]
agramfort: no it cannot.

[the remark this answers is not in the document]
agramfort: well technically it can but if it does then you get a value error below.

hoechenberger (approved):

hoechenberger:
Thanks @agramfort ! Awesome

mne_bids/stats.py
```diff
@@ -0,0 +1,111 @@
...
+# License: BSD (3-clause)
+
+
+from mne_bids import BIDSPath, get_datatypes
+from mne_bids.config import EPHY_ALLOWED_DATATYPES
+
+
+def count_events(root_or_path, datatype='auto'):
+    """Count events present in dataset.
+
```
adam2392, line 14:
```suggestion

Summarizes events per task denoted by the ``task`` BIDS entity.
```

What if there are no "tasks" enumerated, but there is still an events.tsv file?
e.g.

```
sub-001_ses-01_run-01_ieeg.vhdr
sub-001_ses-01_run-01_events.tsv
...
```?

sappelhoff: would that be valid BIDS? no task entity?

adam2392: Looks like no: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html so nvm.

hmm wondering if we should add checks like this eventually into BIDSPath.

If instantiating a BIDSPath that represents a modality-specific file, need certain entities, etc.

hoechenberger: The only thing I would add, if we don’t have it yet, is a check to ensure we have a task entity when writing

agramfort:
try this code on your data and let's improve it together if it does not do
the job fully

PR #631, opened 2020-11-16
[MRG] Edit 'no events found' warning for resting state data - Reopened

ethanknights:
PR Description
--------------

This PR reopens #607 (I took Alex's advice and started a new branch - too late now, but I realise I didn't need to refork for this...)

For converting resting state data to BIDS, where there are no events, the warning is changed to a log (if set to verbose). 

Note I have not yet managed to see the log printed, but I followed the same syntax in `read.py ` in `_handle_events_reading`: 
```
def _handle_events_reading(events_fname, raw):
    """Read associated events.tsv and populate raw.

    Handle onset, duration, and description of each event.
    """
    logger.info('Reading events from {}.'.format(events_fname))
```

I suspect I just have not set the logger up correctly - when trying to print this log, I was testing this [script](https://github.com/ethanknights/Cam-CAN-to-BIDS/blob/CBU_Paths/Code/run_convert_camcan_rest.py) using variations of: `verbose='INFO'`  & `verbose=True` for these commands:
- `mne.set_log_level(verbose='INFO') `
- ```write_raw_bids(raw, bids_path=bids_path,
                   events_data=events,
                   event_id=event_name_to_id_mapping,
                   anonymize=anonDict,
                   overwrite=True,
                   verbose=True)```

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [x] All CIs are happy


mne_bids/read.py
```diff
@@ -147,9 +147,13 @@ def _read_events(events_data, event_id, raw, verbose=None):
...
-    if all_events.size == 0:
-        warn('No events found or provided. Please add annotations to the raw '
-             'data, or provide the events_data and event_id parameters.')
+    if all_events.size == 0 and verbose:
+        # Confirm from filename dataset is not likely resting state
+        if '_task-rest' not in raw.filenames[0]:
+            logger.info('No events found or provided. Please add annotations '
+                        'to the raw data, or provide the events_data and '
+                        'event_id parameters. If this is resting state data '
+                        'it is recommended to name the task "rest".')
```
agramfort, line 156: @ethanknights I pushed a tiny update here.

ok for you?

hoechenberger (approved):

hoechenberger:
CI fails because Node setup doesn't succeed, we probably need to upgrade `setup-node`

```
Run actions/setup-node@v1.4.2
2
  with:
3
    node-version: 12
4
    always-auth: false
5
  env:
6
    TZ: Europe/Berlin
7
    FORCE_COLOR: true
8
Error: Unable to process command '::add-path::/opt/hostedtoolcache/node/12.19.0/x64/bin' successfully.
9
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
10
```

agramfort:
yes all PRs are failing now :( :(

>

hoechenberger:
Working now. You can merge once green!

hoechenberger:
Tests are now failing because we only log info instead of emitting a warning, as we used to
@agramfort Do we really just want to "info" and not "warn" here?

hoechenberger:
(this is just a reminder for @agramfort that not too long ago, you were in favor of **removing** the warning: https://github.com/mne-tools/mne-bids/pull/607#issuecomment-720990228)

hoechenberger:
Some tests need to be fixed as we're not warning in all the places we used to.
Also we need a changelog entry.

Then it should be good to go!

mne_bids/read.py
```diff
@@ -147,9 +147,13 @@ def _read_events(events_data, event_id, raw, verbose=None):
         verbose=verbose
     )
     all_dur = raw.annotations.duration
-    if all_events.size == 0:
-        warn('No events found or provided. Please add annotations to the raw '
-             'data, or provide the events_data and event_id parameters.')
+    if all_events.size == 0 and verbose:
+        # Confirm from filename dataset is not likely resting state
+        if '_task-rest' not in raw.filenames[0]:
```
hoechenberger, line 152:
```suggestion
    if all_events.size == 0 and verbose:
        # Confirm from filename dataset is not likely resting state
        if '_task-rest' not in raw.filenames[0]:
```
Actually **this** needs to be fixed, not the tests per se.

```suggestion
if all_events.size == 0 and `_task_rest` not in raw.filenames[0]:
```

mne_bids/read.py
```diff
@@ -147,9 +147,7 @@ def _read_events(events_data, event_id, raw, verbose=None):
         verbose=verbose
     )
     all_dur = raw.annotations.duration
-    if all_events.size == 0 and verbose:
-        # Confirm from filename dataset is not likely resting state
-        if '_task-rest' not in raw.filenames[0]:
+if all_events.size == 0 and `_task_rest` not in raw.filenames[0]:
             warn('No events found or provided. Please add annotations '
```
hoechenberger: Indentation is off now, sorry ;)

ethanknights: :) 
Thanks - my bad I didn't check the fix above thoroughly enough
Just re-added `verbose == True`
(This is providing that we do indeed want the warning again (not logger.info))

mne_bids/read.py
```diff
@@ -147,11 +147,12 @@ def _read_events(events_data, event_id, raw, verbose=None):
         verbose=verbose
     )
     all_dur = raw.annotations.duration
-if all_events.size == 0 and `_task_rest` not in raw.filenames[0]:
-            warn('No events found or provided. Please add annotations '
-                 'to the raw data, or provide the events_data and '
-                 'event_id parameters. If this is resting state data '
-                 'it is recommended to name the task "rest".')
+    if all_events.size == 0 and '_task_rest' not in raw.filenames[0] and \
+        verbose == True:
```
hoechenberger, line 151: We don't need the verbose switch or do we?
In my view we should always callt he logger and the logger **itself** must decide, based on the verbosity level, whether to print the message or not. It that not how logging in MNE works?

hoechenberger: Also this check wouldn't print a warning if verbose=='info' or verbose=='debug'...

hoechenberger: I'll leave this to @agramfort to assess :)

agramfort:
yes as it's a warn it has nothing with verbose. You can remove the verbose.
Good point

>

ethanknights:
the verbose check is removed again. 



> yes as it's a warn it has nothing with verbose. You can remove the verbose. Good point
> […](#)

Ah I see, verbose is removed now.

hoechenberger:
Hmmm.

So I was just thinking about this. The function we're modifying here, `_read_events()`, is only used by `write_raw_bids()` and `mark_bad_channels()`.

While `mark_bad_channels()` only works with BIDS data, `write_raw_bids()` naturally operates on input data that can be non-BIDS too.

But this means that checking for `_task-rest` in the filename is not necessarily very helpful.

Instead, we shouldmaybe simply check for `rest`?

agramfort:
no objection

>

ethanknights:
> Hmmm.
> 
> So I was just thinking about this. The function we're modifying here, `_read_events()`, is only used by `write_raw_bids()` and `mark_bad_channels()`.
> 
> While `mark_bad_channels()` only works with BIDS data, `write_raw_bids()` naturally operates on input data that can be non-BIDS too.
> 
> But this means that checking for `_task-rest` in the filename is not necessarily very helpful.
> 
> Instead, we shouldmaybe simply check for `rest`?

Ok switched to 'rest' now @hoechenberger 
(Sorry for multiple commits - was struggling to merge my fork with upstream, but figured it out!)

agramfort:
thx @ethanknights !

PR #644, opened 2020-11-27
[MRG] Add Annotations support to MNE-BIDS Inspector; marking bad channels now only alters sidecar, not raw data

hoechenberger:
PR Description
--------------

This adds support for interactively editing Annotations (BIDS events) via `mne_bids.inspect_dataset()`. Will update the raw file and `events.tsv`.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/inspect.py
+++ b/mne_bids/inspect.py
@@ -4,7 +4,7 @@ import mne
 from mne.utils import logger
 
 from mne_bids import read_raw_bids, mark_bad_channels
-from mne_bids.read import _from_tsv, _read_events, _find_matching_sidecar
+from mne_bids.read import _from_tsv, _read_events
 from mne_bids.write import _write_raw_brainvision, _events_tsv
 from mne_bids.config import ALLOWED_DATATYPE_EXTENSIONS
 
--- a/mne_bids/inspect.py
+++ b/mne_bids/inspect.py
@@ -280,8 +280,8 @@ def _save_bads(*, bads, bids_path, verbose):
 
 
 def _save_annotations(*, annotations, bids_path, verbose):
-    # Read the raw data, set our new Annotations, and convert them to events
-    # which we can then stored in the *_events.tsv sidecar.
+    # Attach the new Annotations to our raw data so we can easily convert them
+    # to events, which will be stored in the *_events.tsv sidecar.
     extra_params = dict()
     if bids_path.extension == '.fif':
         extra_params['allow_maxshield'] = True
@@ -289,7 +289,6 @@ def _save_annotations(*, annotations, bids_path, verbose):
     raw = read_raw_bids(bids_path=bids_path, extra_params=extra_params,
                         verbose='warning')
     raw.set_annotations(annotations)
-
     events, durs, descrs = _read_events(events_data=None, event_id=None,
                                         raw=raw, verbose=False)
 
--- a/mne_bids/inspect.py
+++ b/mne_bids/inspect.py
@@ -1,11 +1,16 @@
 from pathlib import Path
 
+import mne
+from mne.utils import logger
+
 from mne_bids import read_raw_bids, mark_bad_channels
-from mne_bids.read import _from_tsv
+from mne_bids.read import _from_tsv, _read_events, _find_matching_sidecar
+from mne_bids.write import _write_raw_brainvision, _events_tsv
 from mne_bids.config import ALLOWED_DATATYPE_EXTENSIONS
 
 
-def inspect_dataset(bids_path, l_freq=None, h_freq=None, verbose=None):
+def inspect_dataset(bids_path, l_freq=None, h_freq=None,
+                    show_annotations=True, verbose=None):
     """Inspect and annotate BIDS raw data.
 
     This function allows you to browse MEG, EEG, and iEEG raw data stored in a
@@ -16,10 +21,12 @@ def inspect_dataset(bids_path, l_freq=None, h_freq=None, verbose=None):
 
     .. warning:: This functionality is still experimental and will be extended
                  in the future. Its API will likely change. Planned features
-                 include modification of annotations and automated bad channel
-                 detection.
+                 include automated bad channel detection and visualization of
+                 MRI images.
 
     .. note:: Currently, only MEG, EEG, and iEEG data can be inspected.
+
+    To add or modify annotations, press ``A`` to toggle annotation mode.
 
     Parameters
     ----------
@@ -40,6 +47,10 @@ def inspect_dataset(bids_path, l_freq=None, h_freq=None, verbose=None):
         The low-pass filter cutoff frequency to apply when displaying the
         data. This can be useful when inspecting data with high-frequency
         artifacts. If ``None``, no low-pass filter will be applied.
+    show_annotations : bool
+        Whether to show annotations (events, bad segments, …) or not. If
+        ``False``, toggling annotations mode by pressing ``A`` will be disabled
+        as well.
     verbose : bool | None
         If a boolean, whether or not to produce verbose output. If ``None``,
         use the default log level.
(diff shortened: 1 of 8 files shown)


hoechenberger:
Updated example:
https://3676-89170358-gh.circle-artifacts.com/0/dev/auto_examples/mark_bad_channels.html#interactive-use

hoechenberger:
How can I fix the Circle issue?

hoechenberger:
So I was running into some weird and (at first) unreproducible issues with the tests locally: tests behaved differently when run individually or "all at once", indicating some kind of side effects.

Turns out the `return_bids_test_dir` we're using ensures that the BIDS root our tests use will be setup once per test module, and then will be reused by all the tests, without any intermediate cleanup, causing us to carry artifacts from one test to the next. We're testing side effects :)

So for `test_inspect.py`, I've changed things now such that the temporary directory (BIDS root) will be recreated for each test function. This slows down testing a bit (but not **that** much really) but is IMHO absolutely necessary to ensure we're actually testing what we **intend** to test.

I propose to make the same adjustment for all other tests, too. (in a separate PR)

mne_bids/inspect.py
```diff
@@ -81,12 +93,24 @@ def _inspect_raw(*, bids_path, l_freq, h_freq, verbose=None):
...
-                              bids_path=bids_path,
-                              verbose=verbose)
+        new_annotations = mne_raw_fig.mne.inst.annotations.copy()
+
+        if not new_annotations:
+            # Ensure it's not an empty list, but an empty set of Annotations.
+            new_annotations = (mne_raw_fig.mne.inst.copy()
+                               .set_annotations(None)
+                               .annotations
+                               .copy())
```
agramfort, line 106: I don't get why you need this? making a copy of potentially a 4GB raw object to only look at annotations ?

mne_bids/inspect.py
```diff
@@ -97,23 +121,47 @@ def _keypress_callback(event):
...
     fig.canvas.mpl_connect('key_press_event', _keypress_callback)
 
+    if not show_annotations:
+        # Remove annotations and kill `_toggle_annotation_fig` method, since
+        # we cannot directly and easily remove the associated `a` keyboard
+        # event callback.
+        fig._clear_annotations()
+        fig._toggle_annotation_fig = lambda: None
+        # Ensure it's not an empty list, but an empty set of Annotations.
+        old_annotations = (raw.copy()
```
agramfort, line 131: same remark here

mne_bids/inspect.py
```diff
@@ -122,9 +170,18 @@ def _save_bads_dialog_box(*, bads, bids_path, verbose):
...
-    message = (f'You have modified the bad channel selection of\n'
-               f'{bids_path.basename}.\n\nWould you like to save these '
-               f'changes to the\nBIDS dataset?')
+    message = 'You have modified '
+    if bads is not None and annotations is None:
+        message += 'the bad channel selection '
+    elif bads is None and annotations is not None:
+        message += 'the bad segments selection '
+    else:
+        message += 'the bad channel and\nsegments selection '
```
agramfort, line 179: when you modify both the text is overlapping in the dialog box

mne_bids/inspect.py
```diff
@@ -205,11 +267,53 @@ def _save_bads(*, bads, bids_path, verbose):
...
+    events, durs, descrs = _read_events(events_data=None, event_id=None,
+                                        raw=raw, verbose=False)
+
+    # Write raw data.
+    if isinstance(raw, mne.io.brainvision.brainvision.RawBrainVision):
+        # XXX We should write durations too, this is supported by pybv.
+        _write_raw_brainvision(raw, bids_path.fpath, events)
+    elif isinstance(raw, mne.io.RawFIF):
+        raw.save(raw.filenames[0], overwrite=True, split_naming='bids',
+                 verbose=False)
```
agramfort, line 300: to me we should update the side car files and not alter the raw fif files

agramfort: cf discussion with Robert

agramfort (commented):
besides works great !

mne_bids/inspect.py
hoechenberger: Interesting, not for me. Could you share a screenshot? Which backend are you using?

mne_bids/inspect.py
hoechenberger: Yes this would make easier.
Just doing it this way here because that's also how we're currently handling bad channel marking as well – was trying to be consistent. But I will gladly remove this…
@sappelhoff WDYT?

mne_bids/inspect.py
hoechenberger: Resolved in bcb277b

mne_bids/inspect.py
hoechenberger: Resolved in bcb277b

mne_bids/inspect.py
hoechenberger: Removing this would also GREATLY speed up operation

sappelhoff: just to clarify: would that mean that when people just load the FIF and ignore the bids events.tsv, they get the "poor, non-updated data"?

I would be fine with this. It's also how I handle my own (BIDS) data --> my events.tsv contains significantly more information than the raw files, and I don't sync them both every time I add an annotation or similar.

However, I *do* make sure that events.tsv is a superset of what's available in the raw data. :thinking: ... not sure if that's necessary though. Software could handle merging the two sets of reading (i) BIDS sidecar, and (ii) raw data

hoechenberger: > Software could handle merging the two sets of reading (i) BIDS sidecar, and (ii) raw data

This is giving me the chills just thinking about it :)

My proposal would be:

- our write functions should drop remove some the from the raw file that can be repopulated from the sidecars: bads, events (annotations), subject info
- sidecars become the single source of truth

hoechenberger: I've addressed this in bafb6f2

agramfort: BIDS sidecar files have precedence. They are the "truth". To me if they are present they define what are the annotations. If you don't use a BIDS file reader for your files then too bad. You have what is in the manufacturer file as written during acquisition. That's the vision to me. See also discussion with Robert O. recently. He said the same thing.

hoechenberger: > If you don't use a BIDS file reader for your files then too bad.

Yes I totally agree. It would be like complaining that you don't get proper formatting of a modern website if you view it in links on your Linux terminal – you're not using the tool the website was designed for. I guess the unfortunate thing is that users are still able to **open** the files, instead of being told, "hey, unsupported format, use xxx instead"

hoechenberger: > You have what is in the manufacturer file as written during acquisition. That's the vision to me.

Ok I can live with that. Although I'm still concerned that users might e.g. assume a file is anonymized if they simply adjust the sidecars.

hoechenberger: Still – that's **our** opinion. It's not what the specs say. We need to push forward an amendment of the BIDS specification.

sappelhoff: yes, that's probably a good idea.

> our write functions (write_raw_bids, write_anat) should drop all info from the raw file

as long as we do not lose resolution ... example: dropping the BrainVision annotations from vmrk --> these are in samples (1-indexed, I think) ... that means in BIDS sidecar events.tsv there MUST be a column `sample`, which can relate a given event back to the exact sample in the data --> not something in seconds.

also, we cannot drop all data unfortunately --> some files will require things like dates, etc. 

Other than that +1

hoechenberger:
I've now dropped "raw rewriting" from `mark_bad_channels()` too, to have things consistent.

mne_bids/inspect.py
hoechenberger: Ping @agramfort, I would appreciate a screenshot if you have a minute :)

agramfort: Screenshot 2020-11-28 at 20 14 15

hoechenberger: Thanks!

hoechenberger: Which backend is this?

hoechenberger: @agramfort I've pushed a change, could you please try again?

agramfort (approved):
works great !

hoechenberger:
Merge if you feel like it :)

Issue #650, opened 2020-12-03
defacing non T1 image eg FLASH


agramfort:
I've being struggling for way too long to be able to write my FLASH sequence with the same defacing at the T1.

Here is my attempt:

```
import os.path as op

import numpy as np
import matplotlib.pyplot as plt

import mne
from mne.datasets import sample
from nilearn.plotting import plot_anat

from mne.source_space import head_to_mri

from mne_bids import BIDSPath, write_anat

data_path = sample.data_path()
raw_fname = op.join(data_path, 'MEG', 'sample', 'sample_audvis_raw.fif')
events_data = op.join(data_path, 'MEG', 'sample', 'sample_audvis_raw-eve.fif')
output_path = op.join(data_path, '..', 'MNE-sample-data-bids-debug')

raw = mne.io.read_raw_fif(raw_fname)
subject = '01'

###############################################################################
# Get Landmarks from MEG file, 0, 1, and 2 correspond to LPA, NAS, RPA
# and the 'r' key will provide us with the xyz coordinates
pos = np.asarray((raw.info['dig'][0]['r'],
                  raw.info['dig'][1]['r'],
                  raw.info['dig'][2]['r']))

# Load the transformation matrix and show what it looks like
trans_fname = op.join(data_path, 'MEG', 'sample',
                      'sample_audvis_raw-trans.fif')
trans = mne.read_trans(trans_fname)

t1_pos = head_to_mri(pos=pos,
                     subject='sample',
                     mri_head_t=trans,
                     subjects_dir=op.join(data_path, 'subjects'))
mri_pos = t1_pos / 1e3  # put landmarks in meters

montage = mne.channels.make_dig_montage(
    lpa=mri_pos[0],
    nasion=mri_pos[1],
    rpa=mri_pos[2],
    coord_frame='mri'
)

###############################################################################
# FLASH
flash_mgh_fname = \
    op.join(data_path, 'subjects', 'sample', 'mri', 'flash', 'mef05.mgz')

flash_bids_path = \
    BIDSPath(subject=subject, root=output_path, suffix='FLASH')

anat_path = write_anat(
    image=flash_mgh_fname,
    bids_path=flash_bids_path,
    landmarks=montage,
    deface=True,
    overwrite=True,
    verbose=True
)

# Plot it
fig, axs = plt.subplots(3, 1)
for point_idx, label in enumerate(('LPA', 'NAS', 'RPA')):
[...]

PR #651, opened 2020-12-04
[MRG, BUG] Fix deface when images are not in the same space as T1 (e.g. T2)

alexrockhill:
PR Description
--------------

Moves away from landmarks into T1 space which doesn't have to be RAS and keeps images in RAS to deface for any image. Address https://github.com/mne-tools/mne-bids/issues/650.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [x] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -1544,9 +1544,9 @@ def test_write_anat(_bids_validate):
         coord_frame='head')
 
     mri_scanner_ras_landmarks = mne.channels.make_dig_montage(
-        lpa=np.array([-74.53101434, 19.62855196, -52.28881913]),
-        nasion=np.array([-1.89452708, 103.69850393, 4.97122087]),
-        rpa=np.array([72.01202598, 21.09274957, -57.53678129]),
+        lpa=np.array([-0.07453101, 0.01962855, -0.05228882]),
+        nasion=np.array([-0.00189453, 0.1036985, 0.00497122]),
+        rpa=np.array([0.07201203, 0.02109275, -0.05753678]),
         coord_frame='ras')
 
     # test meg landmarks
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -1371,7 +1371,7 @@ def test_write_anat(_bids_validate):
 
     bids_path = BIDSPath(subject=subject_id, session=session_id,
                          acquisition=acq, root=bids_root)
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            raw=raw, trans=trans, deface=True, verbose=True,
                            overwrite=True)
     anat_dir = bids_path.directory
@@ -1409,7 +1409,7 @@ def test_write_anat(_bids_validate):
     # Now try some anat writing that will fail
     # We already have some MRI data there
     with pytest.raises(IOError, match='`overwrite` is set to False'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+        write_anat(t1w_mgh, bids_path=bids_path,
                    raw=raw, trans=trans, verbose=True, deface=False,
                    overwrite=False)
 
@@ -1420,7 +1420,7 @@ def test_write_anat(_bids_validate):
 
     # pass some invalid type as T1 MRI
     with pytest.raises(ValueError, match='must be a path to an MRI'):
-        write_anat(9999999999999, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(9999999999999, bids_path=bids_path, raw=raw,
                    trans=trans, verbose=True, deface=False, overwrite=True)
 
     with pytest.warns(RuntimeWarning, match='Ignoring `raw`'):
@@ -1441,7 +1441,7 @@ def test_write_anat(_bids_validate):
     ex = TypeError
 
     with pytest.raises(ex, match=match):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=wrong_type, verbose=True, deface=False,
                    overwrite=True)
 
@@ -1449,11 +1449,11 @@ def test_write_anat(_bids_validate):
     wrong_fname = 'not_a_trans'
     match = 'trans file "{}" not found'.format(wrong_fname)
     with pytest.raises(IOError, match=match):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=wrong_fname, verbose=True, overwrite=True)
 
     # However, reading trans if it is a string pointing to trans is fine
-    write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+    write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                trans=trans_fname, verbose=True, deface=False,
                overwrite=True)
 
@@ -1467,18 +1467,18 @@ def test_write_anat(_bids_validate):
     # specify trans but not raw
     with pytest.raises(ValueError, match='must be specified if `trans`'):
         bids_path.update(session=session_id)
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=None,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=None,
                    trans=trans, verbose=True, deface=False, overwrite=True)
 
     # test deface
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            raw=raw, trans=trans_fname,
                            verbose=True, deface=True, overwrite=True)
     anat_dir = bids_path.directory
     t1w = nib.load(op.join(anat_dir, 'sub-01_ses-01_T1w.nii.gz'))
     vox_sum = t1w.get_fdata().sum()
 
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            raw=raw, trans=trans_fname,
                            verbose=True, deface=dict(inset=25.),
                            overwrite=True)
@@ -1488,7 +1488,7 @@ def test_write_anat(_bids_validate):
 
     assert vox_sum > vox_sum2
 
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            raw=raw, trans=trans_fname,
                            verbose=True, deface=dict(theta=45),
                            overwrite=True)
@@ -1500,37 +1500,32 @@ def test_write_anat(_bids_validate):
 
     flash_mgh = \
         op.join(data_path, 'subjects', 'sample', 'mri', 'flash', 'mef05.mgz')
-    with pytest.raises(ValueError, match='did not contain "T1"'):
-        write_anat(flash_mgh, bids_path=bids_path, raw=raw,
-                   trans=trans, verbose=True, deface=True, overwrite=True)
+    bids_path.update(suffix='FLASH')
+    with pytest.raises(ValueError, match='`trans` only applies to the T1'):
+        write_anat(flash_mgh, bids_path=bids_path, raw=raw, trans=trans,
+                   deface=True, verbose=True, overwrite=True)
 
-    flash_img = nib.load(flash_mgh)
-    with pytest.raises(ValueError, match='did not contain "T1"'):
-        write_anat(flash_img, bids_path=bids_path, raw=raw,
-                   trans=trans, verbose=True, deface=True, overwrite=True)
-
+    bids_path.update(suffix='T1w')
     with pytest.raises(ValueError, match='must be provided to deface'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+        write_anat(t1w_mgh, bids_path=bids_path,
                    verbose=True, deface=True, overwrite=True)
 
     with pytest.raises(ValueError, match='inset must be numeric'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
-                   trans=trans, verbose=True, deface=dict(inset='small'),
-                   overwrite=True)
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw, trans=trans,
+                   deface=dict(inset='small'), verbose=True, overwrite=True)
 
     with pytest.raises(ValueError, match='inset should be positive'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=trans, verbose=True, deface=dict(inset=-2.),
                    overwrite=True)
 
     with pytest.raises(ValueError, match='theta must be numeric'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=trans, verbose=True, deface=dict(theta='big'),
                    overwrite=True)
 
-    with pytest.raises(ValueError,
-                       match='theta should be between 0 and 90 degrees'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+    with pytest.raises(ValueError, match='theta should be between 0 and 90'):
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=trans, verbose=True, deface=dict(theta=100),
                    overwrite=True)
 
@@ -1561,7 +1556,7 @@ def test_write_anat(_bids_validate):
 
     # test mri voxel landmarks
     bids_path.update(acquisition=acq)
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            deface=True, landmarks=mri_voxel_landmarks,
                            verbose=True, overwrite=True)
     anat_dir = bids_path.directory
@@ -1571,7 +1566,7 @@ def test_write_anat(_bids_validate):
     vox1 = t1w1.get_fdata()
 
     # test mri landmarks
-    bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
+    bids_path = write_anat(t1w_mgh, bids_path=bids_path,
                            deface=True, landmarks=mri_landmarks,
                            verbose=True, overwrite=True)
     anat_dir = bids_path.directory
@@ -1586,7 +1581,7 @@ def test_write_anat(_bids_validate):
 
     # crash for raw also
     with pytest.raises(ValueError, match='Please use either `landmarks`'):
-        write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh, raw=raw,
+        write_anat(t1w_mgh, bids_path=bids_path, raw=raw,
                    trans=trans, deface=True, landmarks=mri_landmarks,
                    verbose=True, overwrite=True)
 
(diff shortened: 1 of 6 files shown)


alexrockhill:
Hey @jasmainak or @adam2392 maybe, any advice on tests for the deface changes, they are failing because the image is converted to RAS when I do the updated deface and I don't really think it's practical to convert it back to it's original format...

adam2392:
> Hey @jasmainak or @adam2392 maybe, any advice on tests for the deface changes, they are failing because the image is converted to RAS when I do the updated deface and I don't really think it's practical to convert it back to its original format...

I didn't write any of the defacing code, so not sure how helpful my comments are. 

Taking a quick look, it looks like its failing because the expected coords are still in original image space? If you "deface" an image, why wouldn't you want to convert back to w/e original format you were in?

E.g. if I run an image through FreeSurfer, it's not in RAS anymore, but maybe I want to deface the `T1.mgz`?

alexrockhill:
> > Hey @jasmainak or @adam2392 maybe, any advice on tests for the deface changes, they are failing because the image is converted to RAS when I do the updated deface and I don't really think it's practical to convert it back to its original format...
> 
> I didn't write any of the defacing code, so not sure how helpful my comments are.
> 
> Taking a quick look, it looks like its failing because the expected coords are still in original image space? If you "deface" an image, why wouldn't you want to convert back to w/e original format you were in?
> 
> E.g. if I run an image through FreeSurfer, it's not in RAS anymore, but maybe I want to deface the `T1.mgz`?

I thought maybe you wrote some of the testing code for the `estimate_trans` parts but I'm just guessing. 

So why it's in RAS is that the image gets converted to RAS so that it works with image files that are not in LIA format, for instance the FLASH in the referenced issue was in LAS I think.

alexrockhill:
@adam2392 The part that it would be great to get advice on is the estimate trans. I think it will work even if the anatomical coordinates are found from the T2 file, but maybe something to think about if there is a T1 and a T2 and their trans estimates differ...

alexrockhill:
Ok, I figured out how to put the image back in its original coordinate frame and it works for both the flash and the T1 in the example dataset. This PR is ready for review/okay to merge by me. The only outstanding issue is whether to raise an error if there is a T1 and a T2 and they yield different trans estimates. I think that error would be very helpful personally, but I think it should probably be another PR.

alexrockhill:
I was wrong, it didn't work for different sized images (e.g. the test images) so I changed back to voxel space and this worked for everything so hopefully should be okay.

agramfort:
thanks a lot @alexrockhill !

I tried to deface both the T1 and the flash. Flash is now reasonable but the oblique plane
seems different. See:

Screenshot 2020-12-04 at 10 27 39

can you replicate?

can you update an example converting the anat so we can see the artifact to check things?

alexrockhill:
> thanks a lot @alexrockhill !
> 
> I tried to deface both the T1 and the flash. Flash is now reasonable but the oblique plane
> seems different. See:
> 
> Screenshot 2020-12-04 at 10 27 39
> 
> can you replicate?
> 
> can you update an example converting the anat so we can see the artifact to check things?

Yeah I noticed that too, I'll see if I can figure out what happened.

alexrockhill:
It looks like the T2 has off-diagonal components in the affine which I think is lossy when it is converted and reverted with this method:
```
array([[ -0.0246,  -0.0303,  -0.9992,  97.68  ],
       [ -0.9778,   0.2086,   0.0178, 105.7673],
       [ -0.2079,  -0.9775,   0.0347, 121.1166],
       [  0.    ,   0.    ,   0.    ,   1.    ]])
```

hoechenberger:
Sorry, late to the party and haven't really looked at the code yet :) But I wonder why you're not using `nib.as_closest_canonical()` here?

If you're stuck, I can try to have a closer look too!

alexrockhill:
> Sorry, late to the party and haven't really looked at the code yet :) But I wonder why you're not using `nib.as_closest_canonical()` here?
> 
> If you're stuck, I can try to have a closer look too!

I tried that but they you can't go back to save the image in the defaced form in its former orientation...

alexrockhill:
Ok so in this version, it round trips to RAS and back successfully and the code is more succinct but I think maybe that means the save round trip modifies the FLASH orientations... checking now.

alexrockhill:
Huh seems to round trip okay for saving too...

alexrockhill:
Ahhh so it looks like even without deface, the FLASH image gets rotated when using `plot_anat`, it just wasn't in the example

agramfort:
the defacing planes appear different for me between T1 and flash:

Screenshot 2020-12-04 at 17 45 19

do you confirm?

agramfort:
defacing in
https://3756-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_mri_and_trans.html#sphx-glr-auto-examples-convert-mri-and-trans-py
now cuts some part of the brain

>

alexrockhill:
Yeah, I'm seeing the same but that's because it is oblique to the cardinal directions, AFAIK the only way to solve that would be cubic interpolation and back but it's not implemented in nibabel high-level APIs to my knowledge.

Screen Shot 2020-12-04 at 8 50 15 AM

Is the issue that the planes are different or that the resulting FLASH has a shifted coordinate frame? I think I was only imagining it being shifted, I think the round trip save works, it is a very subtle shift, my bad!

agramfort:
thanks for your effort @alexrockhill 

T1 deface in https://3761-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_mri_and_trans.html#sphx-glr-auto-examples-convert-mri-and-trans-py is now too extreme. Cuts inside the brain.

see how it looks in master: https://mne.tools/mne-bids/dev/auto_examples/convert_mri_and_trans.html

also having to change the theta seems unnecessary and error prone if one has a full study to convert.

did you work with someone else during the code sprint on this piece of code. maybe you can get help there. I don't know this code enough to be able to help quickly

alexrockhill:
I'm reworking it now, I haven't quite got it working but it's very close by applying the affine to the indices to get rid of the oblique offset component. I think I'll have time to finish tonight here (in like 6-10 hours).

agramfort:
thanks !

>

alexrockhill:
So I'm getting to where the new issue is, which isn't actually the deface algorithm itself but loading in MRI images...

When I apply the trans to the landmarks stored in `raw` I get:

```
array([[-69.25741411,  10.58946009, -25.00085936],
       [  3.37909386,  94.65942693,  32.25916435],
       [ 77.28562067,  12.05367187, -30.24882153]])
```

Which is here in the image: 

Screen Shot 2020-12-06 at 3 29 36 PM

Then if I apply the inverse of the `img_mgh.get_vox2ras_tkr` function, the nasion lines up correctly.

This causes issues because the `vox2ras_tkr` and the `image.affine` are different, whereas for the T1 they are: 

```
array([[  -1.,    0.,    0.,  128.],
       [   0.,    0.,    1., -128.],
       [   0.,   -1.,    0.,  128.],
       [   0.,    0.,    0.,    1.]], dtype=float32)
```

and

```
array([[  -1.        ,    0.00000012,   -0.00000019,  122.72639465],
       [   0.00000009,    0.00000002,    1.        , -118.96092987],
       [   0.00000001,   -1.        ,    0.00000001,  100.71203613],
       [   0.        ,    0.        ,    0.        ,    1.        ]])
```

respectively.

So the issue seems to be that either the fiducial coordinates are wrong in RAS space and then we can use the affine transform instead of the header transform or everything is messed up in a way that I don't know how to fix. Maybe it would help to know why @larsoner suggested to use the mgh image in the first place if you have a second.

alexrockhill:
For the FLASH, applying the inverse of `vox2ras_tkr` to the aforementioned RAS coordinates that are slightly off to gives:

```
array([[197.25741411, 153.00085936, 106.58946009],
       [124.62090614,  95.74083565, 190.65942693],
       [ 50.71437933, 158.24882153, 108.05367187]])
```

which the nasion is not even in the voxel space of the flash image which is of shape `(256, 256, 192)`

alexrockhill:
Ok here's where I'm leaving it until I understand why the landmarks are expressed in `vox2ras_tkr` coordinates and not `vox2ras` coordinates. As it is, both examples work but cut into the brain because they use the several-cm-off naison shown above which is what you get when you apply the `vox2ras` instead of the `vox2ras_tkr`.

What I think I have figured out after reading about freesurfer coordinate systems that the landmarks are expressed in T1 surface coordinates which I thought should be an equivalent space to FLASH surface coordinates (as compared to the scanner spaces which should be different) but for some reason the tkr spaces are not equivalent and that is causing the issues. That's from the pdf on this link https://surfer.nmr.mgh.harvard.edu/fswiki/CoordinateSystems. I really don't know how to solve this part.

alexrockhill:
Hmmm also I think there may be a solution if you always supply the T1.

Then you can get the correct TkReg RAS (T1) and from there you can get the RAS which can then be converted into TkReg RAS (mef05).

alexrockhill:
So I think there is a disconnect between the coregistration GUI/process in mne-python which relies on the freesurfer subjects directory and the `mne_bids.write_anat` function which does not. The `trans` or landmarks provided (assuming they are not already in `FIFF.FIFFV_MNE_COORD_MRI_VOXEL`) are transformed (or provided already transformed) from meg to surface RAS space. The issue is that the FLASH and other scans, have different surface RAS coordinates and so need to be coregistered to the T1. From this example https://mne.tools/dev/auto_tutorials/source-modeling/plot_background_freesurfer_mne.html, `mne-python` stores everything as surface RAS coordinates because of all the source estimation reliance on `freesurfer` so when a FLASH scan is provided that is not in the same space, there is no way of transforming coordinates from freesurfer surface coordinates to the FLASH scanner coordinates because the freesurfer coordinates are not provided to `mne_bids.write_anat`. So, I think there are three options:

1) Not support images that aren't in freesurfer surface space for defacing (not very useful)
2) Provide the T1 as well when defacing to define the freesurfer surface coordinates (a bit cumbersome)
3) Redefine landmarks in `mne-python` so that they are in scanner coordinates (a huge undertaking)

I think 2 is probably the best option.

alexrockhill:
Ok, I think this is a pretty good solution, both the FLASH and the T1 deface the same and well and most everything else is left without modification.

Note: I think the saving of the fiducials with the FLASH previously was wrong, it used the FLASH vox2ras_tkr which wasn't in freesurfer surface RAS. Maybe something to check in the future, and worth noting that you need the T1 if you're going to write the json for the FLASH.

agramfort:
@alexrockhill ping me on hangout when you can so we can discuss this. Maybe this will help

alexrockhill:
Changing to support landmarks in MRI scanner RAS for users with a T1 but no recon and for users with recon, allow passing `raw` (for dig), `trans` and `t1w` if the file is not already a T1 later today.

alexrockhill:
Ok @agramfort and @hoechenberger this is ready and I think I understand it this time ;) If you could review that would be great.

PS I didn't end up using the scanner RAS coordinate system for the FLASH, I thought that would require them knowing the difference between scanner and surface RAS and was probably more confusing than just using voxels. The precision is slightly worse, but I think it's okay for this purpose.

agramfort:
thanks @alexrockhill 

some remarks:
- you added the t1w param which I had removed (without deprecation cycle hum...). See stablehttps://mne.tools/mne-bids/stable/generated/mne_bids.write_anat.html.
- now you suggest to pass another set of fiducials

to help here I've put together a script with different equivalent options that I would imagine should work

https://www.dropbox.com/s/zc1kf2odvjevwdj/debug_flash.py?dl=0

let's discuss if this can help. Also it seems the defacing of T1 has changed a bit with current master if deface=True.

agramfort:
I would vote to remove t1w param and stick to the 5 use cases in my script

regarding the slight change the new way is worse than what we have in
master.
If we can avoid this change I would vote for it.

>

alexrockhill:
The last commit (https://github.com/mne-tools/mne-bids/pull/651/commits/0c92c9cec4c3dfadc677610ed90b2fbe7514fc15) has the `t1w` option so that there is a working version with it in case there is ever a need to go back. Next commit I'll delete it.

alexrockhill:
Oh man re the change in angle reference, I didn't really understand how the plane of the fiducials were used to calculate when only the nasion was used and it was in voxel space when Eric and I did it, that was more of a him thing. The only landmark that is used is the nasion so I'm not sure how that works. Also this was before I applied the affine to the coordinate indices so that might change the ability to do that.

alexrockhill:
Hmmm tests failed for saving landmarks to 'ras'... should that be allowed in an mne-python PR?

agramfort:
ping @larsoner

larsoner:
This should be possible in `master` already and should come out with 0.22 in a week or so

agramfort:
why do you want to save landmarks to 'ras' ?

>

alexrockhill:
> why do you want to save landmarks to 'ras' ?
> […](#)

In the tests, that's how all the other landmarks are tested that they work. I think it's done that way to double dip and test the landmark loading at the same time. I could just change it not to test that for now...

agramfort:
the landmarks in the json?

>

alexrockhill:
> the landmarks in the json?
> […](#)

```
meg_landmarks = mne.channels.make_dig_montage(
        lpa=[-7.13766068e-02, 0.00000000e+00, 5.12227416e-09],
        nasion=[3.72529030e-09, 1.02605611e-01, 4.19095159e-09],
        rpa=[7.52676800e-02, 0.00000000e+00, 5.58793545e-09],
        coord_frame='head')
tmp_dir = _TempDir()
meg_landmarks.save(op.join(tmp_dir, 'meg_landmarks.fif'))
bids_path = write_anat(t1w_mgh, bids_path=bids_path, t1w=t1w_mgh,
                       deface=True, trans=trans,
                       landmarks=op.join(tmp_dir, 'meg_landmarks.fif'),
                       verbose=True, overwrite=True)
anat_dir = bids_path.directory
_bids_validate(bids_root)
```

In a DigMontage

agramfort:
for the test just pass without saving in the case of RAS

alexrockhill:
Ok so here is with the `t1w` parameter removed. My preference would definitely be reverting to 13493a3, as I think this pushes unnecessary extraction of the digitization and transformation coordinates onto the user when before they were under the hood. But I would also like to get this moving and I don't think this version is bad, it's a definite improvement on the current master branch.

agramfort:
the debug script I shared with you now works for you?

>

alexrockhill:
> the debug script I shared with you now works for you?
> […](#)

Only if you take out the conversion to meters for the ras coordinates, then yes.

agramfort:
Can you share the correct / fixed script ?

alexrockhill:
```
import os.path as op

import numpy as np
import matplotlib.pyplot as plt

import mne
from mne.datasets import sample
from nilearn.plotting import plot_anat
import nibabel as nib

from mne.source_space import head_to_mri

from mne_bids import BIDSPath, write_anat

data_path = sample.data_path()
raw_fname = op.join(data_path, 'MEG', 'sample', 'sample_audvis_raw.fif')
events_data = op.join(data_path, 'MEG', 'sample', 'sample_audvis_raw-eve.fif')
output_path = op.join(data_path, '..', 'MNE-sample-data-bids-debug')

subject = '01'

# Get Landmarks from MEG file, 0, 1, and 2 correspond to LPA, NAS, RPA
# and the 'r' key will provide us with the xyz coordinates
head_pos = np.asarray((raw.info['dig'][0]['r'],
                       raw.info['dig'][1]['r'],
                       raw.info['dig'][2]['r']))

# Load the transformation matrix and show what it looks like
trans_fname = op.join(data_path, 'MEG', 'sample',
                      'sample_audvis_raw-trans.fif')
trans = mne.read_trans(trans_fname)

fid_pos = head_to_mri(pos=head_pos,
                      subject='sample',
                      mri_head_t=trans,
                      subjects_dir=op.join(data_path, 'subjects'))
ras_pos = fid_pos

montage_ras = mne.channels.make_dig_montage(
    lpa=ras_pos[0],
    nasion=ras_pos[1],
    rpa=ras_pos[2],
    coord_frame='ras'
)

# Get the path to our MRI scan
t1_mgh_fname = op.join(data_path, 'subjects', 'sample', 'mri', 'T1.mgz')

t1_mri_hdr = nib.load(t1_mgh_fname).header
t1_vox_pos = mne.transforms.apply_trans(t1_mri_hdr.get_ras2vox(), ras_pos)

montage_t1_vox = mne.channels.make_dig_montage(
    lpa=t1_vox_pos[0],
    nasion=t1_vox_pos[1],
    rpa=t1_vox_pos[2],
    coord_frame='mri_voxel'
)

# Plot util


def plot_img(img_path):
    fig, axs = plt.subplots(3, 1)
    for point_idx, label in enumerate(('LPA', 'NAS', 'RPA')):
        plot_anat(str(img_path.fpath), axes=axs[point_idx],
                  cut_coords=ras_pos[point_idx, :],
                  title=label, vmax=800)
    plt.show()


# Option 1 : Write T1 using T1.mgz and trans

t1w_bids_path = \
    BIDSPath(subject=subject, root=output_path, suffix='T1w')

img_path = write_anat(
    image=t1_mgh_fname,  # path to the MRI scan
    bids_path=t1w_bids_path,
    raw=raw,  # the raw MEG data file connected to the MRI
    trans=trans,  # our transformation matrix
    deface=True,
    overwrite=True,
    verbose=True  # this will print out the sidecar file
)

plot_img(img_path)

# ###############################################################################
# # Option 2 : Write T1 using T1.mgz and landmarks in RAS

img_path = write_anat(
    image=t1_mgh_fname,  # path to the MRI scan
    bids_path=t1w_bids_path,
    landmarks=montage_ras,
    deface=True,
    overwrite=True,
    verbose=True  # this will print out the sidecar file
)

plot_img(img_path)

# Option 3 : Write T1 using T1.mgz and landmarks in VOX

img_path = write_anat(
    image=t1_mgh_fname,  # path to the MRI scan
    bids_path=t1w_bids_path,
    landmarks=montage_t1_vox,
    deface=True,
    overwrite=True,
    verbose=True  # this will print out the sidecar file
)

plot_img(img_path)

# ###############################################################################
# # Option 4 : Write FLASH using mef05.mgz and landmarks in RAS

flash_mgh_fname = \
    op.join(data_path, 'subjects', 'sample', 'mri', 'flash', 'mef05.mgz')

flash_bids_path = \
    BIDSPath(subject=subject, root=output_path, suffix='FLASH')

img_path = write_anat(
    image=flash_mgh_fname,
    bids_path=flash_bids_path,
    landmarks=montage_ras,
    deface=True,
    overwrite=True,
    verbose=True  # this will print out the sidecar file
)

plot_img(img_path)

# Option 5 : Write FLASH using mef05.mgz and landmarks in VOX

flash_mri_hdr = nib.load(flash_mgh_fname).header
flash_vox_pos = mne.transforms.apply_trans(
    flash_mri_hdr.get_ras2vox(), ras_pos)

montage_flash_vox = mne.channels.make_dig_montage(
    lpa=flash_vox_pos[0],
    nasion=flash_vox_pos[1],
    rpa=flash_vox_pos[2],
    coord_frame='mri_voxel'
)

img_path = write_anat(
    image=flash_mgh_fname,
    bids_path=flash_bids_path,
    landmarks=montage_flash_vox,
    deface=True,
    overwrite=True,
    verbose=True  # this will print out the sidecar file
)

plot_img(img_path)
```

alexrockhill:
Ok @agramfort, I added back the `t1w` argument but now it is `None` by default and I think it's handled very gracefully and now the functionality covers any possible way someone would want to pass coordinates:

1) T1, raw and trans
2) T1 head landmarks and trans
3) T1 mri landmarks
4) T1 scanner RAS
5) T1 voxel space
6) FLASH, raw, trans and T1
7) FLASH head landmarks, trans, T1
8) FLASH mri landmarks, T1
9) FLASH scanner RAS
10) FLASH voxel space

Those are all now supported, and this commit adds back support for 6-8 without checking if the name is T1 but using the `bids_path.suffix`.

hoechenberger:
Amazing!! 😻

alexrockhill:
Okay to merge?

agramfort:
let me do a pass this we or ealy next week but it did the job for openneuro

see

https://openneuro.org/datasets/ds000248/versions/1.2.4

agramfort:
@alexrockhill I did a pass. See my commit.

You'll see that one test is failing. I don't get why I don't get the same landmarks positions
if I pass the landmarks in surface RAS.

thanks a lot !

alexrockhill:
Ok that should fix the test, apparently one of the landmark's numbers was off by a bit but it was small enough you couldn't see it in the figures maybe.

The example looks 10x better, I really should learn a bit more of the markdown notes and stuff. Tests are also more comprehensive, thanks, looks great!

mne_bids/tests/test_write.py
```diff
@@ -1544,11 +1543,10 @@ def test_write_anat(_bids_validate):
...
         coord_frame='head')
 
-    s = 1  # XXX to me it should 1e3 as it shoult be in meters
     mri_scanner_ras_landmarks = mne.channels.make_dig_montage(
-        lpa=np.array([-74.53101434, 19.62855196, -52.28881913]) / s,
-        nasion=np.array([-1.89452708, 103.69850393, 4.97122087]) / s,
-        rpa=np.array([72.01202598, 21.09274957, -57.53678129]) / s,
+        lpa=np.array([-74.53101434, 19.62855196, -52.28881913]),
+        nasion=np.array([-1.89452708, 103.69850393, 4.97122087]),
+        rpa=np.array([72.01202598, 21.09274957, -57.53678129]),
```
agramfort, line 1549: what is the unit of the coordinates in RAS? if it's a measure of distance then we should
use meters to be consistent with MNE conventions.

agramfort:
some CIs are red. no idea why.

here is the rendered doc now https://3805-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_mri_and_trans.html#sphx-glr-auto-examples-convert-mri-and-trans-py

mne_bids/tests/test_write.py
alexrockhill: It's mm so we could convert

alexrockhill:
Hmmm so the warnings both had something to do with copying brainvision files so maybe `pybv` has changed. I just filtered them, hopefully that will fix the CI.

alexrockhill:
Whew okay finally the weird warnings from brainvision files are filtered out. Good by me.

agramfort (approved):

agramfort:
thx a ton @alexrockhill !

larsoner:
I did not look too deeply but the code changes look reasonable to me

Issue #659, opened 2020-12-16
Unable to convert EDF -> BV anymore?


adam2392:
#### Describe the bug
I was doing another conversion again for EDF source iEEG files, and I noticed that I can't convert to BV anymore, even if I pass back anonymization daysback. 

In the BIDs spec, it is said that ~~BV~~ BV and EDF are the RECOMMENDED format for iEEG data.

Is this intended behavior? I built a lot of func that required me to assume that data was stored in BV. Wiil it be possible to add a `convert` kwarg to `write_raw_bids` that converts the said dataset into the `RECOMMENDED` format?

hoechenberger:
Is this because of #636?

adam2392:
> Is this because of #636?

Just took a look, and yes I believe so due to these lines: 

https://github.com/a-hurst/mne-bids/blob/4b65b40911b092026ebb39dc1f9d0cd5b4f8fd76/mne_bids/write.py#L1116-L1122

adam2392:
WDYT, is it okay to support "converting to RECOMMENDED" formats?

hoechenberger:
Yes I think there should be a switch / parameter to toggle this…

@sappelhoff @agramfort @jasmainak WDYT?

sappelhoff:
yes, a switch to toggle conversion if it's wanted could be nice for some users.

jasmainak:
fine by me!

adam2392:
> WDYT, is it okay to support "converting to RECOMMENDED" formats?

Sorry, just re-read the specification, and I suppose EDF and BV are both "RECOMMENDED" formats for iEEG. However, it still brings up an issue because I was pretty used to working w/ BV files, and ideally would like to stick w/ that.

If we wanted to "convert" files from EDF -> BV like this, would it be best to add a kwarg to `write_raw_bids`, or add another function for converting data files? For example:

```Python
def convert_datafile(bids_path, to_format='.vhdr'):
   # convert say EDF to .vhdr and then delete the EDF file
```

hoechenberger:
> If we wanted to "convert" files from EDF -> BV like this, would it be best to add a kwarg to `write_raw_bids`, or add another function for converting data files?

Depends on what you want to do… convert an existing file, no matter whether it's in a BIDS folder structure, and write it anywhere? Store it as BIDS? Or take an existing BIDS dataset and convert it to another BIDS dataset with a different format?

sappelhoff:
[...]

PR #670, opened 2020-12-31
[MRG] Update pybv to write non-voltage channels

adam2392:
PR Description
--------------

Closes: #664 

Once pybv 0.5 is released, the setup.py, requirements.txt file and README can be updated.

Merge checklist
---------------

Maintainer, please confirm the following before merging:

- [x] All comments resolved
- [x] This is not your own PR
- [x] All CIs are happy
- [x] PR title starts with [MRG]
- [x] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/master/doc/whats_new.rst) is updated
- [x] PR description includes phrase "closes <#issue-number>"

PR #672, opened 2021-01-04
[MRG] Allow write_raw_bids to specify a file format explicitly

adam2392:
PR Description
--------------

Closes: #659 

Note: EDF format is not supported due to the lack of a robust EDF format writer.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1274,8 +1274,6 @@ def write_raw_bids(raw, bids_path, events_data=None,
         elif all(format not in values for values in CONVERT_FORMATS.values()):
             raise ValueError(f'The input "format" {format} is not an '
                              f'accepted input format for `write_raw_bids`. '
-                             f'For example, if you passed in "EDF", mne-bids '
-                             f'does not currently have conversion to EDF. '
                              f'Please use one of {CONVERT_FORMATS[datatype]} '
                              f'for {datatype} datatype.')
         elif format not in CONVERT_FORMATS[datatype]:
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1007,8 +1007,8 @@ def write_raw_bids(raw, bids_path, events_data=None,
         If `'auto'` (default), there is no explicit change in how the file is
         converted to BIDS. The native format is used. If a str, then will
         accept RECOMMENDED BIDS file formats. For example, EEG and iEEG
-        RECOMMENDED formats are ``BrainVision`` and ``EDF``. If ``'BrainVision'``
-        is passed in, then file will be converted to the
+        RECOMMENDED formats are ``BrainVision`` and ``EDF``. If
+        ``'BrainVision'`` is passed in, then file will be converted to the
         ``BrainVision`` file format even if the file is originally in ``EDF``
         file format. For MEG, ``'FIF'`` will convert file to ``FIF`` format.
     overwrite : bool
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1097,11 +1097,6 @@ def write_raw_bids(raw, bids_path, events_data=None,
         raise RuntimeError('You passed event_id, but no events_data NumPy '
                            'array. You need to pass both, or neither.')
 
-    if format not in CONVERT_FORMATS:
-        raise ValueError(f'The "format" {format} is not an '
-                         f'accepted format. Please use one of '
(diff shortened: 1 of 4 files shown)


agramfort (approved):
LGTM

adam2392:
Not sure what's up w/ the CI but everything works locally and increased the test coverage.

mne_bids/write.py
```diff
@@ -1095,6 +1104,11 @@ def write_raw_bids(raw, bids_path, events_data=None,
         raise RuntimeError('You passed event_id, but no events_data NumPy '
                            'array. You need to pass both, or neither.')
 
+    if format not in CONVERT_FORMATS:
+        raise ValueError(f'The "format" {format} is not an '
+                         f'accepted format. Please use one of '
+                         f'{CONVERT_FORMATS}')
```
sappelhoff, line 1110: This check should come later after line 1286 IMHO ... for example if a user specifies `BrainVision` but has MEG data, we should raise an error as well.

Finally, the current warning is confusing: you are talking about "accepted format", which will remind users of accepted formats in BIDS ... but what **you** actually mean is accepted input parameters to the write_raw_bids function.

Imagine I pass "EDF" ... you tell me it's not an accepted format ... but I know that it is. --> the warning msg should be something like "not supported input, please use ... "

sappelhoff (changes requested):
LGTM in principle. I suggest we wait for #670, one other suggestion, and one requested change.

mne_bids/write.py
```diff
@@ -1254,6 +1263,21 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
+            bids_path.update(extension='.vhdr')
+        elif format == 'FIF' and bids_path.datatype == 'meg':
+            convert = True
+            bids_path.update(extension='.fif')
+        elif format not in CONVERT_FORMATS:
+            raise ValueError(f'The input "format" {format} is not an '
+                             f'accepted input format for `write_raw_bids`. '
+                             f'For example, if you passed in "EDF", mne-bids '
+                             f'does not currently have conversion to EDF. '
+                             f'Please use one of {CONVERT_FORMATS}.')
```
adam2392, line 1279: @sappelhoff

mne_bids/write.py
adam2392: This is fixed in the latest commit. Lmk wyt.

agramfort (approved):
@sappelhoff floor is yours

mne_bids/write.py
sappelhoff: Looks better, but you are not catching the case where format is Brainvision, and Datatype is meg .... Or format is for and Datatype is eeg.

The rest looks good

sappelhoff (commented):
One more comment

mne_bids/write.py
adam2392: added in new commit.

mne_bids/write.py
```diff
@@ -1276,6 +1271,12 @@ def write_raw_bids(raw, bids_path, events_data=None,
         elif format == 'FIF' and bids_path.datatype == 'meg':
             convert = True
             bids_path.update(extension='.fif')
+        elif format not in CONVERT_FORMATS:
+            raise ValueError(f'The input "format" {format} is not an '
+                             f'accepted input format for `write_raw_bids`. '
+                             f'For example, if you passed in "EDF", mne-bids '
+                             f'does not currently have conversion to EDF. '
```
hoechenberger, line 1278: I don't think this example is necessary, or is it? I assume it would only cause confusion

hoechenberger (approved):

sappelhoff (approved):
thanks a lot!

Just one conflict, then good to go!

hoechenberger:
I've resolved the conflict, you can merge once CI goes green

agramfort:
@adam2392 https://github.com/mne-tools/mne-bids/pull/672/checks?check_run_id=1655522660 failed

hoechenberger:
I've restarted CI, the error @agramfort noticed is sometimes caused by flaky servers when we're downloading the test data. Let's see…

sappelhoff:
>  flaky servers when we're downloading the test data.

I see an increasing frequency of this happening ... better to keep an eye on it.

PR #680, opened 2021-01-14
Fix bug if no event descriptions are present; support stim_type column in events.tsv

hoechenberger:
If no event descriptions could be generated, we'd crash as we'd try
to access descriptions like an array, while our placeholder would only
be a string. This is fixed now.

The other part of this PR is related to the naming of the column in the
events.tsv file that contains the event description. According to
BIDS, this column is optional and shall be named `trial_type`. Now
I've come across multiple (older?) datasets where the column is
called `stim_type` instead. I've now added support for reading this
column if `trial_type` is not present. A warning will be issued in
this case, informing the user that the column name should be changed.

This was necessary to read the ds000117 datset from
https://openneuro.org/datasets/ds000117

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -31,7 +31,7 @@ Detailed list of changes
 Enhancements
 ^^^^^^^^^^^^
 
-- xxx
+- Some datasets out in the real world have a non-standard ``stim_type`` instead of a ``trial_type`` column in ``*_events.tsv``. :func:`mne_bids.read_raw_bids` now makes use of this column, and emits a warning, encouraging users to rename it, by `Richard Höchenberger`_ (:gh:`680`)
 
 API changes
 ^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


sappelhoff (approved):
I don't know where "stim_type" comes from, but I think it's fair enough to handle it as proposed in this PR.

hoechenberger:
1 job failing due to download problems.

agramfort (approved):
LGTM

+1 for MRG when CIs are green

hoechenberger:
Since the download problem seems to be persistent (@adam2392 was running into this issue too), and given that all but that one CI runs are green, this should be good to merge.

sappelhoff:
thanks @hoechenberger

PR #681, opened 2021-01-16
fix stats.count_events for old files with stim_type rather than trial_type

agramfort:


hoechenberger:
Great, thanks @agramfort!

PR #684, opened 2021-01-19
MRG: Fix scans.tsv reading

hoechenberger:
I cannot fully reproduce this, but @dnacombo was experiencing issues
when trying to read a raw file, as reading of `scans.tsv` failed. Reason
was that NumPy's `loadtxt` couldn't cope with the `BIDSPath`. Passing
a `pathlib.Path` instead fixed the problem.

Like I said, I could not reproduce this problem locally, but passing a
`Path` instead of a `BIDSPath` is probably a good idea anyway.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -529,8 +529,8 @@ def read_raw_bids(bids_path, extra_params=None, verbose=True):
         subject=bids_path.subject, session=bids_path.session,
         suffix='scans', extension='.tsv',
         root=bids_path.root
-    )
-    if scans_fname.fpath.exists():
+    ).fpath
+    if scans_fname.exists():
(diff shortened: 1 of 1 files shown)


sappelhoff (commented):
no objection, but if it doesn't fix or improve anything concrete, it feels weird to change it.

hoechenberger:
> no objection, but if it doesn't fix or improve anything concrete, it feels weird to change it.

Yes… I mean I could spend some more time trying to reproduce this. We observed this problem on @dnacombo's system after updating to MNE-BIDS `master`, so I'm pretty clueless what's going on actually :(

agramfort:
Let's move on

PR #689, opened 2021-01-21
minor, maint: remove unneeded pragma: no cover

sappelhoff:
The "pragmas" I removed are already covered here:

https://github.com/mne-tools/mne-bids/blob/717a39037def106fa841d49bb4cb5b9f2bd2a2c8/setup.cfg#L46-L47


agramfort (commented):
@larsoner the floor is yours

PR #690, opened 2021-01-21
[MRG] generate intersphinx links for numpy docstrs

sappelhoff:
I realized that none of our nice intersphinx links work in the API docs, see for yourselves: https://mne.tools/mne-bids/stable/generated/mne_bids.write_anat.html#mne_bids.write_anat

This should fix it (` numpydoc_xref_param_type = True`)

~Do not merge yet, please - I may find one or two other things to update.~

EDIT: okay, done updating.

- I also overhauled the MANIFEST.in file (see: https://docs.python.org/2/distutils/sourcedist.html#commands)
- I removed an unneeded windows BAT file (windows users who want to build the docs can use the makefile, by installing GNU make e.g.via chocolatey)
- I updated the sphinx makefile to the new standard they use


agramfort (approved):

sappelhoff:
I have no clue why, but it keeps complaining about two items:

>mne-bids/mne_bids/copyfiles.py:docstring of mne_bids.copyfiles.copyfile_bti:8: WARNING: py:obj reference target not found: mne.io.raw

and 

>mne-bids/mne_bids/read.py:docstring of mne_bids.read_raw_bids:29: WARNING: py:obj reference target not found: mne.io.raw

this is very peculiar, because for every other instance where we use `mne.io.raw`, it *does* work ... and for the case of the failure in the `read_raw_bids` function it's even more odd, because the function defined right above that one *also* has an `mne.io.raw` param, and it works ...

once more calling @larsoner for help :sob:

mne_bids/copyfiles.py
```diff
@@ -545,7 +545,7 @@ def copyfile_bti(raw, dest):
 
     Parameters
     ----------
-    raw : instance of Raw
+    raw : mne.io.raw
```
larsoner, line 548: There is no `mne.io.raw`, but there is a `mne.io.Raw`

sappelhoff: > There is no mne.io.raw, but there is a mne.io.Raw

uh, good point - thanks! But then I am wondering why it didn't warn me for far more entries? As you can see in https://github.com/mne-tools/mne-bids/pull/690/commits/88dadc20dec689779f27bf6b71f8742e8d05365d, I had it wrong many more times than just the two that raised a warning :thinking:

larsoner: Maybe it only warns once per object-not-found?

sappelhoff: possibly :thinking: thanks a lot anyhow, it's passing :-)

hoechenberger (approved):

hoechenberger:
Thanks, @sappelhoff!

sappelhoff:
Happy to see it in! :partying_face:

PR #708, opened 2021-02-13
MRG: Ensure write_raw_bids() stores original event IDs in events.tsv

### LINKED ISSUE #707: _read_events() changes event values
Currently I use following procedure to write a BIDS dataset (simplified): 
```
raw = mne.io.read_raw_fif(raw_file,preload=False)
events = mne.find_events(raw)
bids_basename = mne_bids.BIDSPath(root = output_path, subject=f'{sub:02d}',session=f'{ses:02d}',task=f'{task}')
mne-bids.write_raw_bids(raw, bids_basename,event_id = event_id[task], events_data = events,overwrite=True)
```
Doing so, the event values that are being written to `events.tsv`are not the ones that are defined in the meg file. Specifically, it seems that the original values are ignored and instead the number of unique codes is simply used incrementally. I started with following events: `1, 2, 4, 8, 16, 512, 2048` and ended up with `1, 2, 3, 4, 5, 6, 7`.
The events in the meg file are unchanged.

From looking at [read.py](https://github.com/mne-tools/mne-bids/blob/master/mne_bids/read.py), the problem seems to be that during conversion of events to annotations the original values get lost (see [here](https://github.com/mne-tools/mne-bids/blob/d65b8ca3edf38a4be5a1db8d24592cf2a1a3d5ef/mne_bids/read.py#L131)). So, there is a good chance that might be an issue with `mne` rather than `mne_bids`

hoechenberger:
Fixes #707

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [x] All CIs are happy


sappelhoff (approved):
nice, thanks for the report @eort and the fix @hoechenberger

Issue #678, opened 2021-01-07
[maint] Possible refactor/organization of unit tests for readability and decreased time (if possible)


adam2392:
"... monitor test time when you add tests. Add such big parametrize has a cost in CI time and needs to be fully justified.  It's a general statement. I'd to avoid having mne-bids having a test time lasting tens of minutes. I find it already quite long actually"

_Originally posted by @agramfort in https://github.com/mne-tools/mne-bids/pull/675#discussion_r553567389_

Figured this is actually pretty useful to log.

sappelhoff:
related: #212

PR #710, opened 2021-02-15
list split files separately in scans.tsv

### LINKED ISSUE #692: scans.tsv for split files
Hi, 
I was running the bids-validator on my data and stumbled upon a BIDS-incompatibility. The information in the `scans.tsv` does not match with the actual scans present in the directory, when the  scans consist of split files. To illustrate, Ihere is the `scans.tsv` file:
~~~
filename	                                        acq_time
meg/sub-01_ses-02_task-ml_meg.fif	                2020-10-23T12:20:27
meg/sub-01_ses-02_task-rdm_meg.fif	                2020-10-23T11:25:01
meg/sub-01_ses-02_task-restingstate_meg.fif	        2020-10-23T10:30:32
meg/sub-01_ses-02_task-vbm_meg.fif	                2020-10-23T10:49:34
~~~

 and a screenshot of the corresponding directory:
image

For all the split files, the bids-validator had following issue: 
~~~
Error 2: [Code 129] SCANS_FILENAME_NOT_MATCH_DATASET
The filename in scans.tsv file does not match what is present in the BIDS dataset.
~~~

Not sure whether this is or the bids-validator-repo would be the place to report this. If more appropriate there, I could move the issue. 

Thanks

eort:
closes #692 

The changes in `read.py` should be clear. In `write.py` it is somewhat more complicated. I didn't know how to infer whether there are split files based on the bids_path, because the splitting is done in mne during saving. What I did now is using the number of files in the `raw.filenames` attribute of the `raw`object, which contains the names of all files that were loaded into this raw instance. That is not ideal, because if for some reason someone wants to combine two files that won't exceed the single-file-size limit, the method would split write split files into `scans.tsv`. I don't think this scenario is very likely though. 

Alternatively, one could try to infer based on the files that have been written, whether a filename is to be split or not. What do you think? 

Aside of that, I couldn't run all tests, because there were missing files for some tests in `test_write.py`: 
- `test_kit()`
- `test_bci()`
- `test_vhdr()`

However, those tests also failed even before I added my changes, so I don't think it has to do with this commit. All other tests passed. 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -787,11 +787,54 @@ def test_write_read_fif_split_file():
 
     bids_path.update(split='01')
     raw2 = read_raw_bids(bids_path=bids_path)
+
     bids_path.update(split='02')
     raw3 = read_raw_bids(bids_path=bids_path)
     assert len(raw) == len(raw1)
     assert len(raw) == len(raw2)
     assert len(raw) > len(raw3)
+
+    # check that the scans list contains two scans
+    scans_tsv = BIDSPath(
+        subject=subject_id, session=session_id,
+        suffix='scans', extension='.tsv',
+        root=bids_root)
+    scan_data = _from_tsv(scans_tsv)
+
+    scan_fnames = scan_data['filename']
+    scan_acqtime = scan_data['acq_time']
+    assert len(scan_fnames) == 2
+    assert 'split-01' in scan_fnames[0] and 'split-02' in scan_fnames[1]
+    # check that the acq_times in scans.tsv are the same
+    assert scan_acqtime[0] == scan_acqtime[1]
+    # check the recordings are in the correct order
+    assert raw2.first_time < raw3.first_time
+
+    # find sidecar scans.tsv file and alter the
+    # acquisition time to not have the optional microseconds
+    acq_time_str = scan_acqtime[0]
+    acq_time = datetime.strptime(acq_time_str, '%Y-%m-%dT%H:%M:%S.%fZ')
+    acq_time = acq_time.replace(tzinfo=timezone.utc)
+    new_acq_time = acq_time_str.split('.')[0]
+    assert acq_time == raw1.info['meas_date']
+    scan_data['acq_time'][0] = new_acq_time
+    _to_tsv(scan_data, scans_tsv)
+
+    """
+    # now re-load the data and it should be different
+    # from the original date and the same as the newly altered date
+    # but as read.py throws a ValueError when acq_time won't match,
+    # the ValueError need to be dealt with here.
+    bids_path.update(split='01')
+    raw4 = read_raw_bids(bids_path)
+    new_acq_time += '.0Z'
+    new_acq_time = datetime.strptime(new_acq_time,
+                                     '%Y-%m-%dT%H:%M:%S.%fZ')
+    new_acq_time = new_acq_time.replace(tzinfo=timezone.utc)
+    assert scan_acqtime[0] != scan_acqtime[1]
+    assert raw4.info['meas_date'] == new_acq_time
+    assert new_acq_time != raw1.info['meas_date']
+    """
 
 
 @pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -769,6 +769,7 @@ def test_bads_reading():
 @pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
 def test_write_read_fif_split_file():
     """Test split files are read correctly."""
+    # load raw test file, extend it to be larger than 2gb, and save it
     bids_root = _TempDir()
     tmp_dir = _TempDir()
     bids_path = _bids_path.copy().update(root=bids_root, datatype='meg')
@@ -782,27 +783,26 @@ def test_write_read_fif_split_file():
     raw = _read_raw_fif(big_fif_fname, verbose=False)
     write_raw_bids(raw, bids_path, verbose=False)
 
+    # test whether split raw files were read correctly
     raw1 = read_raw_bids(bids_path=bids_path)
     assert 'split-01' in str(bids_path.fpath)
-
     bids_path.update(split='01')
     raw2 = read_raw_bids(bids_path=bids_path)
-
     bids_path.update(split='02')
     raw3 = read_raw_bids(bids_path=bids_path)
     assert len(raw) == len(raw1)
     assert len(raw) == len(raw2)
     assert len(raw) > len(raw3)
 
-    # check that the scans list contains two scans
+    # check that split files both appear in scans.tsv
     scans_tsv = BIDSPath(
         subject=subject_id, session=session_id,
         suffix='scans', extension='.tsv',
         root=bids_root)
     scan_data = _from_tsv(scans_tsv)
-
     scan_fnames = scan_data['filename']
     scan_acqtime = scan_data['acq_time']
+
     assert len(scan_fnames) == 2
     assert 'split-01' in scan_fnames[0] and 'split-02' in scan_fnames[1]
     # check that the acq_times in scans.tsv are the same
@@ -810,31 +810,16 @@ def test_write_read_fif_split_file():
     # check the recordings are in the correct order
     assert raw2.first_time < raw3.first_time
 
-    # find sidecar scans.tsv file and alter the
-    # acquisition time to not have the optional microseconds
-    acq_time_str = scan_acqtime[0]
-    acq_time = datetime.strptime(acq_time_str, '%Y-%m-%dT%H:%M:%S.%fZ')
-    acq_time = acq_time.replace(tzinfo=timezone.utc)
-    new_acq_time = acq_time_str.split('.')[0]
-    assert acq_time == raw1.info['meas_date']
-    scan_data['acq_time'][0] = new_acq_time
+    # check whether non-matching acq_times are caught
+    scan_data['acq_time'][0] = scan_acqtime[0].split('.')[0]
     _to_tsv(scan_data, scans_tsv)
+    with pytest.raises(ValueError,
+                       match='Split files must have the same acq_time.'):
+        read_raw_bids(bids_path)
 
-    """
-    # now re-load the data and it should be different
-    # from the original date and the same as the newly altered date
-    # but as read.py throws a ValueError when acq_time won't match,
-    # the ValueError need to be dealt with here.
-    bids_path.update(split='01')
-    raw4 = read_raw_bids(bids_path)
-    new_acq_time += '.0Z'
-    new_acq_time = datetime.strptime(new_acq_time,
-                                     '%Y-%m-%dT%H:%M:%S.%fZ')
-    new_acq_time = new_acq_time.replace(tzinfo=timezone.utc)
-    assert scan_acqtime[0] != scan_acqtime[1]
-    assert raw4.info['meas_date'] == new_acq_time
-    assert new_acq_time != raw1.info['meas_date']
-    """
+    # reset scans.tsv file for downstream tests
+    scan_data['acq_time'][0] = scan_data['acq_time'][1]
+    _to_tsv(scan_data, scans_tsv)
 
 
 @pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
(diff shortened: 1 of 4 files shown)


sappelhoff:
Thanks @eort I'll hopefully be able to take a look later (also at the bids-specification PR). Just a small comment: What platform are you on (Windows, Linux, Mac), and are you working from the command line or via some graphical user interface?

It'd be nice if you could configure your local installation of `git` to know your identity, so that we later (in the codebase) can know who did what.

This should be a general intro: https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup

But if you work on Linux or mac, you can also check for a hidden file `.gitconfig` (note the dot) in your home directory, and edit that so that it contains your identity. For me, the beginning looks like this:

```
[user]
        email = stefan.appelhoff@mailbox.org
        name = Stefan Appelhoff
```

And the `email` there is the one I registered on GitHub as well.

eort:
> What platform are you on (Windows, Linux, Mac), and are you working from the command line or via some graphical user interface?

I'm on Ubuntu 18.04 and work with the command line. Sure, I'll set my .git_config file to include name and email!

sappelhoff:
@eort The continuous integration service (CI) has detected some style issues with your PR, see: https://github.com/mne-tools/mne-bids/pull/710/checks?check_run_id=1904887645

Could you fix these, please?



Screenshot: 

image



Furthermore, could you add tests in `test_read` and `test_write` so what we can be sure your fix works as expected and that future changes don't change the behavior again?

See:

https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L1

and

https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_write.py#L2

sappelhoff (commented):
> Aside of that, I couldn't run all tests, because there were missing files for some tests in test_write.py:

Have you taken a look at the developer installation documentation --> https://github.com/mne-tools/mne-bids/blob/master/CONTRIBUTING.md#running-tests ?

> That is not ideal, because if for some reason someone wants to combine two files that won't exceed the single-file-size limit, the method would split write split files into scans.tsv. I don't think this scenario is very likely though.

Perhaps @hoechenberger has an idea about this

eort:
> Have you taken a look at the developer installation documentation --> https://github.com/mne-tools/mne-bids/blob/master/CONTRIBUTING.md#running-tests ?

Yes, I did, and I think I have working developer environment. The problem started though, when I was trying to invoke `pytest` the first time, and couldn't download the test data, because it was trying to install it into my home directory under `mne_data`. Once I had created this folder manually, the data could be fetched, and some tests were run. However, as it already failed on `test_cli.py` and `test_copyfiles`, I had to invoke the tests in read.py and write.py manually.

eort:
Like so? I have never written a test, so not sure whether that commit made sense. 

wrt a test for `read.py`, I am not sure whether a new test is necessary. The idea would be to test, whether `read_raw_bids` correctly extracts the `fif` files from the `scans.tsv`, right? But isn't this already tested in [test_handle_scans_reading](https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L316)?

sappelhoff:
> But isn't this already tested in test_handle_scans_reading?

but does that test cover reading a scans.tsv with several split files? It doesn't look like it to me at a first glance, perhaps you could extend the test for that case.

adam2392:
I would look at these lines in the `read_raw_bids` function: https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/read.py#L545-L547

For the unit tests, I think in the spirit of saving time, perhaps appending test cases to this function is better?
https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L770
That way, there isn't two different tests doing a very long operation of (writing gigantic 2GB+ files), which has been shown to be the bottleneck in our testing suite (https://github.com/mne-tools/mne-bids/issues/678).

#####Some possible ideas:

Perhaps if the files are "split" files, then do some checks (e.g. all the acq times should be the exact same), and also make sure they are in "order"? 

To test this, you can then modify https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L770 to test that the error is raised. Perhaps, after the write modify the `*scans.tsv` file to do some funky business w/ the `acq_time`.

eort:
I bumped into a problem with the previous commit. The procedure failed if the file size was just below 2gb (but more than 1.9gb), so the approach to check `raw.filenames` as proxy whether a raw file would be split or not doesn't work [see here](https://github.com/mne-tools/mne-python/issues/8885#issuecomment-780705320). Instead, I check whether the file name that would be written to `scans.tsv` was actually written in the corresponding directory. If not, the file was likely split. Not sure how robust that is, though. It might be a problem if a file is split into more than two pieces. I'm not yet happy with this approach either.

> For the unit tests, I think in the spirit of saving time, perhaps appending test cases to this function is better?

Yes, I moved the test from `test_write.py` to `test_read.py` and integrated with

> https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L770

Like that, the big files are only written in one test. Therefore all the scans.tsv testing occurs in that test as well. 

I added tests for: 

- both split files are listed
- they have the same `acq_time`
- their start times are in the correct order


>  Perhaps, after the write modify the *scans.tsv file to do some funky business w/ the acq_time.

Could you elaborate on that funky business ? :) I added tests that are done in https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/tests/test_read.py#L316 but it is not quite clear to me, what is being tested and in which they would need to be updated in order to test split files. Some pointers are appreciated!

@adam2392 if I get you right, you mean that in https://github.com/mne-tools/mne-bids/blob/74ac41a3f2035e4469f9930e725fe36f53f07146/mne_bids/read.py#L190 there should be a check whether the acq_times are identical? I added some code to do that. Is this what you had in mind? 

Obviously, this is work in progress, so let me know whether there is a better way of getting your input, other than pushing more commits.

mne_bids/tests/test_read.py
```diff
@@ -787,12 +787,55 @@ def test_write_read_fif_split_file():
...
+
+    scan_fnames = scan_data['filename']
+    scan_acqtime = scan_data['acq_time']
+    assert len(scan_fnames) == 2
+    assert 'split-01' in scan_fnames[0] and 'split-02' in scan_fnames[1]
+    # check that the acq_times in scans.tsv are the same
+    assert scan_acqtime[0] == scan_acqtime[1]
+    # check the recordings are in the correct order
+    assert raw2.first_time < raw3.first_time
+
```
adam2392, line 812: Now, the scans.tsv have the correct `acq_time` metadata, which is tested. You could then say manually alter the scans.tsv file to not have the same acq_time for one of the split files:

E.g.
```Python
buff_scan_data = scan_data.copy()  # create a copy, so that way you can reset afterwards
scan_data['acq_time'][<one of the split indices>] = <some other datetime>
_to_tsv(scan_data, scans_tsv)
with pytest.raises(ValueError, match='split files must have the same acq_time'):
    read_raw_bids(bids_path)
# reset scans.tsv file for downstream tests
_to_tsv(buff_scan_data, scans_tsv)
```

adam2392:
> Could you elaborate on that funky business ? :) I added tests that are done in
> there should be a check whether the acq_times are identical? I added some code to do that. Is this what you had in mind?

Mostly LGTM! Sorry I should've elaborated :p. I left a comment here to describe how you might do "funky business" to achieve test coverage of the ValueError.

https://github.com/mne-tools/mne-bids/pull/710/files#r578621486

mne_bids/tests/test_read.py
eort: Thanks, that helped a lot!

eort:
I slightly changed the approach. Using the `fname` argument I filter the files in the directory the meg files should be written to. Based on the files in that directory, I extract the filenames that are written to scans.tsv

adam2392 (approved):

agramfort (commented):
@hoechenberger @sappelhoff I let you merge this if happy

mne_bids/write.py
```diff
@@ -413,8 +413,26 @@ def _scans_tsv(raw, raw_fname, fname, overwrite=False, verbose=True):
...
+        # to scans.tsv
+        datatype, basename = raw_fname.split(os.sep)
+        raw_dir = op.join(op.dirname(fname), datatype)
+        raw_files = [f for f in os.listdir(raw_dir) if f.endswith('.fif')]
+        if basename not in raw_files:
+            raw_fnames = []
+            for raw_f in raw_files:
+                if basename.split('_meg.fif')[0] in raw_f:
+                    raw_fnames.append(op.join(datatype, raw_f))
+            raw_fnames.sort()
```
sappelhoff, line 430: I think this is slightly un-robust.

1. in line 425 you check that the `basename` (e.g., `sub-01_task-test_meg.fif`) is in the list of files.
1. if it is ... all is well, it was not split (no `_split-<idx>` was appended), so we can proceed
1. if it is NOT (the interesting part), you proceed with more checks

See this line: `if basename.split('_meg.fif')[0] in raw_f`

The intention is to use e.g. `sub-01_task-test_meg.fif` to catch files like `sub-01_task-test_split-01_meg.fif`, and `sub-01_task-test_split-02_meg.fif` ... and then put these into a list, order them, and write.

But wouldn't the current code also catch something like `sub-01_task-test_acq-MyParam_meg.fif`?

sappelhoff (commented):
It looks good to me overall, but I have one comment where I am not entirely sure that the code is robust.

mne_bids/write.py
eort: Okay, maybe I don't understand what forms `raw_fname` can take then. I assumed that all the BIDS fields that are part of that recording (sub, task and acq in your example) would be part of `raw_fname` (the only thing missing being the split parameter). So that, the filename on disk would be the same as `raw_fname`, independent of which BIDS fields are included. But you are suggesting that `raw_fname` will not necessarily include all fields (e.g. `acq`)?

eort: in other words, I would think line 425 only returns `False` if the files was split and `True` otherwise.

sappelhoff: > But you are suggesting that raw_fname will not necessarily include all fields (e.g. acq)?

I may be wrong, but couldn't you first write a data file to BIDS without an `acq` entity, ... and then a second one WITH that `acq` entity?

eort: Ah, I see. In case there are raw files in the directory that have different number of field, but are the same otherwise? Lines 427-430 would then write twice the file name without `acq` to `scans.tsv`, once because it is found in the file name without `acq` (as it should) and once in the filename with `acq` (because line 428 is too general). Yeah, that's a problem. So better take the basename, add the split field and then check whether there is an exact match.

sappelhoff: > So better take the basename, add the split field and then check whether there is an exact match.

yes, I think that'd be best :+1:

eort: looks a bit hacky now :)

I tested it with some fif files I had available. I suppose these changes should also be covered by the unit test? Like writing one (or more) extra files to that directory and see whether, the correct files will be listed in scans.tsv. Or would that be overkill?

sappelhoff: > looks a bit hacky now :)

but probably works better! :)

But one more comment: you should use the `.split()` method on `"split"`, but on `"_split-"`, because the former may also catch something like `sub-01_task-Mindsplitting_meg.fif` ("Mindsplitting is just an example task here).

`_split-` is guaranteed to match only the split entity.

>  I suppose these changes should also be covered by the unit test? Like writing one (or more) extra files to that directory and see whether, the correct files will be listed in scans.tsv. Or would that be overkill?

If you can manage to cover this case in a couple of lines, that'd be awesome

sappelhoff (approved):
great, thanks a lot @eort.

@hoechenberger can you please make a final check and merge if you are happy?

sappelhoff:
seems like Richard us busy these days, so I'll merge. Thanks a lot @eort for your first PR to mne-bids! :tada:

hoechenberger:
@sappelhoff I think I'd accidentally disabled notifications for this PR, whoops! Thanks for taking care of this! All good with me.

PR #730, opened 2021-03-16
[MRG] MAINT: migrate setup.py to setup.cfg

sappelhoff:
using just setup.cfg as proposed in this PR should be equivalent to our previous setup.py+setup.cfg

It seems like many people have moved towards using the cfg file (which is just parsed) to declare their data instead of the py file (which needs to be executed).

I added:
- author = mne-bids developers
- keywords = meg, eeg, ieeg, bids, brain imaging data structure, neuroscience, neuroimaging

I removed
- download_url (because it was just pointing to GitHub, not to a download page)

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [ ] All CIs are happy
- ~[whats_new.rst](https://github.com/mne-tools/mne-bids/blob/master/doc/whats_new.rst) is updated~
- ~PR description includes phrase "closes <#issue-number>"~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2.1
 jobs:
   docs-build:
     docker:
-      - image: circleci/python:3.8.1-buster
+      - image: cimg/python:3.8
     steps:
         - checkout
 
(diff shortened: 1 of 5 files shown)


agramfort (approved):
diff looks ok but I did not test :)

I would test to make an sdist to check it's ok

sappelhoff:
> I would test to make an sdist to check it's ok

Thanks to the test that @hoechenberger added, this is done in CI :-)

But I also added a quick test to also check that the CLI is still working (it does)

agramfort:
@hoechenberger merge if happy

hoechenberger (approved):

hoechenberger:
Thanks @sappelhoff!

PR #750, opened 2021-03-24
ENH: Switch to pydata-sphinx-theme

### LINKED ISSUE #749: STY: Move to pydata-sphinx-theme
Now that NumPy, SciPy (in progress), pandas, and (most importantly I think) MNE-Python have moved to pydata-sphinx-theme, it might be nice to move `mne-bids` over, too, for cohesiveness. Any thoughts or objections?

larsoner:
Closes #749.

This produces something locally with no warnings, haven't checked to see if much of anything works yet, though.

Also bumps the version to 0.8.dev0 which seems like it should have been done when 0.7 stable was released.


larsoner:
v1: https://4371-89170358-gh.circle-artifacts.com/0/dev/index.html

larsoner:
I removed the `main.yml` as part of this PR.

Actually @sappelhoff I think this might be good enough to merge. Check it out: https://4374-89170358-gh.circle-artifacts.com/0/dev/index.html

Tweaks could be made in subsequent PRs, but at least the top and nav dropdown seem to work etc.

sappelhoff (commented):
The examples that don't have a thumbnail look a bit off :thinking: Any idea how to fix that?
https://4374-89170358-gh.circle-artifacts.com/0/dev/use.html

image

sappelhoff (commented):
and for the CLI docs it seems like sometimes the "content" window is not wide enough and line breaks happen at weird positions: 
https://4374-89170358-gh.circle-artifacts.com/0/dev/generated/cli.html

image


and finally I am not sure whether or not I lke the "footer navigation" (but that's not a bug, just a question of design/taste):

image

larsoner:
> The examples that don't have a thumbnail look a bit off ￼ Any idea how to fix that?

Upgrade your browser and/or machine :) But really I don't see it in chrome on Linux:

Screenshot from 2021-03-24 15-29-09

Do you see it on the MNE-Python website?

https://mne.tools/dev/auto_tutorials/index.html#reading-data-for-different-recording-systems

> or the CLI docs it seems like sometimes the "content" window is not wide enough and line breaks happen at weird positions:

I can get something like this only when I go to a very small window, which also occurs with the current version of the docs. But it's easy enough to fix via CSS so I've pushed a commit to fix it anyway.

> and finally I am not sure whether or not I lke the "footer navigation" (but that's not a bug, just a question of design/taste):

I ported over the same CSS fix that we use in MNE:

Screenshot from 2021-03-24 15-44-42

sappelhoff:
> Upgrade your browser and/or machine :) But really I don't see it in chrome on Linux:

I use Ubuntu 18.04.5 with most recent firefox, so there *is* a newer version of Ubuntu, but 18.04 is still LTS - so I feel like that shouldn't be an issue.

Switching to recent Chrome does make the issue go away though. :+1: 

I see an issue, yes. For example:

image


see the rightmost example thumbnail that hides something underneath.

Curiously that issue **is** also present on chrome.

There is another one here, where there is not enough space and some writing is covered by an image:

image

(that one is again **not** present on chrome)

But overall, there are no examples **without** thumbnail, as is the case in mne-bids. (because there is always *at least* the mne default thumbnail)


---

> I ported over the same CSS fix that we use in MNE:

yes, regarding the footers I'll let Richard and Mainak say what they like better ... I assume these footers are easy to get rid of if more people than just me don't like them too much.

jasmainak (approved):
No objection, looks good.

I tried the binder links while I was at it. Impressed that it works out of the box :)

larsoner:
For the example spacing stuff, let's fix it in MNE-Python (or sphinx gallery or pydata-sphinx-theme, but feel free to add it to @drammock's open issue in MNE-Python about needed tweaks) then port here. This PR basically gets both to repos to the same state and fixes can go both places when they are made, so I'd rather get this in and continue in incremental PRs both places.

drammock:
the spacing issues mentioned are both known to me, and at least one of them (text covered by icon below it) is already in our issues list.  The other one is due to the icon for the MNE-Reports tutorial being really small and its container no longer having a min-width probably.  I was hoping to get to them this week (for MNE-Python).  I plan to patch our CSS first and then see about upstreaming the fix.

drammock:
fixes for those two problems are now in https://github.com/mne-tools/mne-python/pull/9191

hoechenberger:
Will have a look tomorrow! :)

larsoner:
Ported changes from mne-tools/mne-python#9191

sappelhoff (approved):
awesome, thanks a lot.

Please merge when you are happy @hoechenberger

hoechenberger:
Thanks a bunch, @larsoner! Great work!

PR #751, opened 2021-03-24
ENH: Add redirector

larsoner:


larsoner:
Can only really work once it's in `main` so let's merge it and see if it works in PRs....

sappelhoff:
@larsoner is this maybe a duplicate of https://github.com/mne-tools/mne-bids/blob/0bbd41433d56e3765fe0290b2fd94c6ddc71eaf1/.github/workflows/main.yml#L1

?

larsoner:
Ohno! Yes it is. I'll revert mine then

larsoner:
... actually I think I'd rather remove the `main.yml`. I think the `circleci_redirector.yml` is a better name actually.

sappelhoff:
> ... actually I think I'd rather remove the main.yml. I think the circleci_redirector.yml is a better name actually.

agreed @larsoner

PR #754, opened 2021-03-26
MRG, FIX: Retain essential annotations created by acquisition system

hoechenberger:
MNE exposes certain markers added by the acquisition system as Annotations,
for example `BAD_ACQ_SKIP` but possibly others. Typically, these markers are
not transferred to `events.tsv` by BIDS users.

Now, when reading raw files, we moved to superseding Annotations in raw data
with information from `events.tsv`: we first read the sidecar, create Annotations
from those events, and call `raw.set_annotations()`. However, this removes
all existing Annotations – including the special ones like `BAD_ACQ_SKIP`,
leading to problems in downstream processing.

This PR ensures that `BAD_ACQ_SKIP` is retained; the list of Annotations
to retain can be easily expanded.

(The problem does not seem to exist when writing raw data; it only occured
when reading)

cc @agramfort 

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -21,7 +21,9 @@ from mne.transforms import apply_trans
 
 from mne_bids.dig import _read_dig_bids
 from mne_bids.tsv_handler import _from_tsv, _drop
-from mne_bids.config import ALLOWED_DATATYPE_EXTENSIONS, reader, _map_options
+from mne_bids.config import (ALLOWED_DATATYPE_EXTENSIONS,
+                             ANNOTATIONS_TO_KEEP,
+                             reader, _map_options)
 from mne_bids.utils import _extract_landmarks, _get_ch_type_mapping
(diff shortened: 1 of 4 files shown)


mne_bids/read.py
```diff
@@ -339,12 +341,19 @@ def _handle_events_reading(events_fname, raw):
     descriptions = descriptions[good_events_idx]
     del good_events_idx
 
-    # Add Events to raw as annotations
+    # Add events as Annotations, but keep essential Annotations present in
+    # raw file
+    # Ensusure Annotations share the same orig_time
```
agramfort, line 346:
```suggestion
    # Ensure Annotations share the same orig_time
```

agramfort (approved):
diff LGTM

thx @hoechenberger 

let's just make CIs green ...

PR #756, opened 2021-03-26
[BUG] Fix matplotlib for version 3.4

alexrockhill:
PR Description
--------------

Fix failing tests due to matplotlib upgrade.

Merge checklist
---------------

Maintainer, please confirm the following before merging:

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [x] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] PR description includes phrase "closes <#issue-number>"

PR #757, opened 2021-03-28
[BUG] Fix for mpl 3.4 requirements

alexrockhill:
Fix for previous problem https://github.com/mne-tools/mne-bids/pull/756

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_inspect.py
+++ b/mne_bids/tests/test_inspect.py
@@ -28,9 +28,13 @@ _bids_path = BIDSPath(subject='01', session='01', run='01', task='testing',
 
 def _close_non_interactive_fig(fig):
     """Close figure and children figures by explicit function call."""
+    # because the figure is not interactive, these close calls are not
+    # fired by `plt.close`, so we have to call them explicitly, see
+    # https://github.com/matplotlib/matplotlib/issues/18609
     for child_fig in fig.mne.child_figs:
         child_fig._close(event=None)
     fig._close(event=None)
+    fig.canvas.key_press_event(fig.mne.close_key)
 
 
 def setup_bids_test_dir(bids_root):
--- a/mne_bids/tests/test_inspect.py
+++ b/mne_bids/tests/test_inspect.py
@@ -26,6 +26,13 @@ _bids_path = BIDSPath(subject='01', session='01', run='01', task='testing',
                       datatype='meg')
 
 
+def _close_non_interactive_fig(fig):
+    """Close figure and children figures by explicit function call."""
+    for child_fig in fig.mne.child_figs:
+        child_fig._close(event=None)
+    fig._close(event=None)
+
+
 def setup_bids_test_dir(bids_root):
     """Return path to a written test BIDS dir."""
     data_path = testing.data_path()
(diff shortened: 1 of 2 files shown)


hoechenberger (commented):
I'm not sure I understand what's happening here... the idea of this test was to simulate user interaction, which includes the key press event. Why is this not working in 3.4 anymore..?

alexrockhill:
> I'm not sure I understand what's happening here... the idea of this test was to simulate user interaction, which includes the key press event. Why is this not working in 3.4 anymore..?

See the discussion in this issue about close events with a non-interactive backend not firing https://github.com/matplotlib/matplotlib/issues/18609. So you are simulating a click roughly, but for `Agg` you just have to skip to simulating the emitted event.

alexrockhill:
Also, I'm really not sure why it worked in previous versions of mpl, I looked into it a bit and I couldn't tell you

hoechenberger:
Should we maybe use an interactive backend then? And test in a virtual X server?

alexrockhill:
> Should we maybe use an interactive backend then? And test in a virtual X server?

It sounds like that could slow down the tests but maybe that would be good, not sure

agramfort:
Are we sure it’s not treated as a bug by mpl so it’s fixed in one Minor release ?

hoechenberger:
> Are we sure it’s not treated as a bug by mpl so it’s fixed in one Minor release ?

In the linked issue they propose how to fix this, but it's been stale since…

alexrockhill:
> Are we sure it’s not treated as a bug by mpl so it’s fixed in one Minor release ?

It looks like they labeled the issue for discussion and they're not sure if they do want a non-interactive figure to fire a close event because they don't think it makes sense for a non-interactive figure to close. So it might happen but I would guess not in the next version release. It's not a new issue though so, again, I'm really not sure why the tests started failing upon the 3.4.0 release and why adding this `_close` call fixed them, a bit of a mystery...

agramfort:
Ok maybe let’s be pragmatic here and merge ... your call

hoechenberger (approved):

hoechenberger:
Yes let's merge once we're green

hoechenberger:
@alexrockhill The tests are stalled :(

hoechenberger:
I reverted your last commit, thinking this might fix CI, but I'm not sure if it's actually related…

alexrockhill:
> I reverted your last commit, thinking this might fix CI, but I'm not sure if it's actually related…

It has been stalling for the last couple and for a PR on mne-python related to this issue... I think it's a Microsoft server issue

hoechenberger:
Ok … shall I revert my reversal than? ;)

alexrockhill:
Sure, it the last commit doesn't really change anything but it does have the simulated click

alexrockhill:
So I looked a bit more and it looks like this loop in mne.viz._figure.py (line 557) is doing an infinite loop

```
# Clean up child figures (don't pop(), child figs remove themselves)
while len(self.mne.child_figs):
    fig = self.mne.child_figs[-1]
    close(fig)
```

It appears they don't remove themselves any longer maybe

alexrockhill:
Maybe there is also something going on with the test servers, as they are really slow but the tests pass locally so I think this is an okay solution for testing, it would be great if mpl could sort this out so that the tests didn't have to do these weird workarounds

hoechenberger:
yeah let's look at this again tomorrow

alexrockhill:
Could someone with repository ownership/collaborator cancel these workflows? They are on an infinite loop.

PR #759, opened 2021-03-28
[BUG] Fix matplotlib version issues for 3.4.0 release

alexrockhill:
Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_inspect.py
+++ b/mne_bids/tests/test_inspect.py
@@ -26,6 +26,17 @@ _bids_path = BIDSPath(subject='01', session='01', run='01', task='testing',
                       datatype='meg')
 
 
+def _close_non_interactive_fig(fig):
+    """Close figure and children figures by explicit function call."""
+    # because the figure is not interactive, these close calls are not
+    # fired by `plt.close`, so we have to call them explicitly, see
+    # https://github.com/matplotlib/matplotlib/issues/18609
+    for child_fig in fig.mne.child_figs.copy():
+        child_fig._close(event=None)
+    fig._close(event=None)
+    fig.canvas.key_press_event(fig.mne.close_key)
+
+
 def setup_bids_test_dir(bids_root):
     """Return path to a written test BIDS dir."""
     data_path = testing.data_path()
@@ -81,7 +92,7 @@ def test_inspect_single_file(tmp_path, save_changes):
     _click_ch_name(raw_fig, ch_index=4, button=1)
 
     # Closing the window should open a dialog box.
-    raw_fig.canvas.key_press_event(raw_fig.mne.close_key)
+    _close_non_interactive_fig(raw_fig)
     fig_dialog = mne_bids.inspect._global_vars['dialog_fig']
 
     if save_changes:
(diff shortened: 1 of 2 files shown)


hoechenberger:
Same CI problem, same issue with mne-python too

hoechenberger:
ok seems to be back to normal now. canceled like 3 dozen pending workflows.

hoechenberger:
It stalled again, so I canceled it :S

hoechenberger:
Let's try this again once #760 has been merged

alexrockhill:
Thanks @hoechenberger, sorry I made such a mess of this, I was trying to fix it :(

hoechenberger:
It's not your fault, we misconfigured GH Actions 😊

hoechenberger:
It appears this is stalled again, @alexrockhill … maybe @drammock has an idea on what's going on?

alexrockhill:
> It appears this is stalled again, @alexrockhill … maybe @drammock has an idea on what's going on?

Huh, I just ran the tests locally and they didn't hang up...

hoechenberger:
At least the auto-cancellation after 60 min, which I added in #760, seems to be doing its job :)

alexrockhill:
Oh maybe it's still hanging up, I have a meeting but I'll fix it after. I just did all the tests and it hung up...

alexrockhill:
Gosh that was a total rookie move, I was modifying the object I was looping over and didn't copy so that it didn't complete the loop and remove all the child figures and then that caused the parent to hang when it tried to remove all the child figures with the non-functioning `close(fig)` method.

alexrockhill:
Huh, all those tests pass locally, not sure what's wrong now

drammock:
@alexrockhill I can get the MNE-Python tests to fail locally if I set `MNE_BROWSE_RAW_SIZE` to `'6.4,4.8'` which is the default size when it's never been changed by the user.  I've isolated which MPL commit first fails, and am working on figuring out why exactly that particular change causes the fail.

drammock:
hopefully https://github.com/mne-tools/mne-python/pull/9217 will work 🤞🏻

alexrockhill:
Huh, it worked with mne-python version 0.22 locally for this test, not sure if that has to do with the PR but hopefully that fixes things

alexrockhill:
If the double click was the issue then the main tests should pass and the stable should fail ASAIK

alexrockhill:
Well I am just perplexed, `ubuntu-latest` passes but everything else fails that just doesn't make sense

alexrockhill:
> Well I am just perplexed, `ubuntu-latest` passes but everything else fails that just doesn't make sense

Oh wait looking at the workflow that's just because `ubuntu-latest` is the only one that clones the `main` branch of `mne-python` everything else clones the `maint/0.22`. So solved, thanks @drammock!

alexrockhill:
So then the `mne-bids` dependency should be left as mpl<3.4 until the next `mne-python` release so we should wait until then to merge this PR I guess

alexrockhill:
So now that the new stable version of mne is released, I think this should be good to merge after retriggering tests

hoechenberger:
Can you pull in `main` to see if this fixes the tests?

alexrockhill:
> Can you pull in `main` to see if this fixes the tests?

Hmmm, I rebased but this is using the latest stable version of mne-python (0.22.1) so I thought this would work... I'll have to think about what the issue is.

alexrockhill:
I guess since `main` works we could just wait until the next full version is released...

sappelhoff:
@alexrockhill @hoechenberger was this issue resolved somewhere upstream and this PR is now obsolete? CIs seem to pass on main without the tweaks here.

PR #762, opened 2021-03-30
fix orig_times

agramfort:
cc @hoechenberger 

this https://github.com/mne-tools/mne-bids/commit/415988e88f6253527d38d60dd521d0b94d5baa98

created an offset in the annotations

let's see what CIs have to say...


hoechenberger:
Ok this is weird, originally I used the solution you're proposing here, but got a bunch of tests failed, complaining about excluded / dropped annotations :(

hoechenberger (approved):

Issue #764, opened 2021-03-30
write_raw_bids shouldn't be overwriting data in dataset_description.json unless it doesn't exist yet


adam2392:
Rn I have modified my `dataset_description.json` file as:

```
"Name": "Sick Kids Pre and Post Resection Data",
    "BIDSVersion": "1.5.0",
    "DatasetType": "raw",
    "Authors": [
        "Adam Li",
        "Nebras Warsi",
        "George Ibrahim"
    ],
    "Funding": [
    	"NSF-GRFP",
    	"ARCS Scholarship",
    	"Whitaker Fellowship",
    	"Chateaubriand Fellowship",
    ],
    "EthicsApprovals": [
    	"The Hospital for Sick Children IRB",
    ],
    "HowToAcknowledge": "Cite our paper at ...",
    "ReferencesAndLinks": [
    	"https://pubmed.ncbi.nlm.nih.gov/33555026/"
  	],
```

However, I ran write_raw_bids to add more data and I got a new file:

```
{
    "Name": " ",
    "BIDSVersion": "1.4.0",
    "DatasetType": "raw",
    "Authors": [
        "Please cite MNE-BIDS in your publication before removing this (citations in README)"
    ],
    "HowToAcknowledge": "Cite our paper at ...",
    "Funding": [
        "NSF-GRFP",
        "ARCS Scholarship",
        "Whitaker Fellowship",
        "Chateaubriand Fellowship"
    ],
    "ReferencesAndLinks": [
        "https://pubmed.ncbi.nlm.nih.gov/33555026/"
    ]
}
```

Ideally, this does not overwrite 'Authors', and 'Name' actually at all. I feel like `overwrite` in `write_raw_bids` for `make_dataset_description` should ALWAYS be False

Issue #763, opened 2021-03-30
Allow file extensions to be capitalized


adam2392:
#### Describe the bug
Reading in EDF files with extension `.edf` or `.EDF` is the same in mne-python. It's just a casing issue. However, if we have files with `.EDF`, it throws an error right now because of a check.

#### Expected results
~~Apply lower-case to all extensions found in filenames that aren't case-sensitive (e.g. Nihon-Kohden)?~~ Add `.EDF` to acceptable keywords in config.py.

sappelhoff:
+1, it seems that both `.edf` and `.EDF` are valid:

https://www.edfplus.info/specs/edfplus.html

> All EDF+ files must have .edf or .EDF as filename extension

agramfort:
+1

PR #765, opened 2021-03-30
[MRG] Fix dataset description overwriting in write_raw_bids and EDF extension casing.

adam2392:
PR Description
--------------

Closes: #763 
Closes: #764 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -939,7 +939,7 @@ def print_dir_tree(folder, max_depth=None, return_str=False):
         return outfile.getvalue()
 
 
-def _parse_ext(raw_fname, verbose=False):
+def _parse_ext(raw_fname, match_casing=True, verbose=False):
     """Split a filename into its name and extension."""
     raw_fname = str(raw_fname)
     fname, ext = os.path.splitext(raw_fname)
@@ -953,6 +953,13 @@ def _parse_ext(raw_fname, verbose=False):
     elif ext == '.gz' and raw_fname.endswith('.nii.gz'):
         ext = '.nii.gz'
         fname = fname[:-4]  # cut off the .nii
+
+    if not match_casing:
+        # for Nihon Kohden systems with '.EEG' extension, this is
+        # case-sensitive. However, EDF can be `.edf` and `.EDF`.
+        # We default to a lower-case
+        if ext != '.EEG':
+            ext = ext.lower()
     return fname, ext
 
 
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -46,7 +46,7 @@ from mne_bids.utils import (_stamp_to_dt, _get_anonymization_daysback,
                             get_anonymization_daysback, _write_json)
 from mne_bids.tsv_handler import _from_tsv, _to_tsv
 from mne_bids.sidecar_updates import _update_sidecar
-from mne_bids.path import _find_matching_sidecar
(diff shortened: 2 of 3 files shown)


hoechenberger:
Why limit ourselves to EDF files only, can we not do the same for all supported file types?

adam2392:
> Why limit ourselves to EDF files only, can we not do the same for all supported file types?

I don't think it's necessary for all file types. For example, NK is `.EEG`, whereas BrainVision is `.eeg`, so I'm wary of applying a casing rule to everyone. In the case of EDF, they explicitly say `.EDF` is allowed.

hoechenberger:
> For example, NK is `.EEG`, whereas BrainVision is `.eeg`

I'm actually not sure if this heuristic would even work on macOS. By default, the macOS file system is case-agnostic.

hoechenberger:
> By default, the macOS file system is case-agnostic.

```shell
(base) ➜  test ls -lah foo*
(base) ➜  test echo "foo" >foo.eeg
(base) ➜  test echo "bar" >foo.EEG
(base) ➜  test cat foo*
bar
(base) ➜  test ls -lah
total 8
drwxr-xr-x   3 hoechenberger  wheel    96B Mar 30 20:53 .
drwxrwxrwt  13 root           wheel   416B Mar 30 20:52 ..
-rw-r--r--   1 hoechenberger  wheel     4B Mar 30 20:53 foo.eeg
```

adam2392:
> > For example, NK is `.EEG`, whereas BrainVision is `.eeg`
> 
> I'm actually not sure if this heuristic would even work on macOS. By default, the macOS file system is case-agnostic.

The nice thing is for these other cases, `.eeg` for BV is the raw data, whereas in NK, `.EEG` is the header file I believe :), so some sort of runtimeerror will occur that doesn't at least bombard the user with some poorly overwritten data.

I think EDF is the only one to my knowledge that has this casing phenomena allowed tho...

adam2392:
> > By default, the macOS file system is case-agnostic.
> 
> ```shell
> $ echo "foo" >foo.eeg
> $ echo "bar" >foo.EEG
> $ cat foo*
> bar
> $ ls -lah
> total 8
> drwxr-xr-x   3 hoechenberger  wheel    96B Mar 30 20:53 .
> drwxrwxrwt  13 root           wheel   416B Mar 30 20:52 ..
> -rw-r--r--   1 hoechenberger  wheel     4B Mar 30 20:53 foo.eeg
> ```

The issue is that storage of "files" are case-agnostic in a hash-map sense (`blah.edf` and `blah.EDF` cannot both exist in the same directory at the same time).

But! if you use a Python engine/interpreter, and rename a `.edf` with `.EDF` and try the mne-bids, you'll get hit with an error because "reading the file name" does not give you the case-agnostic filename. It gives you whatever you see on your Desktop.

agramfort (approved):
LGTM

agramfort:
+1 for MRG provided you fix conflicts and make CI green

hoechenberger:
Very interesting, thanks for the clarification, @adam2392

adam2392:
Should be fixed + added some more info to the `Notes` section.

hoechenberger:
@adam2392 Unfortunately, CI is still failing

agramfort:
Thanks @adam2392

PR #771, opened 2021-04-03
ENH: Add t1_bids_path parameter to get_head_mri_trans()

hoechenberger:
The new parameter allows to generate the head <> MRI trans from recordings belonging to different sessions or even separate datasets.

Scenario:

- 1st session: MR
- all following sessions: MEG

In `main`, it wouldn't be possible to use `get_head_mri_trans()` to get the transformation matrix, as MNE-BIDS cannot find the required files.

**RFC**: Should the param maybe be called `t1w_bids_path` instead? (t1w instead of t1)

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -585,30 +585,38 @@ def read_raw_bids(bids_path, extra_params=None, verbose=True):
     return raw
 
 
-def get_head_mri_trans(bids_path, extra_params=None):
+def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None):
     """Produce transformation matrix from MEG and MRI landmark points.
 
     Will attempt to read the landmarks of Nasion, LPA, and RPA from the sidecar
-    files of (i) the MEG and (ii) the T1 weighted MRI data. The two sets of
+    files of (i) the MEG and (ii) the T1-weighted MRI data. The two sets of
     points will then be used to calculate a transformation matrix from head
     coordinates to MRI coordinates.
+
+    .. note:: The MEG and MRI data need **not** necessarily be stored in the
+              same session or even in the same BIDS dataset. See the
+              ``t1_bids_path`` parameter for details.
 
     Parameters
     ----------
     bids_path : mne_bids.BIDSPath
-        The path of the recording for which to retrieve the transformation. The
-        :class:`mne_bids.BIDSPath` instance passed here **must** have the
-        ``.root`` attribute set.
+        The path of the MEG recording.
     extra_params : None | dict
-        Extra parameters to be passed to MNE read_raw_* functions when reading
-        the lankmarks from the MEG file.
-        If a dict, for example: ``extra_params=dict(allow_maxshield=True)``.
+        Extra parameters to be passed to :func:`mne.io.read_raw` when reading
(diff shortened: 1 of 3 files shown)


mne_bids/read.py
```diff
@@ -585,30 +585,38 @@ def read_raw_bids(bids_path, extra_params=None, verbose=True):
...
-        the lankmarks from the MEG file.
-        If a dict, for example: ``extra_params=dict(allow_maxshield=True)``.
+        Extra parameters to be passed to :func:`mne.io.read_raw` when reading
+        the MEG file.
+    t1_bids_path : mne_bids.BIDSPath | None
+        You may optionally specify a :class:`mne_bids.BIDSPath` pointing to the
+        T1-weighted MRI scan that shall be used for extraction of MRI
+        landmarks. Use this parameter if the T1 scan was recorded during a
+        different session than the MEG. It is even possible to point to a
+        T1 image stored in a different **BIDS dataset** than the MEG data.
```
agramfort, line 612:
```suggestion
    t1_bids_path : mne_bids.BIDSPath | None
        If None, the T1-weighted MRI file will be found based on MEG bids_path.
        You may optionally specify a :class:`mne_bids.BIDSPath` pointing to the
        T1-weighted MRI scan that shall be used for extraction of MRI
        landmarks. Use this parameter if the T1 scan was recorded during a
        different session than the MEG. It is even possible to point to a
        T1 image stored in a different **BIDS dataset** than the MEG data.
```

mne_bids/read.py
```diff
@@ -619,22 +627,28 @@ def get_head_mri_trans(bids_path, extra_params=None):
...
+                                               extension='.json')
+    else:
+        t1_bids_path = t1_bids_path.copy().update(suffix='T1w',
+                                                  datatype='anat')
+        t1w_json_path = _find_matching_sidecar(t1_bids_path, suffix='T1w',
+                                               extension='.json')
 
     # Get MRI landmarks from the JSON sidecar
-    with open(t1w_json_path, 'r', encoding='utf-8-sig') as f:
+    with open(t1w_json_path, 'r', encoding='utf-8') as f:
```
agramfort, line 651: why did you change this?

agramfort (approved):
besides LGTM !

mne_bids/read.py
hoechenberger: Good catch, forgot to mention this!

So it SHOULD have been `utf-8` all along; the UTF BOM character which is handled by `utf-8-sig` is invalid in JSON. Meaning that we shouldn't even try to read it because … well, it shouldn't be present!

One could argue that this change could break backward-compat for some users; however their datasets wouldn't be BIDS-compatible anyway if their JSON files contain a BOM (as, like I said, the BOM is invalid in JSON)

In all other places of MNE-BIDS, we read & write JSON using `utf-8` these days, not `utf-8-sig`; I suppose this line here was just overlooked in #625

agramfort:
@sappelhoff i let you merge

sappelhoff (approved):
lgtm, thanks :)

PR #778, opened 2021-04-19
MRG, ENH: Add symlink support to write_raw_bids()

hoechenberger:
PR Description
--------------

This adds support to  `write_raw_bids()` for creating symbolic links  (instead of copying data) to preserve storage space, as requested by @SophieHerbst

Current limitations:

- Only for FIFF
- <strike>Not for split files</strike>
- Not on Windows

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1061,7 +1061,10 @@ def write_raw_bids(raw, bids_path, events_data=None,
         data (i.e., ``anonymize`` must be ``None``).
 
         .. note::
-           Symlinks currently only work with FIFF files.
+           Symlinks currently only work with FIFF files. In case of split
+           files, only a link to the first file will be created, and
+           :func:`mne_bids.read_raw_bids` will correctly handle reading the
+           data again.
 
         .. note::
            Symlinks are currently only supported on macOS and Linux. We will
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -36,8 +36,8 @@ from mne_bids.pick import coil_type
 from mne_bids.dig import _write_dig_bids, _write_coordsystem_json
 from mne_bids.utils import (_write_json, _write_tsv, _write_text,
                             _age_on_date, _infer_eeg_placement_scheme,
-                            _handle_datatype, _get_ch_type_mapping,
-                            _check_anonymize, _stamp_to_dt)
+                            _get_ch_type_mapping, _check_anonymize,
+                            _stamp_to_dt, _handle_datatype)
 from mne_bids import BIDSPath
 from mne_bids.path import _parse_ext, _mkdir_p, _path_to_str
 from mne_bids.copyfiles import (copyfile_brainvision, copyfile_eeglab,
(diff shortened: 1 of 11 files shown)


mne_bids/write.py
```diff
@@ -1327,15 +1353,23 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
         convert = ext not in ALLOWED_DATATYPE_EXTENSIONS[bids_path.datatype]
 
-    # check if there is an BIDS-unsupported MEG format
-    if bids_path.datatype == 'meg' and convert and not anonymize:
+        if convert and symlink:
+            raise RuntimeError(
+                'The input file format is not supported by the BIDS standard. '
+                'To store your data, MNE-BIDS would have to convert it. '
+                'However, this is not possible since you set symlink=True. '
+                'Deactiave symbolic links by passing symlink=False to allow '
```
agramfort, line 1361:
```suggestion
                'Deactivate symbolic links by passing symlink=False to allow '
```

agramfort (commented):
looks good !

mne_bids/write.py
SophieHerbst: Looks great @hoechenberger!
About the split files: is this something that could happen directly at storage by the acquisition software, or are those only created by mne functions? (Just wondering if we should instruct people to calculate file size beforehand to prevent this).

SophieHerbst:
Looks great @hoechenberger!
About the split files: is this something that could happen directly at storage by the acquisition software, or are those only created by mne functions? (Just wondering if we should instruct people to calculate file size beforehand to prevent this).

agramfort:
yes it can happen during acquisitions. See how much space a file takes at
NS for 5 mins of data and then conclude
given the fact that fif files cannot be more than 2GB.

>

SophieHerbst:
Thanks @agramfort! We are recording with different sampling rates, but will check file sizes. If I understand the code correctly, it would break if 1 file is split? Could we rather have it issue a warning and write the split files as a physical file then @hoechenberger ?
This would still save us a lot of storage space, compared to writing the whole set of files (usually ~30 x 8) without symlinks.

agramfort:
yes I thought the same. It could be

symlink = True | False | "try"
or
symlink = True | False | "if_possible"

hoechenberger:
Thinking about it, I suppose we could resolve the symlink before calling `mne.io.read_raw_fif()`, and then MNE would automatically find the remaining split files in the correct place.

What I'm saying is:
We can probably totally ignore the fact that a file was split when creating the symlink: we only create ONE symlink to the FIRST file. Upon reading, we don't even tell MNE about the fact that we're using a symlink; we resolve it to the first file, and let MNE do everything the way it always does from there.

agramfort:
it means read_raw_fif then needs to resolve the symlink on read to find
it's next file
as the next file name is written in the first one.

>

hoechenberger:
> it means read_raw_fif then needs to resolve the symlink on read to find it's next file as the next file name is written in the first one.

There are two options here:

- make `mne.io.read_raw_fif` resolve the symlink (not sure about potential side effects!!!)
- make `mne_bids.read_raw_bids` resolve the symlink (potential breakage will be minimized, as the use case is much more "defined" then in "pure" MNE-Python)

So I'm currently working on the 2nd option (adding this to MNE-BIDS) but I'm open to suggestions to move this upstream.

Doing it here also has the advantage that we'll have something ready much sooner (i.e., today, instead of … however long it may take to get this into MNE :))

agramfort (approved):
looks good !

agramfort:
@hoechenberger merge when green !

hoechenberger:
I believe tests are not yet covering everything. Will aim for 100% coverage.

agramfort (approved):

PR #788, opened 2021-04-30
When reading FIFF files, "allow Maxshield" by default

hoechenberger:
In `main`, when trying to read a FIFF file that was recorded with activated Maxshield and hasn't been Maxwell-filtered yet, one needed to explicitly pass `extra_params=dict(allow_maxshield=True)` to `read_raw_bids()`.

However, since we currently only support reading and writing "raw" data properly (i.e., no derivatives), this unncessarily complicates things: raw data is often not Maxwell-filtered, and Maxwell-filtering is one of the very first processing steps after loading the data. In contrast, IF the data was already Maxwell-filtered on the device, this would be clearly recognizable by its `proc` entity, which should indicate such processing, making any further complications just for the sake of "making users aware they need to Maxwell-filter" a bit pointless.

So this commit makes it such that FIFF files will be read with `allow_maxshield=True` by default, unless explicitly disabled via e.g. `extra_params=dict(allow_maxshield=False)`.

I also removed the `--allow_maxshield` switch from the `raw_to_bids` command, as this seems similarly superfluous and is not in line with how `write_raw_bids()` handles things (auto-sets `allow_maxshield=True`)

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -45,7 +45,9 @@ API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Writing datasets via :func:`write_raw_bids`, will now never overwrite ``dataset_description.json`` file, by `Adam Li`_ (:gh:`765`)
-- When writing BIDS datasets, MNE-BIDS now tags them as BIDS 1.6.0 (we previously tagged them as BIDS 1.4.0), `Richard Höchenberger`_ (:gh:`782`)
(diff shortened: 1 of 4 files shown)


agramfort (approved):

sappelhoff (approved):
I am impartial on this, I you both think it's an improvement I'll trust your choice :+1:

SophieHerbst:
A bit late to the party, sorry. 
I think it would still be good to issue a warning when loading data recorded with Maxshield on to remind the user to apply maxfilter.

hoechenberger:
I think MNE issues a warning nevertheless. Now I'm wondering why we don't see this in our tests 🤔

PR #794, opened 2021-05-05
MRG: Add cHPI info to MEG sidecars

### LINKED ISSUE #792: Add cHPI info to MEG sidecars
MNE now allows to retrieve HPI info: https://github.com/mne-tools/mne-python/pull/9369

We can use that to write the following info to `*_meg.json`:

Screen Shot 2021-05-04 at 22 31 47

See: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#sidecar-json-_megjson

hoechenberger:
Fixes #792

PR Description
--------------

Describe your PR here

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -241,7 +241,7 @@ def _handle_scans_reading(scans_fname, raw, bids_path, verbose=False):
 
 
 def _handle_info_reading(sidecar_fname, raw, verbose=None):
-    """Read associated sidecar.json and populate raw.
+    """Read associated sidecar JSON and populate raw.
 
     Handle PowerLineFrequency of recording.
     """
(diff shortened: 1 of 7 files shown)


mne_bids/read.py
```diff
@@ -268,6 +268,29 @@ def _handle_info_reading(sidecar_fname, raw, verbose=None):
...
+                f'stored in the raw data:\n    {hpi_freqs_data}\n'
+                f'Cannot proceed.'
+            )
+    else:
+        if raw.info['hpi_subsystem']:
+            logger.info('Dropping cHPI information stored in raw data, '
+                        'following specification in sidecar file')
+        raw.info['hpi_subsystem'] = None
+        raw.info['hpi_meas'] = []
+
```
sappelhoff, line 293: I don't get this part:

- when chpi info is present in BIDS but conflicts with raw (either absent in raw, or different in raw), you raise
- when chpi info is absent in BIDS, but present in raw, you remove the info from raw

This would make more sense to me:
- when chpi info is present in BIDS: use it ... if it conflicts with raw, send a warning, but overwrite raw with BIDS info
- when chpi info is absent in BIDS, but present in raw: send warning that BIDS has no info ... but we'll be using what's in raw

what was your reasoning for these cases?

sappelhoff (commented):
thanks @hoechenberger looks clean to me, just one confusion I have (see below).

mne_bids/read.py
hoechenberger: > I don't get this part:
> 
> * when chpi info is present in BIDS but conflicts with raw (either absent in raw, or different in raw), you raise

This is correct.

> * when chpi info is absent in BIDS, but present in raw, you remove the info from raw

This is not exactly correct: only if there **is** a `ContinuousHeadLocalization` setting in the sidecar **and** it is set to `false`, we drop the info from the raw.

If we dropped the info whenever `ContinuousHeadLocalization` was missing, we'd just throw away some data that's actually there … Does that make sense?

> 
> This would make more sense to me:
> 
> * when chpi info is present in BIDS: use it ... if it conflicts with raw, send a warning, but overwrite raw with BIDS info

The issue here is that the cHPI info stored in BIDS is not sufficient to restore much of anything that would actually be needed to to any kind of cHPI-related processing. The sidecar really stores "there was an cHPI in use" and "those were the frequency bands that the cHPI coils occupied". Really just very "meta" :)

sappelhoff: I see, thanks for the explanation :thinking: so BIDS would need additional metadata fields for cHPI-related data if one wants to **really** make use of it?

In any case, this makes sense to me now, +1 for merge

sappelhoff (approved):

agramfort (approved):

agramfort:
Thx @hoechenberger

hoechenberger:
@sappelhoff Let's just see how this works out in practice and adjust things if we figure it's not optimal! :) But I guess for now this should be okay

PR #812, opened 2021-06-22
MRG: Wipe out all measurement date info when reading scans.tsv

hoechenberger:
If we find the acquisition date in `scans.tsv`, we use it to replace `raw.info['meas_date']` in `main`.

However, the acquisition time might be stored in a few other places too, which can lead to issues when trying to write the file again. For example, I updated `scans.tsv` of one of my datasets so MNE could load it properly. But when trying to write the data again, MNE raised an exception, because `raw.info['meas_id']['secs']` was out of range.

This PR ensures that these problems are a thing of the past 💪

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -237,6 +237,11 @@ def _handle_scans_reading(scans_fname, raw, bids_path, verbose=False):
         if verbose:
             logger.debug(f'Loaded {scans_fname} scans file to set '
                          f'acq_time as {acq_time}.')
+        # First set meas_date to None and call anonymize_info to remove any
+        # traces of the measurement date we wish to replace – it might lurk out
+        # in more places than just raw.info['meas_date']!
+        raw.set_meas_date(None)
+        mne.io.anonymize_info(info=raw.info, keep_his=True)
         raw.set_meas_date(acq_time)
     return raw
 
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -237,11 +237,11 @@ def _handle_scans_reading(scans_fname, raw, bids_path, verbose=False):
         if verbose:
             logger.debug(f'Loaded {scans_fname} scans file to set '
                          f'acq_time as {acq_time}.')
-        # First set meas_date to None and call anonymize_info to remove any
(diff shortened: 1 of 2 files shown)


mne_bids/read.py
```diff
@@ -237,6 +237,11 @@ def _handle_scans_reading(scans_fname, raw, bids_path, verbose=False):
         if verbose:
             logger.debug(f'Loaded {scans_fname} scans file to set '
                          f'acq_time as {acq_time}.')
+        # First set meas_date to None and call anonymize_info to remove any
+        # traces of the measurement date we wish to replace – it might lurk out
+        # in more places than just raw.info['meas_date']!
+        raw.set_meas_date(None)
+        mne.io.anonymize_info(info=raw.info, keep_his=True)
```
sappelhoff, line 244: Why not call `raw.anonymize` as is recommended in the docstr? https://mne.tools/dev/generated/mne.io.anonymize_info.html

Also: The comment doesn't really tell me **where** a meas date might still lurk after purging it via `raw.set_meas_date(None)` :thinking: That is: the anonamiyze_info call seems unneeded (with the information that I have from its docstr and your comment)

hoechenberger: Would the info I added to my PR description help?


> However, the acquisition time might be stored in a few other places too, which can lead to issues when trying to write the file again. For example, I updated scans.tsv of one of my datasets so MNE could load it properly. But when trying to write the data again, MNE raised an exception, because raw.info['meas_id']['secs'] was out of range.

hoechenberger: > Why not call `raw.anonymize` as is recommended in the docstr?

It's the same thing, yes, can do!

sappelhoff: > Would the info I added to my PR description help?

if you can get it into a condensed form, then it'd improve the current comment :)

apart from that, LGTM

sappelhoff (approved):
Thanks @hoechenberger +1 to merge once CIs are green

agramfort (approved):

PR #819, opened 2021-06-29
[MRG] Add preloaded option to write_raw_bids

AKSoo:
PR Description
--------------

Add ``preloaded`` option to ``mne_bids.write_raw_bids`` as discussed in #803. Also solves #714 partially.

Merge checklist
---------------

- [ ] All comments resolved
- [x] This is not your own PR
- [ ] All CIs are happy


mne_bids/tests/test_write.py
```diff
@@ -509,7 +509,7 @@ def test_fif(_bids_validate, tmpdir):
     raw = _read_raw_fif(raw_fname, preload=True)
     with pytest.raises(ValueError, match='preloaded'):
         write_raw_bids(raw, bids_path_meg, events_data=events,
-                       event_id=event_id, overwrite=False)
+                       event_id=event_id, preloaded=False, overwrite=False)
```
jasmainak, line 512: I like `allow_preloaded` more that @agramfort suggested :)

can you update a test that checks that the error is not raised when `preload=False` and `allow_preloaded=True` ?

hoechenberger: +1 for allow_preload(ed)

AKSoo: > can you update a test that checks that the error is not raised when `preload=False` and `allow_preload=True` ?

Would you like a simple test for no error or a full test that writes, reads, and checks that the data are equal?

jasmainak: a simple test is fine, maybe with the `EvokedArray` example that you shared when reporting the issue. So we know that it works

AKSoo: As you [mentioned](https://github.com/mne-tools/mne-bids/issues/803#issuecomment-855973739), looks like format should be also explicit in this case. Some filename checks also have to be skipped, since preloaded data are not necessarily tied to a file. I will add a separate test for this.

AKSoo:
I think `write_raw_bids` is too monolithic and could really benefit from a refactor that breaks its subroutines (handling anonymize, emptyroom, events, coordinates, etc. etc.) into different functions. This would improve the function's readability and reduce redundancies, which would help with adding future features and debugging.

mne_bids/read.py
```diff
@@ -154,7 +154,7 @@ def _read_events(events_data, event_id, raw, verbose=None):
         verbose=verbose
     )
     all_dur = raw.annotations.duration
-    if all_events.size == 0 and 'rest' not in raw.filenames[0]:
+    if all_events.size == 0:
```
hoechenberger, line 157: Why did you change this? :)
This was here specifically to suppress a warning in case of resting-state data

hoechenberger:
> I think `write_raw_bids` is too monolithic and could really benefit from a refactor that breaks its subroutines (handling anonymize, emptyroom, events, coordinates, etc. etc.) into different functions.

Yes, there are many parts of MNE-BIDS that could use a good refactoring! If you're motivated to work on this, please open an issue so we can discuss how to best proceed. 👌

mne_bids/read.py
AKSoo: This was causing an error when `raw.filename` did not exist.

Searching for 'rest' in file name seems to be a dubious way of checking for resting state data. It would believe any subject, session, or task with string 'rest' to be resting-state data (e.g. watching clips from "Forrest Gump"). On the other hand, non-resting data with no events are also possible.

hoechenberger: It's clear that the existing heuristic for detecting resting-state data could be improved, but it was added because there was an actual demand / use case. So please either keep it in or try to make it better, but don't just remove it.

AKSoo: Changing `_read_events` function seemed dangerous, so I have improved the warning based on BIDS recommendations.

mne_bids/write.py
```diff
@@ -1195,12 +1207,14 @@ def write_raw_bids(raw, bids_path, events_data=None,
...
+    if raw.preload is not False and not allow_preload:
+        raise ValueError('The data is already loaded from disk and may be '
+                         'altered. See warning for "allow_preload".')
 
-    if raw.preload is not False:
-        raise ValueError('The data should not be preloaded.')
+    if ((not hasattr(raw, 'filenames') or raw.filenames[0] is None) and
+            not allow_preload):
+        raise ValueError('raw.filenames is missing. Please set raw.filenames '
+                         'as a list with the full path of original raw file.')
```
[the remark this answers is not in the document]
AKSoo: That is a good point. I can't think of when non-preloaded raw would be missing `filenames`. However, I am just maintaining the behavior of the existing code, which also checked `filenames` after `raw.preload`. This is tested by running `del raw._filenames` and meddling with the MNE backend, which seems like a highly unrealistic use case.

mne_bids/write.py
```diff
@@ -1125,6 +1124,19 @@ def write_raw_bids(raw, bids_path, events_data=None,
         ``bids_path`` and ``empty_room`` are the same. Pass ``None``
         (default) if you do not wish to specify an associated empty-room
         recording.
+    allow_preload : bool
+        If ``True``, allow writing of preloaded raw objects (i.e.,
+        ``raw.preload`` is ``True``). Because the original file is ignored, you
+        must specify what ``format`` to write (not ``auto``).
```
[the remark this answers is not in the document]
AKSoo: According to docstring, `'auto'` tries to infer format from filenames, which preloaded data may not necessarily have. However, we could implement `'auto'` by inferring from data type. MNE `raw.save` defaults to FIF, but BIDS recommends saving EEG data as BrainVision or EDF.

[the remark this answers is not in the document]
agramfort: my bad I read format at float32, int16 or float64 :) forget it ...

mne_bids/write.py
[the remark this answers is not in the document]
agramfort: can you git blame to know where this comes from?

agramfort:
@AKSoo you have a conflict to fix

mne_bids/write.py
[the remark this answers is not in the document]
AKSoo: Looks like `raw.preload` check was added after `raw.filenames` check for different reasons. They do seem redundant now though.

I am not familiar enough with MNE to know if it has any explicit checks that guarantees non-preloaded raws have `filenames`, but we could remove one of them.

mne_bids/tests/test_write.py
```diff
@@ -2820,4 +2846,4 @@ def test_write_raw_special_paths(tmpdir, dir_name):
 
     root = Path(tmpdir) / dir_name
     bids_path = _bids_path.copy().update(root=root)
-    write_raw_bids(raw=raw, bids_path=bids_path)
+    write_raw_bids(raw=raw, bids_path=bids_path)
```
sappelhoff, line 2849:
```suggestion
    write_raw_bids(raw=raw, bids_path=bids_path)

```

mne_bids/write.py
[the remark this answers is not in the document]
AKSoo: I decided to just remove the `raw.filenames` check.

In addition to being redundant, the error message doesn't make much sense either because MNE actually doesn't let you change `raw.filenames`. The only way to alter it is through a backend variable `raw._filenames`, which is probably not intended to be used that way.

agramfort (approved):
@AKSoo can you just add a what's new entry?

in this section:

https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst#api-and-behavior-changes

🙏

mne_bids/read.py
```diff
@@ -155,11 +155,14 @@ def _read_events(events_data, event_id, raw, verbose=None):
...
     all_dur = raw.annotations.duration
-    if all_events.size == 0 and 'rest' not in raw.filenames[0]:
-        warn('No events found or provided. Please add annotations '
-             'to the raw data, or provide the events_data and '
-             'event_id parameters. If this is resting state data '
-             'it is recommended to name the task "rest".')
+
+    # warn no events if not rest
+    if (all_events.size == 0 and hasattr(raw, 'filenames') and raw.filenames[0]
+            is not None and '_task-rest' not in raw.filenames[0]):
```
hoechenberger, line 161: Can this not be simplified to
```suggestion
    if (all_events.size == 0 and  '_task-rest' not in raw.filenames[0]):
```
?

@agramfort

AKSoo: No, this function is still called when `filenames` may not exist (preloaded). It is necessary to check whether it exists before checking for "task-rest".

AKSoo: Actually, I just realized checking `raw.filenames` doesn't make sense, since the function shouldn't care what people name their files before BIDS conversion.

I will just catch this warning outside the function after checking `bids_path` task for "rest".

doc/whats_new.rst
```diff
@@ -55,6 +55,7 @@ API and behavior changes
...
 - If :func:`mne_bids.read_raw_bids` encounters raw data with the ``STI 014`` stimulus channel and this channel is not explicitly listed in ``*_channels.tsv``, it is now automatically removed upon reading, by `Richard Höchenberger`_ (:gh:`823`)
 - :func:`mne_bids.get_anat_landmarks` was added to clarify and simplify the process of generating landmarks that now need to be passed to :func:`mne_bids.write_anat`; this depreciates the arguments ``raw``, ``trans`` and ``t1w`` of :func:`mne_bids.write_anat`, by `Alex Rockhill`_ and `Alexandre Gramfort`_ (:gh:`827`)
+- :func:`write_raw_bids` now accepts preloaded raws as input with some caveats if the new parameter ``allow_preload`` is explicitly set to True. This enables some preliminary support for uncommon file formats, generated data, processed derivatives etc., by `Sin Kim`_ (:gh:`819`)
```
hoechenberger, line 58:
```suggestion
- :func:`write_raw_bids` now accepts preloaded raws as input with some caveats if the new parameter ``allow_preload`` is explicitly set to ``True``. This enables some preliminary support for uncommon file formats, generated data, processed derivatives etc., by `Sin Kim`_ (:gh:`819`)
```

hoechenberger (approved):

hoechenberger:
Thanks @AKSoo!! This is a great addition on the way to supporting derivatives

AKSoo:
I just edited the event warning to check the output `bids_path` not input raw... I will open a quick new pull request. @hoechenberger

hoechenberger:
Oh, cool, sorry about merging prematurely then! Yes please feel free to open a followup PR! Your work is greatly appreciated!!

mne_bids/read.py
AKSoo: resolved with #838

PR #823, opened 2021-07-02
MRG: Handle sidecar <> raw disparities in channels

hoechenberger:
@agramfort feel free to take over, this seems to fix your problem

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -43,6 +43,7 @@ Enhancements
 - :func:`mne_bids.write_raw_bids` now stores information about continuous head localization measurements (e.g., Elekta/Neuromag cHPI) in the MEG sidecar file, by `Richard Höchenberger`_ (:gh:`794`)
 - :func:`mne_bids.write_raw_bids` gained a new parameter `empty_room` that allows to specify an associated empty-room recording when writing an MEG data file. This information will be stored in the ``AssociatedEmptyRoom`` field of the MEG JSON sidecar file, by `Richard Höchenberger`_ (:gh:`795`)
 - Added support for the new channel type `'dbs'` (Deep Brain Stimulation), which was introduced in MNE-Python 0.23, by `Richard Köhler`_ (:gh:`800`)
+- :func:`mne_bids.read_raw_bids` now warns in many situations when it encounters a mismatch between the channels in ``*_channels.tsv`` and the raw data, by `Richard Höchenberger`_ (:gh:`823`)
 
 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


agramfort (commented):
thx !

you plan to add a test @hoechenberger ?

hoechenberger:
If you think this is good, I can add a test!

@sappelhoff WDYT? Issue was that we had data with channels listed in the sidecar, which were not present in the raw. (HPI channels I believe. One could argue that these should be specified differently then -- which BIDS readily allows -- however we still have problematic datasets out in the wild we'll have to deal with...)

sappelhoff (approved):
looks like a good approach to me and I like the warnings. :+1:

hoechenberger:
@sappelhoff @agramfort I've added tests, good to go from my end

hoechenberger:
the KIT test fails because after writing, `channels.tsv` doesn't list `STI 014`, but this channel is still included in the raw data (or, rather, synthesized by MNE-Python upon file reading, right?)

How shall we deal with this?

hoechenberger:
Related:
https://github.com/mne-tools/mne-bids/issues/51

sappelhoff:
>  synthesized by MNE-Python upon file reading, right?

I am not 100% sure, but I thought that we wanted to stop synthesizing stim channels in MNE-Python🤔 or was that only for some particular cases (e.g., for EEG only data)?

I think it'd be good to ignore any channels that MNE-Python synthesizes ... or stop it from *automatically* synthesizing them in the first place.

hoechenberger:
@agramfort any insights on this one?

hoechenberger:
My personal opinion is that since all events should be listed in the sidecar, MNE-BIDS could simply drop `STI 014` after reading the file …

sappelhoff:
> My personal opinion is that since all events should be listed in the sidecar, MNE-BIDS could simply drop STI 014 after reading the file …

Sounds reasonable to me. If it's synthesized by MNE during reading, we wouldn't even have to log information about it. Else, an info/warning may be good.

hoechenberger:
> Sounds reasonable to me. If it's synthesized by MNE during reading, we wouldn't even have to log information about it. Else, an info/warning may be good.

Thing is, I don't know if this channel is always synthesized (and for which file formats) …

What I would propose is:
When reading data and `STI 014` is not present in `channels.tsv`, but appears in the raw, remove it from the raw. This should be rather conservative.

Thoughts?

sappelhoff (approved):
LGTM, provided CIs pass, +1 to merge

agramfort (approved):

agramfort:
thx @hoechenberger

PR #824, opened 2021-07-04
add plot_anat_landmarks function

agramfort:
let me know what you think

### DIFF
--- a/examples/convert_mri_and_trans.py
+++ b/examples/convert_mri_and_trans.py
@@ -51,6 +51,7 @@ from mne.source_space import head_to_mri
 
 from mne_bids import (write_raw_bids, BIDSPath, write_anat,
                       get_head_mri_trans, print_dir_tree)
+from mne_bids.viz import plot_anat_landmarks
 
 ###############################################################################
 # We will be using the `MNE sample data <mne_sample_data_>`_ and write a basic
@@ -134,16 +135,20 @@ t1w_bids_path = write_anat(
 anat_dir = t1w_bids_path.directory
 
 ###############################################################################
-# Let's have another look at our BIDS directory
+# Let's have another look at our BIDS directory and plot the written landmarks
 print_dir_tree(output_path)
+
+plot_anat_landmarks(t1w_bids_path, vmax=160)
+plt.suptitle('T1 MRI')
 
 ###############################################################################
 # Our BIDS dataset is now ready to be shared. We can easily estimate the
-# transformation matrix using ``MNE-BIDS`` and the BIDS dataset.
+# transformation matrix using ``MNE-BIDS`` and the BIDS dataset since we
+# have now anatomical landmarks.
 estim_trans = get_head_mri_trans(bids_path=bids_path)
 
 ###############################################################################
-# Finally, let's use the T1 weighted MRI image and plot the anatomical
+# One can also directly use the T1 weighted MRI image and plot the anatomical
 # landmarks Nasion, LPA, and RPA onto the brain image. For that, we can
 # extract the location of Nasion, LPA, and RPA from the MEG file, apply our
 # transformation matrix :code:`trans`, and plot the results.
@@ -169,7 +174,7 @@ mri_pos = head_to_mri(pos=pos,
 t1_nii_fname = op.join(anat_dir, 'sub-01_ses-01_T1w.nii.gz')
 
 # Plot it
-fig, axs = plt.subplots(3, 1, figsize=(7, 7), facecolor='k')
+fig, axs = plt.subplots(3, 1, figsize=(7, 7))
 for point_idx, label in enumerate(('LPA', 'NAS', 'RPA')):
     plot_anat(t1_nii_fname, axes=axs[point_idx],
               cut_coords=mri_pos[point_idx, :],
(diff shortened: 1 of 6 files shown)


agramfort:
ok this one should be good for review. See rendered doc:

https://4812-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_mri_and_trans.html

I will add a what's new entry at the end.

examples/convert_mri_and_trans.py
```diff
@@ -134,16 +135,20 @@
 anat_dir = t1w_bids_path.directory
 
 ###############################################################################
-# Let's have another look at our BIDS directory
+# Let's have another look at our BIDS directory and plot the written landmarks
 print_dir_tree(output_path)
 
+plot_anat_landmarks(t1w_bids_path, vmax=160)
+plt.suptitle('T1 MRI')
```
hoechenberger, line 142: Can this be moved into `plot_anat_landmarks`? It would be nice if the figure came with a sensible title by default

examples/convert_mri_and_trans.py
```diff
@@ -134,16 +135,20 @@
...
 ###############################################################################
 # Our BIDS dataset is now ready to be shared. We can easily estimate the
-# transformation matrix using ``MNE-BIDS`` and the BIDS dataset.
+# transformation matrix using ``MNE-BIDS`` and the BIDS dataset since we
+# have now anatomical landmarks.
 estim_trans = get_head_mri_trans(bids_path=bids_path)
 
 ###############################################################################
-# Finally, let's use the T1 weighted MRI image and plot the anatomical
+# One can also directly use the T1 weighted MRI image and plot the anatomical
```
hoechenberger, line 151:
```suggestion
# One can also directly use the T1-weighted MR image and plot the anatomical
```

examples/convert_mri_and_trans.py
```diff
@@ -134,16 +135,20 @@
...
 # landmarks Nasion, LPA, and RPA onto the brain image. For that, we can
 # extract the location of Nasion, LPA, and RPA from the MEG file, apply our
 # transformation matrix :code:`trans`, and plot the results.
```
hoechenberger, line 154: Wondering if we should start off this paragraph differently or put it into a new subsection. The first sentence made me think, "umm, but we just DID that??" until I got to the point where it says, " ... from the MEG file"

examples/convert_mri_and_trans.py
hoechenberger: Would also be good to briefly explain what `vmax` is doing here?

mne_bids/tests/test_viz.py
```diff
@@ -0,0 +1,36 @@
+from pathlib import Path
+
+import pytest
+
+import mne
+from mne.datasets import testing
+from mne_bids.viz import plot_anat_landmarks
+from mne_bids import BIDSPath, write_anat
+
+from mne.utils import requires_nibabel, requires_version
```
hoechenberger, line 10:
```suggestion
import mne
from mne.datasets import testing
from mne.utils import requires_nibabel, requires_version
from mne_bids.viz import plot_anat_landmarks
from mne_bids import BIDSPath, write_anat
```

mne_bids/viz.py
```diff
@@ -0,0 +1,70 @@
+import json
+
+import numpy as np
+from scipy import linalg
+
+import nibabel as nib
+from mne.utils import warn
+import mne
```
hoechenberger, line 8:
```suggestion
import numpy as np
from scipy import linalg
import nibabel as nib

import mne
from mne.utils import warn
```

mne_bids/viz.py
```diff
@@ -0,0 +1,70 @@
...
+
+def plot_anat_landmarks(bids_path, vmax=None, show=True):
+    """Plot anatomical landmarks attached to an MRI image.
+
+    Parameters
+    ----------
+    bids_path : mne_bids.BIDSPath
+        Path of the MRI image.
+    vmax : float
+        Maximum colormap value.
```
hoechenberger, line 19: Should we limit the acceptable range, e.g. to the interval [0, 255]?

mne_bids/viz.py
```diff
@@ -0,0 +1,70 @@
...
+    show : bool
+        Call pyplot.show() at the end. Defaults to True.
```
hoechenberger, line 21:
```suggestion
        Whether to show the figure after plotting. Defaults to ``True``.
```

mne_bids/viz.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure | None
+        The figure object containing the plot. None if no landmarks
+        are avaiable.
```
hoechenberger, line 27: Shouldn't we raise an exception if there are no landmarks?

mne_bids/viz.py
```diff
@@ -0,0 +1,70 @@
...
+        facecolor="w")
+
+    for point_idx, (label, ras_pos) in enumerate(coords_dict.items()):
+        plot_anat(
+            str(bids_path), axes=axs[point_idx], cut_coords=ras_pos,
+            title=label, vmax=vmax,
+        )
+
+    if show:
+        plt.show()
```
hoechenberger, line 68: Should this better be `fig.show()`?

sappelhoff (commented):
Why do we need this? Is nilearn's `plot_anat` not sufficient?

agramfort:
@sappelhoff it's ok for me if we don't merge this. It was a very convenient tool for debugging recently and I see this as being potentially useful to check some existing dataset. but It's ok if we don't merge this. Maybe it's not useful in the end...

sappelhoff:
Okay, I'll leave the decision to you and @hoechenberger - I am 50-50

agramfort:
I would say let's wait.... it also adds new dependencies on nilearn and
matplotlib which is not great....

>

Issue #822, opened 2021-07-02
Bug trans with get_head_mri_trans on ds000117


agramfort:
Download some files from ds117 with openneuro-py

```
openneuro-py download --dataset=ds000117 --include "sub-01"
openneuro-py download --dataset=ds000117 --include "derivatives/freesurfer/sub-01"
```

then run the following script:

```python
from pathlib import Path
import json

import nibabel as nib
import numpy as np
from scipy import linalg

import matplotlib.pyplot as plt
from nilearn.plotting import plot_anat

import mne
from mne_bids import BIDSPath, read_raw_bids, get_head_mri_trans

bids_root = Path("./ds000117")

raw_fname = BIDSPath(
    subject="01",
    root=bids_root,
    session="meg",
    suffix="meg",
    run="01",
    extension=".fif",
    task="facerecognition",
)

raw = read_raw_bids(raw_fname)

t1_fname = BIDSPath(
    subject="01",
    root=bids_root,
    session="mri",
    suffix="T1w",
    acquisition="mprage",
    extension=".nii.gz",
)
t1_json_fname = t1_fname.copy().update(extension=".json")

t1_img = nib.load(str(t1_fname))

t1_json = json.load(open(t1_json_fname))
mri_coords_dict = t1_json.get("AnatomicalLandmarkCoordinates", dict())
trans = get_head_mri_trans(raw_fname, t1_bids_path=t1_fname)

mgh = nib.MGHImage(t1_img.dataobj, t1_img.affine)
ras2vox = mgh.header.get_ras2vox()
vox2ras = linalg.inv(ras2vox)

########################################################################
# Plot it with nilearn
fig, axs = plt.subplots(3, 1, figsize=(7, 7), facecolor="k")
if "NAS" in mri_coords_dict:
    labels = ("LPA", "NAS", "RPA")
else:
    labels = ("LPA", "Nasion", "RPA")

for point_idx, label in enumerate(labels):
    vox_pos = np.array(mri_coords_dict[label])
    ras_pos = mne.transforms.apply_trans(vox2ras, vox_pos)

    plot_anat(
        str(t1_fname), axes=axs[point_idx], cut_coords=ras_pos, title=label, vmax=500
    )
fig.savefig("fids.png")
plt.show()

########################################################################
# Plot coreg in 3D
fs_subject = "anat"
subjects_dir = bids_root / "derivatives" / "freesurfer" / "sub-01" / "ses-mri"

[...]

PR #825, opened 2021-07-04
enforce image is moved to LIA orientation before getting the trans

agramfort:
attempt to fix #822

but still not ok...

it starts to really get on my nerves...

coreg

PR #826, opened 2021-07-04
[BUG] fix orientations if not LIA

alexrockhill:
This fixes non-LIA orientation image determining of `trans` since freesurfer stores images internally in LIA.

Fixes https://github.com/mne-tools/mne-bids/issues/822, adds on https://github.com/mne-tools/mne-bids/pull/825

Screen Shot 2021-07-04 at 10 57 35 AM

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -69,6 +69,7 @@ Bug fixes
 - :func:`mne_bids.write_raw_bids` now works across data types with ``overwrite=True``, by `Alexandre Gramfort`_ (:gh:`791`)
 - :func:`mne_bids.read_raw_bids` didn't always replace all traces of the measurement date and time stored in the raw data with the date found in `*_scans.tsv`, by `Richard Höchenberger`_ (:gh:`812`, :gh:`815`)
 - :func:`mne_bids.read_raw_bids` crashed when the (optional) ``acq_time`` column was missing in ``*_scans.tsv``, by `Alexandre Gramfort`_ (:gh:`814`)
+- :func:`mne_bids.get_head_mri_trans` gave incorrect results when the T1 image was not in LIA format, now all formats function properly, by `Alex Rockhill`_ (:gh:`826`)
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -782,21 +782,29 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None):
                            'with "{}". Tried: "{}" but it does not exist.'
                            .format(t1w_json_path, t1w_path))
     t1_nifti = nib.load(t1w_path)
+
     # Convert T1 to LIA (freesurfer stores scanner RAS in LIA)
-    ori_trans = nib.orientations.ornt_transform(
-        nib.orientations.axcodes2ornt(
-            nib.orientations.aff2axcodes(t1_nifti.affine)),
-        nib.orientations.axcodes2ornt('LIA'))
-    t1_lia = t1_nifti.as_reoriented(ori_trans)
+    # If already in LIA, avoid potentially lossy conversion
+    if nib.orientations.aff2axcodes(t1_nifti.affine) != tuple('LIA'):
+        # Reorient T1
+        ori_trans = nib.orientations.ornt_transform(
+            nib.orientations.axcodes2ornt(
+                nib.orientations.aff2axcodes(t1_nifti.affine)),
+            nib.orientations.axcodes2ornt('LIA'))
+        t1_lia = t1_nifti.as_reoriented(ori_trans)
+
+        # Convert the landmarks to LIA as well
+        # Original orientation -> real world coordinates
+        mri_landmarks = apply_trans(t1_nifti.affine, mri_landmarks)
+        # Real world coordinates -> LIA
+        mri_landmarks = apply_trans(np.linalg.inv(t1_lia.affine),
+                                    mri_landmarks)
+
+        # Assign nifti to transformed image
+        t1_nifti = t1_lia
 
     # Convert to MGH format to access vox2ras method
-    t1_mgh = nib.MGHImage(t1_lia.dataobj, t1_lia.affine)
-
-    # Convert the landmarks to LIA as well
-    # Original orientation -> real world coordinates
-    mri_landmarks = apply_trans(t1_nifti.affine, mri_landmarks)
-    # Real world coordinates -> LIA
-    mri_landmarks = apply_trans(np.linalg.inv(t1_lia.affine), mri_landmarks)
+    t1_mgh = nib.MGHImage(t1_nifti.dataobj, t1_nifti.affine)
 
     # now extract transformation matrix and put back to RAS coordinates of MRI
     vox2ras_tkr = t1_mgh.header.get_vox2ras_tkr()
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -782,8 +782,21 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None):
                            'with "{}". Tried: "{}" but it does not exist.'
                            .format(t1w_json_path, t1w_path))
     t1_nifti = nib.load(t1w_path)
+    # Convert T1 to LIA (freesurfer stores scanner RAS in LIA)
+    ori_trans = nib.orientations.ornt_transform(
+        nib.orientations.axcodes2ornt(
+            nib.orientations.aff2axcodes(t1_nifti.affine)),
+        nib.orientations.axcodes2ornt('LIA'))
+    t1_lia = t1_nifti.as_reoriented(ori_trans)
+
     # Convert to MGH format to access vox2ras method
-    t1_mgh = nib.MGHImage(t1_nifti.dataobj, t1_nifti.affine)
+    t1_mgh = nib.MGHImage(t1_lia.dataobj, t1_lia.affine)
+
+    # Convert the landmarks to LIA as well
+    # Original orientation -> real world coordinates
+    mri_landmarks = apply_trans(t1_nifti.affine, mri_landmarks)
+    # Real world coordinates -> LIA
+    mri_landmarks = apply_trans(np.linalg.inv(t1_lia.affine), mri_landmarks)
 
     # now extract transformation matrix and put back to RAS coordinates of MRI
     vox2ras_tkr = t1_mgh.header.get_vox2ras_tkr()


hoechenberger (approved):
LGTM if it provides the correct results now. A changelog entry would be nice.

alexrockhill:
@agramfort said he would check it and that there was some weird freesurfer transform beyond this LIA thing so let's just make sure it works on his end.

agramfort:
it's not exactly the same that I obtain using the freesurfer T1.mgz file using this branch:

https://github.com/agramfort/mne-bids/tree/fs_req

and this script

```python
from pathlib import Path

import mne
from mne_bids import BIDSPath, read_raw_bids, get_head_mri_trans

bids_root = Path(".")

raw_fname = BIDSPath(
    subject="01",
    root=bids_root,
    session="meg",
    suffix="meg",
    run="01",
    extension=".fif",
    task="facerecognition",
)

raw = read_raw_bids(raw_fname)

t1_fname = BIDSPath(
    subject="01",
    root=bids_root,
    session="mri",
    suffix="T1w",
    acquisition="mprage",
    extension=".nii.gz",
)

t1_mgz = 'derivatives_fs/freesurfer/subjects/sub-01/mri/T1.mgz'

trans = get_head_mri_trans(raw_fname, t1_bids_path=t1_fname,
                           t1_freesurfer_fname=t1_mgz)
print(trans)

########################################################################
# Plot coreg in 3D
fs_subject = "anat"
subjects_dir = bids_root / "derivatives" / "freesurfer" / "sub-01" / "ses-mri"

# fs_subject = "sub-01"
# subjects_dir = bids_root / "derivatives_fs" / "freesurfer" / "subjects"

info = mne.io.read_info(raw_fname)
fig = mne.viz.plot_alignment(
    info=info,
    subject=fs_subject,
    subjects_dir=subjects_dir,
    trans=trans,
    surfaces="white",
    coord_frame="mri",
    meg="helmet",
    eeg=True,
    dig=True,
)

fig.plotter.screenshot("coreg_with_fs.png")
```

I've put my T1.mgz file at: https://www.dropbox.com/s/59s9d00oryo4kut/T1.mgz?dl=0

With Freesurfer:

coreg_with_fs

With this branch:

coreg_ori_branch

I plot in the MRI coord system to amplify the difference.

do you replicate @alexrockhill ? thoughts?

alexrockhill:
Can you point me to the difference? Sorry I feel like Eric with the MR-CT alignment stuff where I could see a big difference and he couldn't see it lol

Where should I be looking in the image?

agramfort:
here is the affine from the image you convert to LIA:

voxel to ras transform:
               -1.0000  -0.0091   0.0037   100.4465
                0.0034   0.0398   0.9992   -98.2742
                0.0092  -0.9992   0.0398   142.5270
                0.0000   0.0000   0.0000     1.0000

and the one from T1.mgz

voxel to ras transform:
               -1.0000   0.0000  -0.0000   132.7760
                0.0000  -0.0000   1.0000   -94.0022
               -0.0000  -1.0000  -0.0000   149.5548
                0.0000   0.0000   0.0000     1.0000

it's close but not identical.

agramfort:
@alexrockhill download the 2 images and alternate them on your screen :)

alexrockhill:
> here is the affine from the image you convert to LIA:
> 
> voxel to ras transform:
> -1.0000 -0.0091 0.0037 100.4465
> 0.0034 0.0398 0.9992 -98.2742
> 0.0092 -0.9992 0.0398 142.5270
> 0.0000 0.0000 0.0000 1.0000
> 
> and the one from T1.mgz
> 
> voxel to ras transform:
> -1.0000 0.0000 -0.0000 132.7760
> 0.0000 -0.0000 1.0000 -94.0022
> -0.0000 -1.0000 -0.0000 149.5548
> 0.0000 0.0000 0.0000 1.0000
> 
> it's close but not identical.

Hmmm those are really far off on the translations, are you sure that's what's being compared?

alexrockhill:
Hmmm, I see now. Maybe it would help to have the seghead to get the ground truth? You're saying the T1.mgz one is the ground truth because it's been processed by freesurfer right? That makes sense to me, but just to make sure.

alexrockhill:
compare

alexrockhill:
alignment

Hmmm, this is with the alignment in this PR. Looks about as good as it can get to me. Checking the other version...

alexrockhill:
alignment_t1

Ok don't ask me to explain it, but this is what I got using the method posted in the last comment in https://github.com/mne-tools/mne-bids/issues/822.

alexrockhill:
So I'm more confused than ever about what is going on with the affine of the T1/the scanner RAS to surface RAS transform but on the bright side it does look like this PR fixes this specific example...

alexrockhill:
So there are differences between the `T1.mgz` affine and the affine from converting the `orig.mgz` or original image to LIA that we're tracking down what is happening there.

Here is also where freesurfer defines scanner space as LIA:
Screen Shot 2021-07-04 at 1 06 21 PM

alexrockhill:
So the T1 affine for this example is
```
array([[ -1.        ,  -0.00000015,  -0.00000003, 132.77600098],
           [  0.00000003,   0.00000001,   1.        , -94.00223541],
           [ -0.00000002,  -1.00000012,  -0.00000001, 149.55487061],
           [  0.        ,   0.        ,   0.        ,   1.        ]])
```

The affine from reorienting to LIA from the original orientation is

```
array([[ -0.99995208,  -0.00905184,   0.00373623, 100.44650959],
           [  0.00337327,   0.03978604,   0.99920255, -98.27423529],
           [  0.00919327,  -0.9991672 ,   0.03975359, 142.52701772],
           [  0.        ,   0.        ,   0.        ,   1.        ]])
```

and the affine from resample to `(256, 256, 256)` first is

```
array([[ -0.99995208,  -0.00905184,   0.00373623, 164.44344257],
           [  0.00337327,   0.03978604,   0.99920255, -98.49012464],
           [  0.00919327,  -0.9991672 ,   0.03975359, 141.93864847],
           [  0.        ,   0.        ,   0.        ,   1.        ]])
```

So there must be some rotation transform that freesurfer is applying that is not just zero-centering and going to LIA...

agramfort:
Either we find a way to replicate the import step from freesurfer with nibabel or I think we have to choice but to require to have the freesurfer T1.mgz when writing landmarks from trans and when reading trans from landmarks in from json

alexrockhill:
Ok, I've found the call in freesurfer that changes the affine: `mri_convert my_T1.nii.gz orig.mgz --conform` is what does it

The text output is
```
$Id: mri_convert.c,v 1.226 2016/02/26 16:15:24 mreuter Exp $
reading from /Users/alexrockhill/Downloads/test/mri/rawavg.mgz...
TR=0.00, TE=0.00, TI=0.00, flip angle=0.00
i_ras = (-0.00373623, -0.999203, -0.0397536)
j_ras = (0.00905184, -0.039786, 0.999167)
k_ras = (0.999952, -0.00337327, -0.00919327)
changing data type from float to uchar (noscale = 0)...
MRIchangeType: Building histogram 
Reslicing using trilinear interpolation
```

One worthwhile point is that this takes ~1 second and could just be redone and the output read in. The downside to that is that freesurfer would be required. Ideally we could recreate this histogram registration but realistically this might be a lot. A third option is that we could just merge this PR as is or with a `warning` that `AnatomicalLandmarks` maybe be off by as much as the off-diagonal components of the scan. This might actually not be too huge of an issue since these components tend to be minor. Or also we could take out all support for writing in anything other than voxel coordinates. This puts potentially more code on the user but it is definitely the most straightforward as far as what it means and not needing to convert.

alexrockhill:
Ok one last thought

```
pad = (np.array([256, 256, 256]) - np.array(t1_nifti.shape)) / 2
align_trans = np.zeros((4, 4))
align_trans[3, 3] = 1
align_trans[:3, :3] = np.linalg.inv(t1_nifti.affine[:3, :3])

fs_trans = np.dot(align_trans, t1_nifti.affine)
fs_trans[:3, 3] += np.array(t1_nifti.shape) + pad
```

yields

```
array([[  1.   ,  -0.   ,  -0.   ,  93.155],
       [ -0.   ,   1.   ,   0.   , 148.227],
       [ -0.   ,  -0.   ,   1.   , 132.463],
       [  0.   ,   0.   ,   0.   ,   1.   ]])
```

which is nearly the same as the shifts in the `T1.mgz`. It's not perfect so I'm not sure if this is what's happening though...

agramfort:
@alexrockhill nibabel as a CLI to do the same as mri_conform from freesurfer:

`nib-conform sub-01/ses-mri/anat/sub-01_ses-mri_acq-mprage_T1w.nii.gz nib_cli_conform.nii.gz --orientation LIA``

which should be equivalent to:

`mri_convert --conform sub-01/ses-mri/anat/sub-01_ses-mri_acq-mprage_T1w.nii.gz T1_conform.mgz`

it's close but not exactly the same:

```
(base)➜  ds000117 mri_info --vox2ras nib_cli_conform.nii.gz
  -0.99995   -0.00905    0.00374  132.44499
   0.00337    0.03979    0.99920  -98.38218
   0.00919   -0.99917    0.03975  142.23283
   0.00000    0.00000    0.00000    1.00000
(base)➜  ds000117 mri_info --vox2ras T1_conform.mgz
  -1.00000    0.00000   -0.00000  132.77597
   0.00000   -0.00000    1.00000  -94.00224
  -0.00000   -1.00000   -0.00000  149.55484
   0.00000    0.00000    0.00000    1.00000
```

sleeping over it I think we should enforce that image in `write_anat` is a freesurfer conformed image when a `trans` is passed or if landmarks are in surface RAS. And we should enforce that we pass the path to the T1.mgz file in `get_head_mri_trans` to get a valid trans file. The rest seems too much magic and too risky.

hoechenberger:
> sleeping over it I think we should enforce that image in `write_anat` is a freesurfer conformed image when a `trans` is passed or if landmarks are in surface RAS. And we should enforce that we pass the path to the T1.mgz file in `get_head_mri_trans` to get a valid trans file. The rest seems too much magic and too risky.

+1 on this, following the discussion(s)!!!

alexrockhill:
So the one mystery to me is that this PR works for your example as is, @agramfort. It seems to exact to be just luck.

There does seem to be one issue as well which is that the ``trans`` would be hard to estimate if the recon-all subject's directory is not provided in the dataset. That would be a lot more data to share. The only way around that as far as I can tell is to link to the `T1.mgz` specifically instead of providing the whole `subjects_dir` and `subject`.  This would be error-prone to me because it would be very hard to tell if the T1 is in fact the freesurfer T1. It is in LIA with no off-diagonal coordinates but that could also be other scans that have different translations, head-first supine is a typical MRI position.

Maybe the best long-term solution would be to add the `T1.mgz` affine to the ``trans`` object or something like that but that would be a huge overhaul.

agramfort:
or the trans should be from head to MRI RAS (and not surface RAS...)

but we make it work without this big change... let's try this first

>

PR #828, opened 2021-07-06
Update example entity key

rob-luke:
PR Description
--------------

Running the provided example did not work. So I changed the example to what does work.

Merge checklist
---------------

Maintainer, please confirm the following before merging:

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] PR description includes phrase "closes <#issue-number>"

PR #831, opened 2021-07-06
[MRG] Add tiny BIDS test dataset, fix doctests, and run it in CI

### LINKED ISSUE #829: bidspath example contains runtimewarning
https://mne.tools/mne-bids/dev/auto_examples/bidspath.html#bidspath-example

> /home/circleci/project/mne_bids/path.py:587: RuntimeWarning: Could not locate a data file of a supported format. This is likely a problem with your BIDS dataset. Please run the BIDS validator on your data. (root=my_bids_root, basename=sub-123_ses-test). []
  warn(msg)
  
and there are 2 more of those. At least it should be discussed in the example why there is such a warning.

sappelhoff:
follow up to #828
closes #829 

This adds a tiny (800kb) BIDS dataset to a new `tests/data` directory for testing purposes.

Code for generating the data is shipped in `mne_bids/tests/data/tiny_bids/code/make_tiny_bids_dataset.py`, this dataset should be easily extensible in the future to add MEG, MRIs, iEEG, etc. when the need for it arises.

I make use of that dataset in the doctests, which I fix in this PR and "turn on" in CI checks.

I also use that dataset in an example, fixing #829 

Adding the example dataset also showed me in its git diff that we have been writing TSV files without newline characters, violating UNIX/POSIX standards :wink: I remedied that.

Finally, this PR contains 3 "fixes" where I change json.dump to our internal _write_json, thereby simplifying some lines.

The diff is large, but most of it is the newly added data.

Merge checklist
---------------

- [x] All comments resolved
- [x] This is not your own PR
- [x] All CIs are happy


mne_bids/path.py
```diff
@@ -225,36 +225,39 @@ class BIDSPath(object):
...
     sub-test2_ses-one_task-mytask_ieeg.edf
     >>> # printing the BIDSPath will show relative path when
     >>> # root is not set
     >>> print(new_bids_path)
     sub-test2/ses-one/ieeg/sub-test2_ses-one_task-mytask_ieeg.edf
-    >>> new_bids_path.update(suffix='channels', extension='.tsv')
+    >>> new_bids_path = new_bids_path.update(suffix='channels',
+    ...                                      extension='.tsv')
     >>> # setting suffix without an identifiable datatype will
     >>> # result in a wildcard at the datatype directory level
```
sappelhoff, line 248: this probably needs to be overhauled, in my tests this was set to "ieeg", instead of being "*"

agramfort:
+1 for pytest --doctest-modules

is it a lot of work?

sappelhoff:
> is it a lot of work?

Maybe :thinking: Let's see, I'll work on this a bit in the next days.

sappelhoff:
I think the problem is that many examples rely on us having some files. So I'd have to "skip" executing most of the examples, which defeats the purpose. I think a solution would be to "ship" a tiny BIDS dataset as part of the test directory that we can use to actually execute the examples.

WDYT @hoechenberger @agramfort ?

agramfort:
how would you "ship" it?

>

sappelhoff:
> how would you "ship" it?

1. just putting it there ... like the data in MNE-Python's IO module
2. as a git submodule ... so that getting it is optional (unless you want to run tests)
3. ... do you have another idea?

agramfort:
if we manage to make it less than 30MB I would say why not to add it to the
git repo like mne/io/tests/data folder in mne-python

mne_bids/path.py
sappelhoff: done

sappelhoff (commented):
CircleCI fails with:

> ImportError: cannot import name 'head_to_mri' from 'mne.source_space'

I also get this problem locally. Has something changed upstream? cc @agramfort 

Otherwise this PR is ready for review cc @hoechenberger 


EDIT: got it, it's failing due to https://github.com/mne-tools/mne-python/pull/9543

agramfort (approved):
source dist is still tiny. Great job.

+1 for MRG when CIs are green.

mne_bids/path.py
```diff
@@ -142,6 +145,10 @@ def _find_matched_empty_room(bids_path):
     return best_er_bids_path
 
 
+@pytest.mark.xfail(condition="os.name == 'nt'",
+                   reason='doctest fails on Windows due to os.sep difference',
```
hoechenberger, line 149: Wait, shouldn't we assure that the `str()` and `repr()` are the same on all platforms?

sappelhoff: but would the paths then still work on Windows?

sappelhoff: maybe like this?

```diff
    def __str__(self):
        """Return the string representation of the path."""
-        return str(self.fpath)
+       return str(self.fpath.as_posix()) 
```

sappelhoff: I just pushed it. We'll see if it works, I think that's a good idea and simplifies the whole thing. And BIDS wants only forward slashes anyhow :-) Fingers crossed that the test suite works

hoechenberger: > but would the paths then still work on Windows?

My idea is that we wouldn't touch .fpath, just change the way we str() and repr() things :)

hoechenberger:
@sappelhoff Apparently you forgot one or two places? @ CI failing

sappelhoff:
one is about `bids_path.root`, which can return a str OR a Path --> In this case it returns a Path, and on Windows this happens to be a WindowsPath, not posix. :man_shrugging: 

The other one, I am still investigating

hoechenberger:
> The other one, I am still investigating

I think it's because in the test you're comparing to a path constructed via `os.path.join()`, which will use `\\` as separator by default on Windows.

hoechenberger:
> one is about `bids_path.root`, which can return a str OR a Path

Wait, why is that? IMHO it should always be a path …

sappelhoff:
> Wait, why is that? IMHO it should always be a path …

yes, also according to the attributes docs in the docstr. But see:

https://github.com/mne-tools/mne-bids/blob/2d8a7216b547baedd525991f97aa4537d5fe98dd/mne_bids/path.py#L442-L449

I am a little confused by that Union(str, Path)

sappelhoff:
Maybe the solution is:

```diff
    def __repr__(self):
        """Representation in the style of `pathlib.Path`."""
        return f'{self.__class__.__name__}(\n' \
-               f'root: {self.root}\n' \
+               f'root: {self.root.as_posix()}\n' \
               f'datatype: {self.datatype}\n' \
               f'basename: {self.basename})'
```

but root can be "None" :thinking:

sappelhoff:
I think there is something I don't understand in https://github.com/mne-tools/mne-bids/blob/2d8a7216b547baedd525991f97aa4537d5fe98dd/mne_bids/path.py#L442-L449


1. BIDSPath has an attribute "root" ... here defined with the `@property` decorator
2. the type hints show us it can return str or Path ... **Is this an error? I think it should be Path and None ... never str**
3. then it returns `self._root`, which is being set in the update method and is either None, or `Path(val).expanduser()`, where `val` is either Path or str
4. the setter is fine again: calling `bids_path.root = Path("/home/bids_ds")` (or with None, or a str) just calls "update" method again as in the step before

sappelhoff:
Do we need a `__str__` method for attributes? the `bids_path.root` attribute is just printed "as is" (a Path) :confused: :question:

mne_bids/path.py
sappelhoff: that solved the first problem - thanks!

hoechenberger:
> 1. BIDSPath has an attribute "root" ... here defined with the `@property` decorator
> 2. the type hints show us it can return str or Path ... **Is this an error? I think it should be Path and None ... never str**
> 3. then it returns `self._root`, which is being set in the update method and is either None, or `Path(val).expanduser()`, where `val` is either Path or str

Considering point 3, I would agree that the type hint is incorrect and should be `Optional[Path]`

sappelhoff:
> Considering point 3, I would agree that the type hint is incorrect and should be Optional[Path]

mmh okay - but it could be `None` though. What do you think of https://github.com/mne-tools/mne-bids/pull/831/commits/459c723fa4a6947d050917b0f4ec2b95c5dc9c39 ?

BTW: the errors are now fixed. I circumvented the repr/str issue for bids_path attributes by calling `.as_posix` on them in the example. I feel like further improving this should be a future PR (I am exhausted with this one)

hoechenberger:
> Do we need a `__str__` method for attributes? the `bids_path.root` attribute is just printed "as is" (a Path) 😕 ❓

we'd only need it for `root`, right? All other parts of the `BIDSPath` are strings or None anyway?
How would we do that? Would we need a new class `_BIDSRoot` or something?

hoechenberger:
> > Considering point 3, I would agree that the type hint is incorrect and should be Optional[Path]
> 
> mmh okay - but it could be `None` though. What do you think of [459c723](https://github.com/mne-tools/mne-bids/commit/459c723fa4a6947d050917b0f4ec2b95c5dc9c39) ?

`Optional[foo]` is just shorthand for `Union[foo, None]`.

In Python 3.10 we'll be able to simply write `foo | None` like in TypeScript, but … well, we're not quite there yet.

But your proposal in 459c723 is redundant; you can safely switch to `Optional[Path]` (more "idiomic") or `Union[Path, None]` (more explicit, but rather uncommon as far as I can tell)

sappelhoff:
Thanks for the hint, will push a commit!

hoechenberger:
> Thanks for the **hint**

☝️ I see what you did there! 😎

hoechenberger:
> we'd only need it for `root`, right? All other parts of the `BIDSPath` are strings or None anyway?
> How would we do that? Would we need a new class `_BIDSRoot` or something?

How about we always store `BIDSPath.root` as a string (`str(Path(root).expanduser())` or something) and whenever we need it to concat a path, we just cat it to a Path again?

sappelhoff:
> How about we always store BIDSPath.root as a string (str(Path(root).expanduser()) or something) and whenever we need it to concat a path, we just cat it to a Path again?

Can we store it as a `Path(root).as_posix()` (posix str) and then later on call Path on that posix_str and it's work on Windows? If yes, that seems like a good way to go

hoechenberger:
> Can we store it as a `Path(root).as_posix()` (posix str) and then later on call Path on that posix_str and it's work on Windows? If yes, that seems like a good way to go

Nice idea, I think this could work

sappelhoff:
Ok cool, I opened a new issue for it :-)

Please merge this one if you are happy @hoechenberger 

I could then do the release next.

hoechenberger (approved):

hoechenberger:
Thanks a bunch, @sappelhoff!

PR #837, opened 2021-07-08
MRG: Add update_anat_landmarks()

hoechenberger:
Allows updating the anatomical landmarks of an MR scan
without writing the entire data again.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/sidecar_updates.py
+++ b/mne_bids/sidecar_updates.py
@@ -227,7 +227,7 @@ def update_anat_landmarks(bids_path, landmarks):
     # Check if coordinates for any cardinal point are missing, and convert to
     # a list so we can easily store the data in JSON format
     missing_points = []
-    for name, coords in name_to_coords_map.copy().items():
+    for name, coords in name_to_coords_map.items():
         if coords is None:
             missing_points.append(name)
         else:
--- a/mne_bids/sidecar_updates.py
+++ b/mne_bids/sidecar_updates.py
@@ -141,14 +141,27 @@ def _update_sidecar(sidecar_fname, key, val):
 
 
 def update_anat_landmarks(bids_path, landmarks):
-    """[summary]
+    """Update the anatomical landmark coordinates of an MRI scan.
+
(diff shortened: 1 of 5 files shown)


mne_bids/sidecar_updates.py
```diff
@@ -136,3 +137,72 @@ def _update_sidecar(sidecar_fname, key, val):
...
+    name_to_coords_map = {
+        'LPA': positions['lpa'],
+        'NAS': positions['nasion'],
+        'RPA': positions['rpa']
+    }
+
+    # Check if coordinates for any cardinal point are missing, and convert to
+    # a list so we can easily store the data in JSON format
+    missing_points = []
+    for name, coords in name_to_coords_map.copy().items():
```
agramfort, line 185: why the copy here? I think the line below will modify still the original name_to_coords_map

hoechenberger: Because without the copy, inside the loop I'd modifying the object I'm iterating over, which sometimes has weird side effects. Not sure if the copy is necessary in this specific case here, though? WDYT?

agramfort: so you want name_to_coords_map to be modified outside of the function? if so I think the copy() is not necessary

hoechenberger: No, we're modifying `name_to_coords_map` here in this loop (in the `else` branch).
I'm not sure if this is "okay" to do while looping over `name_to_coords_map`?

hoechenberger: But maybe it doesn't matter as long as we're not deleting keys

agramfort:
yes

>

hoechenberger:
Rendered docs: https://4903-89170358-gh.circle-artifacts.com/0/dev/generated/mne_bids.update_anat_landmarks.html

agramfort (approved):

agramfort:
@hoechenberger merge when green !

Issue #863, opened 2021-08-24
Support exporting to EDF format


adam2392:
#### Describe the problem
EDF is a RECOMMENDED data format for EEG and iEEG, but we cannot export to it.

#### Describe your solution
The latest PR https://github.com/mne-tools/mne-python/pull/9643 enables exporting to EDF using EDFlib-Python, a fully compliant EDF specification Python writer maintained by the creator of EDF and EDFBrowser.

I will add the export option in `write_raw_bids`.

adam2392:
In order to support this, I also wonder if the team would be okay with me refactoring the tests to:

- test_write_raw.py  # test general writing stuff
- test_write_metadata.py. # test writing metadata
- test_write_format.py  # test converting format of raw files

? I just took a look at the tests I would need to enable, and they are quite messy on the write end :p.

cc: @sappelhoff @jasmainak @agramfort @hoechenberger

hoechenberger:
Refactoring of tests is highly welcomed by me 😅 we just have to be super careful not to accidentally break or disable some test cases...

PR #864, opened 2021-08-24
[MRG] Refactor writing tests

adam2392:
PR Description
--------------

First part of #863 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


mne_bids/tests/write/test_write.py
```diff
@@ -2543,137 +2536,6 @@ def test_sidecar_encoding(_bids_validate, tmpdir):
...
-    # if we accidentally add MEG channels, then an error will occur
-    raw.set_channel_types({raw.info['ch_names'][0]: 'mag'})
-    with pytest.raises(ValueError, match='Got file extension .*'
-                                         'for MEG data'):
-        write_raw_bids(**kwargs)
-
-
-@pytest.mark.parametrize('dir_name, fname, reader', test_convert_data)
-@pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
-def test_convert_dataset_format(dir_name, fname, reader, tmpdir):
```
adam2392, line 2615: These error checks and assertions could be moved into the existing tests actually.

agramfort:
to me the logic is the following. To test the functions in foo.py you add tests in tests/test_foo.py
so when you look for where a function is tested there is one obvious place to look. We never have
subfolders in tests folders AFAIK

adam2392:
> to me the logic is the following. To test the functions in foo.py you add tests in tests/test_foo.py
> so when you look for where a function is tested there is one obvious place to look. We never have
> subfolders in tests folders AFAIK

Should I get rid of the subfolder then? And just have those two files?

Or should I just close this PR and proceed with #863 ?

agramfort:
honestly splitting things in different files is not fundamental when you work with a good editor that allows you to jump between functions within a file.

what is the rationale ?

now i agree that we should maybe factorize some tests expescially with new formats support

PR #862, opened 2021-08-22
[MAINT] Require MNE-Python `main` until `0.23.1` release

alexrockhill:
I have no idea what's causing the CIs to fail here https://github.com/mne-tools/mne-bids/pull/859 and I assume it's not something to do with the PR but I have failed to replicate the issue locally so I'm just pushing an empty commit just to make sure.

PR #865, opened 2021-08-24
[maint] pin mne to 0.23.1

sappelhoff:
Maybe this won't work, I think what I wrote in https://github.com/mne-tools/mne-bids/pull/862#issuecomment-904664791 was confused stuff :)

We'll probably have to wait until 0.24 then. Let's check anyhow.


sappelhoff:
okay, I am fine with waiting until 0.24. It seems the only alternative is to add some logic to read.py and write.py to check the mne version before calling `mne.chpi.get_chpi_info`.

If somebody wants to do that here, it's fine by me --> else: Please feel free to close :-)

PR #866, opened 2021-08-25
[WIP] Enable write_raw_bids to format as EDF files

adam2392:
PR Description
--------------

Closes: #863
Closes: #864

- Refactors one unit test into two existing unit tests to improve readability
- Refactors current unit test of "converting to BrainVision" to just "converting EEG data in general"
- Adds EDF export support

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -24,7 +24,8 @@ import codecs
 from pkg_resources import parse_version
 
 import numpy as np
-from numpy.testing import assert_array_equal, assert_array_almost_equal
+from numpy.testing import (assert_allclose, assert_array_equal,
+                           assert_array_almost_equal)
 
 # This is here to handle mne-python <0.20
 import warnings
@@ -78,7 +79,9 @@ warning_str = dict(
     encountered_data_in='ignore:Encountered data in*.:RuntimeWarning:mne',
     edf_warning=r'ignore:^EDF\/EDF\+\/BDF files contain two fields .*'
                 r':RuntimeWarning:mne',
-    maxshield='ignore:.*Internal Active Shielding:RuntimeWarning:mne'
+    maxshield='ignore:.*Internal Active Shielding:RuntimeWarning:mne',
+    edfblocks='ignore:.*EDF format requires equal-length data '
+              'blocks:RuntimeWarning:mne'
 )
 
 
@@ -114,7 +117,8 @@ test_converteeg_data = [
     ('Persyst', 'BrainVision', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
     ('NihonKohden', 'BrainVision', 'MB0400FU.EEG', _read_raw_nihon),
     ('Persyst', 'EDF', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
-    ('NihonKohden', 'EDF', 'MB0400FU.EEG', _read_raw_nihon),
+    # XXX: Idk why this one doesn't work.
+    # ('NihonKohden', 'EDF', 'MB0400FU.EEG', _read_raw_nihon),
 ]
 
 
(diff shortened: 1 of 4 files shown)


adam2392:
@agramfort instead of refactoring the test into multiple files, I did what you suggested. I also added in the EDF export ability now, and added corresponding unit tests.

mne_bids/tests/test_write.py
```diff
@@ -110,9 +113,12 @@ def fn(fname, *args, **kwargs):
...
 
-test_convertbrainvision_data = [
-    ('Persyst', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
-    ('NihonKohden', 'MB0400FU.EEG', _read_raw_nihon)
+test_converteeg_data = [
+    ('Persyst', 'BrainVision', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
+    ('NihonKohden', 'BrainVision', 'MB0400FU.EEG', _read_raw_nihon),
+    ('Persyst', 'EDF', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
+    # XXX: Idk why this one doesn't work.
+    # ('NihonKohden', 'EDF', 'MB0400FU.EEG', _read_raw_nihon),
```
agramfort, line 121: did you find out?

agramfort (commented):
so far so good !

mne_bids/tests/test_write.py
adam2392: No, there's some weird electrodes perhaps in that file, or something fishy going on in mne reading perhaps. Not sure how to pin it down yet

adam2392:
```
>           assert_array_almost_equal(
                raw.get_data(), raw2.get_data()[:, :orig_len], decimal=6)
E           AssertionError: 
E           Arrays are not almost equal to 6 decimals
E           
E           Mismatched elements: 139199 / 139200 (100%)
E           Max absolute difference: 10.80262834
E           Max relative difference: 9.00052626
E            x: array([[-1.931642e-04, -2.970705e-04,  1.092772e-04, ...,  4.902326e-05,
E                   -6.357439e-05, -1.533205e-04],
E                  [ 2.416990e-04,  7.587873e-05,  3.805662e-04, ...,  1.666990e-04,...
E            y: array([[ 0.001253,  0.001234,  0.001271, ...,  0.001271,  0.001253,
E                    0.001253],
E                  [ 0.001289,  0.001271,  0.001308, ...,  0.001289,  0.001271,...
```

adam2392: Maybe it's something to do with:

1. Nihon reader
2. the Nihon dataset
3. EDF export not handling cals/range correctly? Although, I don't see why those are needed, cuz `get_data` should return data in the physical format, and then cal/range for EDF is just `1`.
4. Something else...

adam2392:
@agramfort so the error goes a bit deeper, but I fixed it. It raises two issues in my head.

- The Nihon dataset contains two channels, `$A1, $A2`, which are these non-electrode channels. I think ref? or misc? These are read in as EEG channels. I can PR into mne-python to add a 1 character fix. See below for code snippet.
- EDF 'auto' will set the min/max physical ranges based on channel groups. When say certain channels are in the "wrong group", then their actual physical ranges will differ by a lot. Say `$A1` has a range of (-3200, 3200), and then all actual EEG channels have a range of (-1, 1), then they will get hit with a divisor of 6400 physical range. This then makes all the values as we saw in the test look "exactly the same". 

#### Fix for EDF:
This begs the question. Is there a way we can control against this? This is undesired runtime behavior that is dependent on how cognizant the user was on setting channel types. I was thinking of augmenting the export EDF behavior to look at the approximate ranges of every channel in each channel type, and if they are more then 2 decimal places away, then raise a warning. 

For example, if you have 10 eeg channel types and 2 are actually emg, then those emg channels will forsure have a physical range an order of magnitude away from the eeg channels. You would want to set types accordingly when exporting, or be wary of that. 

#### Fix for Nihon:
```
# fix in nihon.py
def _map_ch_to_type(ch_name):
    ch_type_pattern = OrderedDict([
        ('stim', ('Mark',)), ('misc', ('DC', 'NA', 'Z', '$')), ('bio', ('X',))])
    for key, kinds in ch_type_pattern.items():
        if any(kind in ch_name for kind in kinds):
            return key
    return 'eeg'
```

Issue #867, opened 2021-09-01
`copyfile_*` fails in `write_raw_bids` when `overwrite=True` and `src==dest`


adam2392:
#### Describe the bug
I am trying to append annotations to the events.tsv file. I found that the easiest way to do so is to simply do: i) read_raw_bids, ii) append annotations to the read in raw object and then iii) write_raw_bids again.

However, if I'm writing to the exact same location, then the file actually is removed. I think this is perhaps hidden behavior of `shuttle.copyfile(src, dest)`, where if src and dest are the same, then the file is deleted.

#### Steps to reproduce
```
raw = read_raw_edf(<some_edf_file>)
bids_path = BIDSPath(root=root, subject='001', datatype='eeg')
output_bids_path = write_raw_bids(raw, bids_path)

raw = read_raw_bids(output_bids_path)
write_raw_bids(raw, output_bids_path)
```

#### Expected results
At the very least, an error should be raised, and the file shouldn't be deleted.

Otherwise, if the file already exists, then copy somewhere temporarily and then copy to the same place again? The con is that one writes twice now.

#### Actual results
File is deleted, and the program errors.

#### Additional information
〈Replace this text with information about your system. For example through
using MNE-Python and running `mne.sys_info()` and pasting the output here.〉

hoechenberger:
According to the Python docs:

> If src and dst specify the same file, SameFileError is raised.

So if the file disappears then this is not due to `copyfile`, but the deletion must occur somewhere else

sappelhoff:
I can't replicate this issue on mne-bids and mne-python `main`:

```python
import os
import mne
from mne_bids import BIDSPath, write_raw_bids, read_raw_bids


testing = mne.datasets.testing.data_path()
fin = os.path.join(testing, "EDF", "test_generator_2.edf")
root = os.path.join(os.path.expanduser("~"), "Desktop", "mne_bids_issue-867")
os.makedirs(root, exist_ok=True)
raw = mne.io.read_raw_edf(fin)
bids_path = BIDSPath(root=root, subject='001', datatype='eeg')
output_bids_path = write_raw_bids(raw, bids_path)

raw = read_raw_bids(output_bids_path)
write_raw_bids(raw, output_bids_path)
```

It raises a `FileExistsError`, as expected (see Richards comment above)

hoechenberger:
> It raises a `FileExistsError`, as expected (see Richards comment above)

[...]

PR #889, opened 2021-10-21
[MRG] Raise an error message when src==dest in write_raw_bids

adam2392:
PR Description
--------------

Closes: #867 

- Adds an error for src == dest during write_raw_bids
- Changes writing of EDF to use `_write_raw_edf`, since EDFlib-Python is a supported EDF writer now
- Adds a unit test to track this behavior

The reason we can't accommodate copying files is outlined in the issue.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


adam2392:
Ugh million tests to change cuz of write_edf...

adam2392:
Ugh million tests to change cuz of write_edf...

hoechenberger:
What's the status here?

sappelhoff (commented):
>1 Adds an error for src == dest during write_raw_bids
    2 Changes writing of EDF to use _write_raw_edf, since EDFlib-Python is a supported EDF writer now
    3 Adds a unit test to track this behavior


I think point 2 should be dropped from this PR and (if at all) be done in a separate PR. I say "if at all", because 

> I don't think you should switch to _write_edf. We want to just copy the files in order to manipulate the raw files themselves as little as possible :-)

(see https://github.com/mne-tools/mne-bids/pull/889#issuecomment-949330962 and https://github.com/mne-tools/mne-bids/pull/889#discussion_r734271335)

Regarding point 1 and 3: Yes, +1 to do this. Having an explicit (and covered) error for this scenario would be an improvement over the status quo. But that error doesn't need to be a big and advertised thing, and not framed as a "limitation of mne-bids" IMHO (see https://github.com/mne-tools/mne-bids/pull/889#discussion_r734271335)

All in all, I think this PR could be fixed with a small couple of lines to 

1. detect the copy issue scenario
2. raise a small error that we don't support that
3. add a test to trigger that error

sappelhoff (commented):
ready for review, not that I also bumped minimum requirements for mne to 1.0 -- wdyt?

sappelhoff:
@hoechenberger @adam2392 @agramfort

mne_bids/path.py
```diff
@@ -142,7 +142,7 @@ def _find_matched_empty_room(bids_path):
 
 
 class BIDSPath(object):
-    """A BIDS path object.
+    """A BIDSPath object.
```
hoechenberger, line 145: I don't agree with any of these changes from "BIDS path" to "BIDSPath". The idea of the original doctrings was to convey that we're talking about paths in the BIDS dataset. Now you've turned all of those into self-referential thingies, which I think isn't helpful. For example here, you explain a BIDSPath b saying that it's a BIDSPath, this is … circular.

I do agree that the docstring should get reworked, but I don't think these changes here are the way we should go.

I hope this doesn't sound too harsh :D

mne_bids/report/_report.py
```diff
@@ -315,7 +315,7 @@ def _summarize_sidecar_json(root, scans_fpaths):
 
             n_scans += 1
 
-            # convert to BIDS Path
+            # convert to BIDSPath
```
hoechenberger, line 318: these changes, OTOH, are good

mne_bids/write.py
```diff
@@ -1881,6 +1851,30 @@ def write_raw_bids(raw, bids_path, events_data=None, event_id=None,
                              f'Please use one of {CONVERT_FORMATS[datatype]} '
                              f'for {datatype} datatype.')
 
+    # raise error when trying to copy files (copyfile_*) into same location
+    # (src == dest, see https://github.com/mne-tools/mne-bids/issues/867)
+    if bids_path.fpath.exists() and not convert and \
+            bids_path.fpath.as_posix() == Path(raw_fname).as_posix():
+        raise RuntimeError(f'Desired output BIDSPath ("{bids_path.fpath}") is'
```
hoechenberger, line 1858: Should be a `FileExistsError`

I would also suggest to change indentation to the "black" style

mne_bids/write.py
```diff
@@ -1881,6 +1851,30 @@ def write_raw_bids(raw, bids_path, events_data=None, event_id=None,
...
+    if bids_path.fpath.exists():
+        if overwrite:
+            # Need to load data before removing its source
+            raw.load_data()
+            if bids_path.fpath.is_dir():
+                shutil.rmtree(bids_path.fpath)
+            else:
+                bids_path.fpath.unlink()
+        else:
+            raise FileExistsError(
```
hoechenberger, line 1874: why is this line never hit by tests? should get coverage IMHO

hoechenberger (changes requested):
all good, except for changes to the docstring of BIDSPath, and the type of an exception you're raising

mne_bids/path.py
hoechenberger: just for clarification, my comment above only applies to the changes in the docstring(s), not to the inline comments in the code or tests

mne_bids/write.py
sappelhoff: The line is tested somewhat here: https://github.com/mne-tools/mne-bids/blob/213c3789390dfacc285418d8f54c6575ca104e7f/mne_bids/tests/test_write.py#L1858-L1866

The reason it is never hit is because other files (like events.tsv) get written earlier than the neural data file in `write_raw_bids`, and these raise an identical error:

https://github.com/mne-tools/mne-bids/blob/213c3789390dfacc285418d8f54c6575ca104e7f/mne_bids/utils.py#L206-L207

hoechenberger: So … then we can just remove this `else` branch, no?

sappelhoff: I'd leave it here. People may manually remove all other files and **then** this line would reasonably trigger. it's just not covered in our tests, which I find okay (and which I did not introduce here) :)

hoechenberger: alright, let's leave it here then!

hoechenberger:
Thanks, @sappelhoff!

PR #896, opened 2021-11-02
MRG: Replace tmpdir with tmp_path fixture

cbrnr:
This PR replaces the deprecated pytest `tmpdir` fixture with `tmp_path`. Because the tests do not run on my local system (mainly because the MNE package doesn't contain any test data), this is a first try (blindly replacing all occurrences).

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,5 @@
 nilearn
+sphinx <4.3
 sphinx_gallery
 sphinx-copybutton
 pydata-sphinx-theme
--- a/mne_bids/commands/tests/test_cli.py
+++ b/mne_bids/commands/tests/test_cli.py
@@ -52,9 +52,9 @@ def check_usage(module, force_help=False):
         assert 'Usage: ' in out.stdout.getvalue()
 
 
-def test_raw_to_bids(tmpdir):
+def test_raw_to_bids(tmp_path):
     """Test mne_bids raw_to_bids."""
-    output_path = str(tmpdir)
+    output_path = str(tmp_path)
     data_path = testing.data_path()
     raw_fname = op.join(data_path, 'MEG', 'sample',
                         'sample_audvis_trunc_raw.fif')
@@ -87,9 +87,9 @@ def test_raw_to_bids(tmpdir):
             mne_bids_cp.run()
 
 
-def test_cp(tmpdir):
+def test_cp(tmp_path):
     """Test mne_bids cp."""
-    output_path = str(tmpdir)
+    output_path = str(tmp_path)
     data_path = op.join(base_path, 'brainvision', 'tests', 'data')
     raw_fname = op.join(data_path, 'test.vhdr')
     outname = op.join(output_path, 'test2.vhdr')
@@ -107,14 +107,14 @@ def test_cp(tmpdir):
             mne_bids_cp.run()
 
 
-def test_mark_bad_chanels_single_file(tmpdir):
+def test_mark_bad_chanels_single_file(tmp_path):
     """Test mne_bids mark_channels."""
 
     # Check that help is printed
     check_usage(mne_bids_mark_channels)
 
     # Create test dataset.
-    output_path = str(tmpdir)
+    output_path = str(tmp_path)
     data_path = testing.data_path()
     raw_fname = op.join(data_path, 'MEG', 'sample',
                         'sample_audvis_trunc_raw.fif')
@@ -160,14 +160,14 @@ def test_mark_bad_chanels_single_file(tmpdir):
     assert raw.info['bads'] == []
 
 
-def test_mark_bad_chanels_multiple_files(tmpdir):
+def test_mark_bad_chanels_multiple_files(tmp_path):
     """Test mne_bids mark_channels."""
 
     # Check that help is printed
     check_usage(mne_bids_mark_channels)
 
     # Create test dataset.
-    output_path = str(tmpdir)
+    output_path = str(tmp_path)
     data_path = testing.data_path()
     raw_fname = op.join(data_path, 'MEG', 'sample',
                         'sample_audvis_trunc_raw.fif')
@@ -202,13 +202,13 @@ def test_mark_bad_chanels_multiple_files(tmpdir):
         assert set(old_bads + ch_names) == set(raw.info['bads'])
 
 
-def test_calibration_to_bids(tmpdir):
+def test_calibration_to_bids(tmp_path):
     """Test mne_bids calibration_to_bids."""
 
     # Check that help is printed
     check_usage(mne_bids_calibration_to_bids)
 
-    output_path = str(tmpdir)
+    output_path = str(tmp_path)
     data_path = Path(testing.data_path())
     fine_cal_fname = data_path / 'SSS' / 'sss_cal_mgh.dat'
     bids_path = BIDSPath(subject=subject_id, root=output_path)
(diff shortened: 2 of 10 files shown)


cbrnr:
At some point (i.e. in another PR) we should also replace `os.path` with `pathlib`...

cbrnr:
@sappelhoff @hoechenberger is it possible to still use a one-liner in this pattern? Unfortunately, `Path.mkdir()` returns `None` it seems...

Old:
```
bids_root = tmp_path.mkdir("bids")
```

New:
```
bids_root = tmp_path / "bids"
bids_root.mkdir()
```

cbrnr:
```python
(bids_root := tmp_path / "bids").mkdir()
```

👀  - any objections? Of course this requires Python 3.8+...

sappelhoff:
haha wow, that looks like making the most out of new features :laughing: Personally, I'd prefer two lines and keeping 3.7 compatibility. But I appreciate that new language feature, and if others like it better than the two-liner, I am happy to be overruled.

cbrnr:
We can always change to `:=` once we drop support for 3.7 (MNE 0.25?).

agramfort:
yes it's cool but let me play the grumpy one here (one more time)

I often say in my talks on software that should "write dumb code". Why?
Not everyone is a Python guru in our field and I want to maximize our
chances to welcome new contributors. All advanced code patterns
that are not absolutely necessary makes our projects less attractive
to new comers.

cbrnr:
That's a valid point, and in general I agree. I just saw a meme on Twitter that went something like this:

- Beginner: `if`/`elif`/`else`
- Junior: `switch`
- Experienced: state machine
- Senior: `if`/`elif`/`else`

However, I think this needs to be decided case by case, otherwise we'll be stuck forever writing Python 1.0 code. For example, f-strings were a really nice addition that made things easier for users in most cases. The assignment operator is definitely more restricted to specific use cases (but this use case was actually one of the very few ones for me where I think it made the code simpler).

In any case, I'm fine with the 2-line variant here (especially since we still support 3.7).

cbrnr:
> Yes, my personally feeling re the assignment expressions is that they REALLY make sense in for and while loops, but not sooo much in this particular case here :)

I really like that one here as well. Once again I wish Python was more like Julia, where an assignment is always an expression...

agramfort:
I can live with 3 lines :)

sappelhoff:
> I can live with 3 lines :)

me too :-)

+1 for walrus operator when we drop Python 3.7 support in the future (when it's end of life, I guess)

agramfort:
I'll stick to 3 lines and if/then/else as a Beginner or Senior developer :)

>

hoechenberger:
I'm all for the walrus but with the specific case here I think it makes things slightly more difficult to read. I agree with Clemens that it's a pity that Path.mkdir() doesn't return self. ☠️

hoechenberger:
> > I can live with 3 lines :)
> 
> 
> 
> me too :-)
> 
> 
> 
> +1 for walrus operator when we drop Python 3.7 support in the future (when it's end of life, I guess)

Is actuallynopt for dropping it next year, so that we always support the latest 3 versions. To keep CI resource demands manageable if nothing else ...

sappelhoff:
> Because the tests do not run on my local system (mainly because the MNE package doesn't contain any test data)

Couldn't you just clone MNE-Python and in your mne-bids environment do a local `pip -e .` install of the MNE-Python clone?

cbrnr:
> Couldn't you just clone MNE-Python and in your mne-bids environment do a local pip -e . install of the MNE-Python clone?

I don't know, shouldn't the tests also work with a pip installed mne? It really looks like the pip package doesn't contain the test datasets.

sappelhoff:
> I don't know, shouldn't the tests also work with a pip installed mne? It really looks like the pip package doesn't contain the test datasets.

Yes, MNE via pip install does not include the test data as far as I can tell: https://github.com/mne-tools/mne-python/blob/5bafe5287dd27ef6e07a64609615326110536578/MANIFEST.in#L75-L84

That's been this way since mne-bids was started, so we always relied on a local clone + local install:

https://github.com/mne-tools/mne-bids/blob/main/CONTRIBUTING.md#clone-mne-python-and-install-it-from-the-git-repository


Whether that's a good thing or not is another question :)

cbrnr:
Thanks for clarifying @sappelhoff. I think it is confusing that a git clone behaves differently than the pip package. I did ask myself why a git clone of the current main branch was necessary, but thought that I could just use the pip package instead.

hoechenberger:
> I did ask myself why a git clone of the current main branch was necessary, but thought that I could just use the pip package instead.

Oh wow, mind = blown 🤯 

I never realized this before either

cbrnr:
So I guess this pretty much answers @sappelhoff's previous question whether this was a good thing... 😄 

Maybe all tests can be changed to use data from the `mne-testing-data` repo (that can be downloaded with `mne.datasets.testing.data_path()`?

hoechenberger:
I don't think it's necessary, since the development version of MNE-BIDS typically requires the development version of MNE-Python, which cannot be installed from PyPI…

sappelhoff:
> Maybe all tests can be changed to use data from the mne-testing-data repo (that can be downloaded with mne.datasets.testing.data_path()?

or MNE-Python starts to ship the test data :)

IMHO we don't need to fix it - it's more effort than it's worth

cbrnr:
True. If at all MNE should include the test data.

agramfort:
We cannot ship them. It’s too big for pypi. I tried ;)

hoechenberger:
@cbrnr What's the status here? Need a hand to get this finished?

cbrnr:
That would be great!

hoechenberger:
@cbrnr Could you give me a quick reminder what remains to be done?

cbrnr:
Basically get all tests to pass. I've searched and replaced, but now e.g. `tmpdir.mkdir()` needs to be replaced with the two/three lines that we discussed in another issue (where I wanted to use the assignment operator).

hoechenberger:
@cbrnr Turns out we didn't need those `mkdir()`s in most cases. So this debate was a little misguided 😅 But it was fun to discuss anyway haha

cbrnr:
@hoechenberger this was a discussion in the `mne-python` repo – how is `mkdir()` not necessary?

hoechenberger:
@cbrnr In many of the tests we created BIDS root dirs manually, but `write_raw_bids()` creates them automatically for us, so I could drop loads of `mkdir()` calls

cbrnr:
OK, so it's specific to the code in this repository and not a general optimization you could also apply to the code in `mne-python`.

hoechenberger:
@cbrnr Yep, sorry for the confusion! :)

hoechenberger:
I'm not sure why the doc build is failing..?

cbrnr:
Seems unrelated, maybe restart?

hoechenberger:
@cbrnr I've restarted Circle, let's see …

hoechenberger:
Now it's a different error … I'm a bit clueless. Seems like a problem with `sphinx-gallery`?

@larsoner Any idea what's up with our documentation build here?

https://circleci.com/gh/mne-tools/mne-bids/5501

hoechenberger:
Seems there was a new Sphinx release a few hours ago (4.3.0); let's see if pinning to an older version helps.

larsoner:
Always check MNE-Python first :)

https://github.com/mne-tools/mne-python/pull/9992

hoechenberger:
You telling me that you pushed a fix to MNE-Python 2 hours ago, while the problematic new Sphinx release was made just one hour before that?? 

**DAMN** 🤩🥸🥳😎❤️

cbrnr:
I wouldn't be surprised if @larsoner also made the Sphinx release himself 😂.

larsoner:
Technically I pushed the fix to SciPy first, then MNE-Python

larsoner:
(And really this is partially my fault because I had seen the error on Sphinx 5.0.dev0 some weeks ago but didn't take the time to open an upstream issue with pydata-sphinx-theme :( )

hoechenberger:
Alright, we're green.

agramfort (approved):

agramfort:
thx @cbrnr !

cbrnr:
Thanks @hoechenberger!

sappelhoff:
Thanks all! :rocket:

PR #899, opened 2021-11-10
Confine get_entity_vals() to /sub-* directories

adam2392:
PR Description
--------------

`sourcedata/` should not be checked for entity_vals when getting a dataset. Quick 1 liner fix...

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -87,6 +87,8 @@ Bug fixes
 
 - Fix erroneous measurement date check in :func:`mne_bids.write_raw_bids` when requesting to anonymize empty-room data, by `Richard Höchenberger`_ (:gh:`893`)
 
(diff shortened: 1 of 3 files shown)


hoechenberger:
Actually the best solution probably would be to only check `/sub-*`, right?

hoechenberger:
So I would do sth like:
```python
    filenames = (Path(root)
                  .glob(f'sub-*/**/*{entity_long_abbr_map[entity_key]}-*_*'))
```
WDYT?

hoechenberger:
@adam2392 I've pushed a commit to implement the logic I had proposed, LMK what you thing

adam2392:
LGTM!

hoechenberger:
@adam2392 can you check whether it works for your use case? :) I will then add a changelog entry and we can merge

hoechenberger (approved):

hoechenberger:
Thanks @adam2392!

Issue #895, opened 2021-11-01
Simplify storage of empty-room data


hoechenberger:
Currently, storing empty-room data is unnecessarily complicated and error-prone.

Current approach:

- manually create a `BIDSPath` with `subject='emptyroom'`, `task='noise'`, and `session='YYYYMMDD'` (meaning one first has to manually construct the date/session string)
- pass this `BIDSPath` and the empty-room raw to `write_raw_bids`

Problems:

- `write_raw_bids` special-handles the `emptyroom` subject
- this is not documented anywhere
- if there's a typo in the subject name, the empty-room-special-handling branches of code will be omitted and the user won't be notified, as empty-room data will be treated like regular experimental data
- inconsistent behavior when the user requests anonymization: suddenly, `write_raw_bids` starts fiddling with `bids_path.session` to ensure that `daysback` is applied there as well, meaning that the data is written to a session directory that differs from what the user explicitly passed
- before we anonymize, we demand that empty-room session and the date in `raw.info` are the same and raise an exception otherwise; again, this behavior is not documented

Proposal:

- Add a kwarg `is_emptyroom` to `write_raw_bids`; if `True`, automatically set `subject`, `task`, `session` (and ignore values potentially present in `bids_path`)

Problem with proposal:

- We already have a kwarg `emptyroom`, which could lead to confusion. We could rename it to `emptyroom_bids_path` or `er_bids_path`, though?

Alternatives:

- Add a dedicated `write_raw_bids_emptyroom` function. Seems overkill?

agramfort:
can you paste how the user code  looks now and how it would look like with
your proposal?

>

hoechenberger:
Current approach (important bits indicated by the pointing fingers):
```python
from pathlib import Path

import mne
from mne.datasets import sample

from mne_bids import (write_raw_bids, BIDSPath)

data_path = Path(sample.data_path())
raw_fname = data_path / 'MEG' / 'sample' / 'sample_audvis_raw.fif'
er_raw_fname = data_path / 'MEG' / 'sample' / 'ernoise_raw.fif'
bids_root = data_path.parent /  'MNE-sample-data-bids'

raw = mne.io.read_raw_fif(raw_fname)
raw_er = mne.io.read_raw_fif(er_raw_fname)
raw.info['line_freq'] = 60
raw_er.info['line_freq'] = 60


bids_path = BIDSPath(
    subject='01',
    task='audiovisual',
    root=bids_root,
    suffix='meg',
[...]

PR #902, opened 2021-11-11
MRG: Rename empty_room param to er_bids_path in write_raw_bids()

hoechenberger:
The is a preparation to make resolving #895 easier. Also I believe the new parameter name is now more explicit, which is good.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/examples/anonymize_dataset.py
+++ b/examples/anonymize_dataset.py
@@ -76,7 +76,7 @@ write_raw_bids(raw=raw_er, bids_path=bids_path_er, verbose=False)
 # Write experimental MEG data, fine-calibration and crosstalk files
 write_raw_bids(
     raw=raw, bids_path=bids_path, events_data=events_path, event_id=event_id,
-    empty_room=bids_path_er, verbose=False
+    er_bids_path=bids_path_er, verbose=False
 )
 write_meg_calibration(cal_path, bids_path=bids_path, verbose=False)
 write_meg_crosstalk(ct_path, bids_path=bids_path, verbose=False)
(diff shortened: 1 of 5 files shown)


mne_bids/write.py
```diff
@@ -1181,6 +1182,11 @@ def write_raw_bids(raw, bids_path, events_data=None, event_id=None,
            add support for Windows 10 at a later time.
 
     empty_room : mne_bids.BIDSPath | None
+
+        .. deprecated::
+           This parameter has been deprecated in favor of ``er_bids_path`` and
+           will be removed in a future version of MNE-Python.
```
hoechenberger, line 1188:
```suggestion
           will be removed in a future version of MNE-BIDS.
```

examples/anonymize_dataset.py
```diff
@@ -76,7 +76,7 @@
 # Write experimental MEG data, fine-calibration and crosstalk files
 write_raw_bids(
     raw=raw, bids_path=bids_path, events_data=events_path, event_id=event_id,
-    empty_room=bids_path_er, verbose=False
+    er_bids_path=bids_path_er, verbose=False
```
agramfort, line 79: er_bids_path=bids_path_er

hum... one of the 2 needs to be changed so it's consistent

agramfort (commented):
but honestly I am not convinced...

er_ is very cryptic in a parameter name and we break backward compat
unless people passed named parameters which was never enforced

so I have strong doubts on this one...

sappelhoff (commented):
I don't like the `er_bids_path` too much either, I agree with @agramfort that `er` is too cryptic.

`empty_room_bids_path` is too long, but I'd find it better than `er_bids_path`.

Is this all in preparation to add a `empty_room` boolean flag in a future release for #895 ? If you are not sure yet how you are going to solve that (or somebody else), I am not sure that we should prepare something right before a release. IMHO it can wait and then we can iterate after 0.9 and before 0.10

hoechenberger:
Fair enough, let's reconsider this sometime later.

PR #906, opened 2021-11-12
[MRG] Force all written files to be lower case EDF

adam2392:
PR Description
--------------

Fixes clarification from upstream in BIDS. https://github.com/bids-standard/bids-specification/pull/927

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -422,7 +422,7 @@ def copyfile_edf(src, dest, anonymize=None):
     # Ensure source & destination extensions are the same
     fname_src, ext_src = _parse_ext(src)
     fname_dest, ext_dest = _parse_ext(dest)
-    if ext_src != ext_dest:
+    if ext_src != ext_dest and ext_src != '.EDF':
         raise ValueError(f'Need to move data with same extension, '
                          f' but got "{ext_src}" and "{ext_dest}"')
 
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -439,7 +439,7 @@ def copyfile_edf(src, dest, anonymize=None):
 
     # Anonymize EDF/BDF data, if requested
     if anonymize is not None:
-        if ext_src == ['.bdf', '.BDF']:
+        if ext_src in ['.bdf', '.BDF']:
             raw = read_raw_bdf(dest, preload=False, verbose=0)
         elif ext_src in ['.edf', '.EDF']:
             raw = read_raw_edf(dest, preload=False, verbose=0)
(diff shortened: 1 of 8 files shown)


hoechenberger:
We should still support READING files with uppercase extensions, otherwise we might break functionality with datasets that users created with MNE-BIDS in the past.

hoechenberger (changes requested):
We should still support READING files with uppercase extensions, otherwise we might break functionality with datasets that users created with MNE-BIDS in the past.

sappelhoff (commented):
I agree with Richard, but other than that, +1

mne_bids/copyfiles.py
```diff
@@ -422,7 +422,7 @@ def copyfile_edf(src, dest, anonymize=None):
     # Ensure source & destination extensions are the same
     fname_src, ext_src = _parse_ext(src)
     fname_dest, ext_dest = _parse_ext(dest)
-    if ext_src != ext_dest:
+    if ext_src != ext_dest and ext_src != '.EDF':
```
sappelhoff, line 425: if the original extension is different from the new extension: raise an error

why the additional check to **only raise if the original extension is not EDF**?

if it makes sense (please tell me), why are you not also covering BDF?

I think it'd be nicer if we tweak copyfile to always write EDF files with lower-case `.edf`: Feed in `.EDF`, get a warning and a `.edf` :-)

here you want to prevent users from calling `copyfile_edf` on data with an EDF extension, right?

mne_bids/write.py
```diff
@@ -1264,6 +1264,11 @@ def write_raw_bids(raw, bids_path, events_data=None, event_id=None,
     If you need to add more data there, or overwrite it, then you should
     call :func:`mne_bids.make_dataset_description` directly.
 
+    When writing EDF files, all EDF file extensions are forced to be
+    lower-case incompliance with the BIDS specification. Although EDF
+    supports both upper and lower case extension (i.e. '.EDF' and '.edf'),
+    BIDS requires '.edf'.
```
sappelhoff, line 1270:
```suggestion
    When writing EDF or BDF files, all file extensions are forced to be
    lower-case, in compliance with the BIDS specification.
```

mne_bids/copyfiles.py
adam2392: Yeah makes sense. I did it the way you said.

adam2392: Lmk if this looks good?

hoechenberger:
Let's hold off until it's actually part of the standard. The upstream PR hasn't been merged yet.

mne_bids/copyfiles.py
sappelhoff: yes, that's better -thank you

hoechenberger (approved):

hoechenberger:
Thanks @adam2392!

PR #914, opened 2021-11-17
MRG: When anonymizing, transfer trigger codes

hoechenberger:
Previously, anonymization would cause a new assignment of trigger codes, which could lead to confusion and inconsistencies across participants – or even within participants across runs (different values for the same event types)

We now simply overwrite the trigger codes in the anonymized `*_events.tsv` files with the original codes.

This problem is directly related to #223, and the solution implemented here is merely a workaround (which is good enough for now)

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


agramfort (approved):

agramfort:
thanks a ton @hoechenberger 🙏

Issue #929, opened 2021-12-11
When converting to EDF with _write_raw_edf, overwrite parameter should be passed


adam2392:
#### Describe the bug
Overwrite is not passed to _write_raw_edf, so if you do `write_raw_bids(..., overwrite=True, format='EDF')`, and the file path already exists, then it will result in an error.

#### Steps to reproduce
See PR for unit test.

adam2392:
Related to #867 ... slightly.

PR #930, opened 2021-12-11
[BUG] Add overwrite to export_raw for EDF conversions

adam2392:
PR Description
--------------

Fixes: #929 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -2858,7 +2858,8 @@ def test_convert_eeg_formats(dir_name, format, fname, reader, tmp_path):
     'dir_name, format, fname, reader', test_converteeg_data)
 @pytest.mark.filterwarnings(
     warning_str['channel_unit_changed'], warning_str['edfblocks'])
-def test_format_conversion_overwrite(dir_name, format, fname, reader, tmp_path):
+def test_format_conversion_overwrite(dir_name, format, fname, reader,
+                                     tmp_path):
     """Test that overwrite works when format is passed to write_raw_bids."""
     bids_root = tmp_path / format
     data_path = op.join(testing.data_path(), dir_name)
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -2851,6 +2851,50 @@ def test_convert_eeg_formats(dir_name, format, fname, reader, tmp_path):
         raw.get_data(), raw2.get_data()[:, :orig_len], decimal=6)
 
 
+@requires_version('mne', '0.24')
+@requires_version('pybv', '0.6')
+@pytest.mark.parametrize(
+    'dir_name, format, fname, reader', test_converteeg_data)
+@pytest.mark.filterwarnings(
+    warning_str['channel_unit_changed'], warning_str['edfblocks'])
+def test_convert_multiple(dir_name, format, fname, reader, tmp_path):
+    bids_root = tmp_path / format
+    data_path = op.join(testing.data_path(), dir_name)
+    raw_fname = op.join(data_path, fname)
+
+    # the BIDS path for test datasets to get written to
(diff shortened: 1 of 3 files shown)


mne_bids/tests/test_write.py
```diff
@@ -2851,6 +2851,50 @@ def test_convert_eeg_formats(dir_name, format, fname, reader, tmp_path):
...
+        else:
+            with pytest.warns(RuntimeWarning,
+                              match='Encountered data in "double" format'):
+                write_raw_bids(**kwargs)
+
+                # should work when writing the second time
+                write_raw_bids(**kwargs)
+    else:
+        with pytest.warns(RuntimeWarning,
+                          match='Converting data files to EDF format'):
```
adam2392, line 2891: These if/else statements are annoying to catch... Is there a way we can just ignore all warnings in a unit test?

hoechenberger: Yes, but we want to be as specific as we can be. So I think we'll have to live with this…

But these lines after `# should work when writing the second time` should be dedented one level, no?!

adam2392: Good point. 

I kept them within the pytest warning catch because I think those warnings will still fire off no matter what.

hoechenberger: Then I don't understand why we're even calling the function twice 🤔

sappelhoff (approved):
+1 to merge when CIs are green.

Thanks Adam!

hoechenberger (changes requested):
I'll want have a look at the tests before we merge, will get back to you shortly.

mne_bids/tests/test_write.py
```diff
@@ -2851,6 +2851,50 @@ def test_convert_eeg_formats(dir_name, format, fname, reader, tmp_path):
...
+                  verbose=False)
+
+    if format in ['BrainVision', 'auto']:
+        if dir_name == 'NihonKohden':
+            with pytest.warns(RuntimeWarning,
+                              match='Encountered data in "short" format'):
+                write_raw_bids(**kwargs)
+
+                # should work when writing the second time
+                write_raw_bids(**kwargs)
```
hoechenberger, line 2881: Can you explain what that is supposed to achieve?

I dedented these lines and then discovered the same warning is still being thrown. So I'm wondering why you're even testing this?

mne_bids/tests/test_write.py
```diff
@@ -2851,6 +2851,50 @@ def test_convert_eeg_formats(dir_name, format, fname, reader, tmp_path):
...
 
 
+@requires_version('mne', '0.24')
+@requires_version('pybv', '0.6')
+@pytest.mark.parametrize(
+    'dir_name, format, fname, reader', test_converteeg_data)
+@pytest.mark.filterwarnings(
+    warning_str['channel_unit_changed'], warning_str['edfblocks'])
+def test_convert_multiple(dir_name, format, fname, reader, tmp_path):
+    bids_root = tmp_path / format
```
hoechenberger, line 2861: Can you add a tiny docstring here, explaining what is being tested?

mne_bids/tests/test_write.py
adam2392: The first write_raw_bids gets the files there. The second one tests that overwrite=True works. If you remove the changed lines in `_export_edf()`, then this test will fail.

I'll add some details to the doc string

mne_bids/tests/test_write.py
adam2392: Done.

agramfort (approved):

agramfort:
@hoechenberger merge if happy

🙏 @adam2392

hoechenberger (approved):
Pushed a tiny change, good to merge once green

hoechenberger:
Thanks @adam2392!

PR #950, opened 2022-01-30
FIX: make update unalter the path instance if check fails

### LINKED ISSUE #949: [BUG] BIDSpath.update changes suffix option even if suffix isn't supported
#### Describe the bug
When using the BIDSPath option, if you try to update the suffix with a non-supported suffix (e.g. `_ave`), you will receive an error that the suffix is not supported (intended behavior) but the BIDSPath option is updated regardless (not intended behavior). I believe this is because the `_check` method is called after setting the suffix instead of being called first.

#### Steps to reproduce
```python
import mne_bids


bids_path = mne_bids.BIDSPath(subject='test')
bids_path.update(suffix='ave')
bids_path
```

#### Expected results
BIDSPath object should not update with the unsupported extension or there should be an ignore option so that the error doesn't occur.

#### Actual results
```python
In [6]: bids_path.update(suffix='ave')  
ValueError: Suffix ave is not allowed. Use one of these suffixes ['meg', 'markers', 'eeg', 'ieeg', 'T1w', 'FLASH', 'participants', 'scans', 'electrodes', 'channels', 'coordsystem', 'events', 'headshape', 'digitizer', 'beh', 'physio', 'stim'].

In [7]: bids_path                                                               
Out[7]: 
BIDSPath(
root: None
datatype: None
basename: sub-test_ave)
```

agramfort:
closes #949

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,12 +11,12 @@ jobs:
         # restore cache from last build. Unless __init__.py has changed since then
         - restore_cache:
             keys:
-              - data-cache-4-{{ checksum "./mne_bids/__init__.py" }}
+              - data-cache-5-{{ checksum "./mne_bids/__init__.py" }}
 
         # Also restore pip cache to speed up installations
         - restore_cache:
             keys:
-              - pip-cache-0-{{ checksum "./test_requirements.txt" }}-{{ checksum "./doc/requirements.txt" }}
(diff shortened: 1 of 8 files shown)


agramfort:
@teonbrooks let me know if it fixes your pb

hoechenberger (approved):
@agramfort Can you cook up a changelog entry for this one? Otherwise LGTM

sappelhoff (approved):

hoechenberger:
Thanks @agramfort!

teonbrooks:
wfm, thanks @agramfort!

PR #951, opened 2022-01-30
bump cache, install with deps

sappelhoff:
See:

- https://github.com/mne-tools/mne-python/pull/10261#issuecomment-1025174845


sappelhoff:
probably the fix here is rather than installing mne[hdf5], to update mne-bids dependencies to optionally include pymatreader when EEGLAB operations are wanted, WDYT?

cc @hoechenberger

agramfort (approved):
nice catch !

agramfort:
@hoechenberger I let you merge if happy

hoechenberger:
Thanks @sappelhoff!

PR #955, opened 2022-02-07
ENH : add support for landmarks as dict in write_anat and kind param …

agramfort:
…in get_head_mri_trans

this is to be able to have landmarks eg for different sessions for coregistration

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,7 +40,7 @@ Enhancements
 
 - Add support for CNT (Neuroscan) files in :func:`mne_bids.write_raw_bids`, by `Yorguin Mantilla`_ (:gh:`924`)
 
-- Add the ability to write multiple landmarks with :func:`mne_bids.write_anat` (e.g. to have different landmarks per session) by `Alexandre Gramfort`_ (:gh:`955`)
+- Add the ability to write multiple landmarks with :func:`mne_bids.write_anat` (e.g. to have separate landmarks for different sessions) by `Alexandre Gramfort`_ (:gh:`955`)
 
 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,6 +40,8 @@ Enhancements
 
 - Add support for CNT (Neuroscan) files in :func:`mne_bids.write_raw_bids`, by `Yorguin Mantilla`_ (:gh:`924`)
 
+- Add the ability to write multiple landmarks with :func:`mne_bids.write_anat` (e.g. to have different landmarks per session) by `Alexandre Gramfort`_ (:gh:`955`)
+
 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -761,7 +761,7 @@ def read_raw_bids(bids_path, extra_params=None, verbose=None):
 
 @verbose
 def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
-                       fs_subject=None, fs_subjects_dir=None, kind="",
+                       fs_subject=None, fs_subjects_dir=None, *, kind="",
                        verbose=None):
     """Produce transformation matrix from MEG and MRI landmark points.
 
@@ -799,10 +799,14 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
 
         .. versionadded:: 0.8
     kind : str | None
-        The suffix of the fiducials names to be used. If None, no suffix
-        is appended and one uses the landmarks called Nasion/NAS, LPA, and RPA.
-        The kind should be without the ``_`` inserted between landmark name
-        and the suffix kind.
+        The suffix of the anatomical landmark names in the JSON sidecar.
+        A suffix might be present e.g. to distinguish landmarks between
+        sessions. If provided, should not include a leading underscore ``_``.
+        For example, if the landmark names in the JSON sidecar file are
+        ``LPA_ses-1``, ``RPA_ses-1``, ``NAS_ses-1``, you should pass
+        ``'ses-1'`` here.
+        If ``None``, no suffix is appended, the landmarks named
+        ``Nasion`` (or ``NAS``), ``LPA``, and ``RPA`` will be used.
 
         .. versionadded:: 0.10
     %(verbose)s
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -761,7 +761,8 @@ def read_raw_bids(bids_path, extra_params=None, verbose=None):
 
 @verbose
 def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
-                       fs_subject=None, fs_subjects_dir=None, verbose=None):
+                       fs_subject=None, fs_subjects_dir=None, kind="",
+                       verbose=None):
     """Produce transformation matrix from MEG and MRI landmark points.
 
     Will attempt to read the landmarks of Nasion, LPA, and RPA from the sidecar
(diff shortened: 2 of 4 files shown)


mne_bids/read.py
```diff
@@ -761,7 +761,8 @@ def read_raw_bids(bids_path, extra_params=None, verbose=None):
 
 @verbose
 def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
-                       fs_subject=None, fs_subjects_dir=None, verbose=None):
+                       fs_subject=None, fs_subjects_dir=None, kind="",
```
sappelhoff, line 764: can we make all future params keyword only?

```suggestion
                       fs_subject=None, fs_subjects_dir=None, *, kind="",
```

sappelhoff (commented):
LGTM in principle, and seems okay from a BIDS point of view (https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/02-magnetoencephalography.html).

I am missing a changelog entry, and maybe (optional) some example.

doc/whats_new.rst
```diff
@@ -40,6 +40,8 @@ Enhancements
 
 - Add support for CNT (Neuroscan) files in :func:`mne_bids.write_raw_bids`, by `Yorguin Mantilla`_ (:gh:`924`)
 
+- Add the ability to write multiple landmarks with :func:`mne_bids.write_anat` (e.g. to have different landmarks per session) by `Alexandre Gramfort`_ (:gh:`955`)
```
hoechenberger, line 43:
```suggestion
- Add the ability to write multiple landmarks with :func:`mne_bids.write_anat` (e.g. to have separate landmarks for different sessions) by `Alexandre Gramfort`_ (:gh:`955`)
```

mne_bids/read.py
```diff
@@ -797,6 +798,13 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
         ``SUBJECTS_DIR`` environment variable.
 
         .. versionadded:: 0.8
+    kind : str | None
+        The suffix of the fiducials names to be used. If None, no suffix
+        is appended and one uses the landmarks called Nasion/NAS, LPA, and RPA.
+        The kind should be without the ``_`` inserted between landmark name
+        and the suffix kind.
+
+        .. versionadded:: 0.10
```
hoechenberger, line 807:
```suggestion
    kind : str | None
        The suffix of the anatomical landmark names in the JSON sidecar.
        A suffix might be present e.g. to distinguish landmarks between
        sessions. If provided, should not include a leading underscore ``_``.
        For example, if the landmark names in the JSON sidecar file are
        ``LPA_ses-1``, ``RPA_ses-1``, ``NAS_ses-1``, you should pass
        ``'ses-1'`` here.
        If ``None``, no suffix is appended, the landmarks named
        ``Nasion`` (or ``NAS``), ``LPA``, and ``RPA`` will be used.

        .. versionadded:: 0.10
```

mne_bids/read.py
```diff
@@ -863,14 +871,15 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
     mri_coords_dict = t1w_json.get('AnatomicalLandmarkCoordinates', dict())
 
     # landmarks array: rows: [LPA, NAS, RPA]; columns: [x, y, z]
+    suffix = f"_{kind}" if kind else ""
     mri_landmarks = np.full((3, 3), np.nan)
     for landmark_name, coords in mri_coords_dict.items():
-        if landmark_name.upper() == 'LPA':
+        if landmark_name.upper() == ('LPA' + suffix).upper():
```
hoechenberger, line 877: Shouldn't we be case-sensitive when it comes to the suffix?

```suggestion
        if landmark_name.upper() == 'LPA' + suffix
```

mne_bids/read.py
```diff
@@ -863,14 +871,15 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
...
             mri_landmarks[0, :] = coords
-        elif landmark_name.upper() == 'RPA':
+        elif landmark_name.upper() == ('RPA' + suffix).upper():
```
hoechenberger, line 879:
```suggestion
        elif landmark_name.upper() == 'RPA' + suffix
```

mne_bids/read.py
```diff
@@ -863,14 +871,15 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
...
             mri_landmarks[2, :] = coords
-        elif (landmark_name.upper() == 'NAS' or
-              landmark_name.lower() == 'nasion'):
+        elif (landmark_name.upper() == ('NAS' + suffix).upper() or
+              landmark_name.lower() == ('nasion' + suffix).lower()):
```
hoechenberger, line 882:
```suggestion
        elif (landmark_name.upper() == 'NAS' + suffix or
              landmark_name.lower() == 'nasion' + suffix):
```

mne_bids/write.py
```diff
@@ -1774,6 +1774,42 @@ def get_anat_landmarks(image, info, trans, fs_subject, fs_subjects_dir=None):
     return landmarks
 
 
+def _get_landmarks(landmarks, image_nii, kind=''):
+    import nibabel as nib
+    if isinstance(landmarks, (str, Path)):
+        landmarks, coord_frame = read_fiducials(landmarks)
+        landmarks = np.array([landmark['r'] for landmark in
```
hoechenberger, line 1781: It seems we're missing a test for this code path

mne_bids/write.py
```diff
@@ -1800,16 +1836,22 @@ def write_anat(image, bids_path, landmarks=None, deface=False, overwrite=False,
...
         passed to provide information for defacing. Landmarks can be determined
         from the head model using `mne coreg` GUI, or they can be determined
-        from the MRI using freeview.  If ``None`` and no ``trans`` parameter
-        is passed, no sidecar JSON file will be created.
+        from the MRI using freeview.  If a dict is passed then the values
+        must be instances of ``DigMontage`` or path-like pointing to filenames,
+        and the keys of the dict must be strings (e.g. ``'session1'``) which
+        will be used as naming suffix of the points in the sidecar JSON file.
+        If ``None`` and no ``trans`` parameter is passed, no sidecar JSON
+        file will be created.
```
hoechenberger, line 1848:
```suggestion
        from the MRI using ``freeview``.  If a dictionary is passed, then the
        values must be instances of :class:`~mne.channels.DigMontage` or
        path-like objects pointing to a :class:`~mne.channels.DigMontage`
        stored on disk, and the keys of the must be strings
        (e.g. ``'session-1'``) which will be used as naming suffix for the
        landmarks in the sidecar JSON file. If ``None``, no sidecar JSON file
        will be created.
```

mne_bids/write.py
```diff
@@ -1800,16 +1836,22 @@ def write_anat(image, bids_path, landmarks=None, deface=False, overwrite=False,
...
     deface : bool | dict
         If False, no defacing is performed.
-        If True, deface with default parameters.
-        `trans` and `raw` must not be `None` if True.
+        If True, deface with default parameters using ``landmarks``
+        that are mandatory. If multiple landmarks are provide, it will
+        use the ones called ``"deface"`` and if not present the first
+        ones in the ``landmarks`` dictionary are used.
```
hoechenberger, line 1854:
```suggestion
        If ``True``, deface with default parameters using the provided
        ``landmarks``. If multiple landmarks are provided, will
        use the ones with the suffix ``'deface'``; if no landmarks with this
        suffix exist, will use the first ones in the ``landmarks`` dictionary.
```

mne_bids/read.py
hoechenberger: Fair enough, I missed that

hoechenberger (approved):
LGTM!

mne_bids/read.py
agramfort: should be fixed now

agramfort:
@hoechenberger this one is good to go from my end

hoechenberger:
Thanks @agramfort

PR #964, opened 2022-02-22
doc: remove sphinx gallery style thats now handled upstream

### LINKED ISSUE #963: sphinx gallery update, mirror mne-python fix
https://github.com/mne-tools/mne-bids/blob/81bccd63f44099af30846af895d46ed0d112483d/doc/_static/style.css#L11-L15

we might need to mirror this change: https://github.com/mne-tools/mne-python/pull/10364

sappelhoff:
copying Eric's fix from https://github.com/mne-tools/mne-python/pull/10364 to close #963 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -7,7 +7,6 @@ import os
 import os.path as op
 import pathlib
 from datetime import datetime, timezone
-import warnings
 
 import pytest
 import shutil as sh
@@ -952,14 +951,12 @@ def test_bads_reading(tmp_path):
 @pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
 def test_write_read_fif_split_file(tmp_path):
     """Test split files are read correctly."""
-    # load raw test file and save it as as split file
+    # load raw test file, extend it to be larger than 2gb, and save it
     bids_root = tmp_path / 'bids'
     tmp_dir = tmp_path / 'tmp'
     tmp_dir.mkdir()
 
-    bids_path = _bids_path.copy().update(
-        root=bids_root, datatype='meg', extension='.fif'
-    )
+    bids_path = _bids_path.copy().update(root=bids_root, datatype='meg')
     raw = _read_raw_fif(raw_fname, verbose=False)
     bids_path.update(acquisition=None)
     write_raw_bids(raw, bids_path, verbose=False)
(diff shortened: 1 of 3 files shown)


agramfort (approved):
but not sure why CIs complain

sappelhoff:
some memory error apparently (windows only)

> numpy.core._exceptions._ArrayMemoryError: Unable to allocate 4.10 GiB for an array with shape (376, 1462765) and data type float64

agramfort:
I am sure it's the test I added months ago for split files as it produces
huge raw arrays.... crap...

hoechenberger:
I pushed a commit that aims to fix the issue, but I haven't had the time to test it locally. But even if it fails it should provide a starting point.

hoechenberger:
So the issue here is that `write_raw_bids()` doesn't accept a `split_size` parameter, hence it will only start splitting if the input `Raw` file size exceeds what can be fit into a single FIFF output file.

hoechenberger:
I think we opened a can of worms here 😅

mne_bids/tests/test_read.py
```diff
@@ -967,19 +970,58 @@ def test_write_read_fif_split_file(tmp_path):
     raw = mne.io.RawArray(data, raw.info)
     big_fif_fname = pathlib.Path(tmp_dir) / 'test_raw.fif'
     raw.save(big_fif_fname, split_size='10MB')
-    raw = _read_raw_fif(big_fif_fname, verbose=False)
-    write_raw_bids(raw, bids_path, verbose=False)
+
+    # write_raw_bids() doesn't support a `split_size` parameter, and would
+    # normally only start splitting FIFF files exceeding 2GB in size. To
+    # write split files smaller than this, we hack our way around things and
+    # manually write two files with a `split` entity.
```
hoechenberger, line 977: But this means reading the data back by passing the first filename won't automatically read the remaining splits in this test. So this is **very** synthetic.

hoechenberger:
Maybe the cleanest solution would be to simply add a private `_SPLIT_SIZE` value to `write.py`, defaulting to the MNE-Python default, but easily adjustable for our tests.
@larsoner Got an opinion on that?

hoechenberger:
Ok I introduced `write._FIFF_SPLIT_SIZE`, which can be monkey-patched during testing.
Seems to do the trick!

hoechenberger (approved):

hoechenberger:
@sappelhoff Feel free to merge if happy.

sappelhoff:
Thanks for the fix @hoechenberger!

PR #969, opened 2022-02-25
MRG: Make get_head_mri_trans() respect datatype and suffix attributes of MEG BIDSPath

hoechenberger:
`get_head_mri_trans()` currently blindly replaces the datatype and suffix of the provided MEG `BIDSPath` with `'meg'`. However, when working with derivative data, the suffix (and maybe the datatype too? needs to be discussed!) might be different, causing the function to fail as it will create a `BIDSPath` that points to a non-existent file.

The change here makes it such that if the provided MEG `BIDSPath` already has a suffix and / or datatype attribute set, they will not be modified. Only if the attributes are not set, we will fall back to `'meg'`.

This change is required to use `get_head_mri_trans()` with Maxwell-filtered data produced by the MNE-BIDS-Pipeline.

The problem was discovered yesterday while working with @SophieHerbst.

WIP:
- [x] Tests
- [x] Changelog entry
- [x] Decide whether we **really** need this for datatype too, or whether the change for suffix alone is sufficient
- [x] Why do we limit this to MEG only anyway? Should it, theoretically, not work with any recording that has digitization points?

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [x] All CIs are happy


agramfort:
can you add a test to make the pb more explicit to me? 🙏

hoechenberger:
@agramfort Yes, this is all still very much WIP, just wanted to get this started so I don't forget about it

hoechenberger:
@alexrockhill @agramfort Do you have any idea why in `get_head_mri_trans()`, we always (sometimes silently, sometimes explicitly) demand that the `BIDSPath` that basically leads to the `DigPoints` must be for an MEG recording? In theory, iEEG and EEG should work as well, no? At least as long as we add a check that the coordinates are in the `head` system?

hoechenberger:
@agramfort I've added a test that hopefully demonstrates why this change is needed, at least until we've fully sketched out derivatives in BIDS and the BIDS-Pipeline.

alexrockhill:
> @alexrockhill @agramfort Do you have any idea why in `get_head_mri_trans()`, we always (sometimes silently, sometimes explicitly) demand that the `BIDSPath` that basically leads to the `DigPoints` must be for an MEG recording? In theory, iEEG and EEG should work as well, no? At least as long as we add a check that the coordinates are in the `head` system?

Can you explain what you mean by this? I don't follow.

mne_bids/utils.py
```diff
@@ -272,18 +272,41 @@ def _infer_eeg_placement_scheme(raw):
...
-    """Extract NAS, LPA, and RPA from raw.info['dig']."""
+def _extract_landmarks(dig, assert_coord_frame=FIFF.FIFFV_COORD_HEAD):
+    """Extract NAS, LPA, and RPA from raw.info['dig'].
+
+    Parameters
+    ----------
+    assert_coord_frame : None | int
+        A coordinate frame, as defined in FIFF.FIFFV_COORD_*. If the landmark
+        coordinates are not in this coordinate frame, an exception will be
+        raised. If ``None``, no such checks will be performed.
```
hoechenberger, line 283: The idea here is to prevent silent bugs where we accidentally use the wrong coordinate system.

hoechenberger:
I will try to add a test that provokes a failure by using an unsupported coordinate system

agramfort (approved):
Ok to merge when green

thx @hoechenberger

hoechenberger:
@alexrockhill never mind, I discussed this with @agramfort and I think this approach here fixes things for us. We're dealing with files that do not conform to the BIDS naming conventions (they are derivatives) and we needed a way to use `get_head_mri_trans()` from them. I think all is working now!

PR #975, opened 2022-03-03
[ENH, MRG] Clarify intracranial narrative for why individual coordinates are preferred

alexrockhill:
Covering the use of the Talairach transform for intracranial data without pointing out that it's quite inaccurate and citing examples MNE-Python was fixed because it's misleading to present it as just as good without comment.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [x] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -45,6 +45,8 @@ Enhancements
 
 - :func:`mne_bids.get_head_mri_trans` and :func:`mne_bids.update_anat_landmarks` gained a new``kind`` parameter to specify which of multiple landmark sets to retrieve, by `Alexandre Gramfort`_ and `Richard Höchenberger`_ (:gh:`955`, :gh:`957`)
 
+- Add an explanation in :ref:`ieeg-example` why it is better to have intracranial data (especially but really MEG and EEG too) in individual rather than template coordinates, by `Alex Rockhill`_ (:gh:`975`)
+
 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -45,7 +45,7 @@ Enhancements
 
 - :func:`mne_bids.get_head_mri_trans` and :func:`mne_bids.update_anat_landmarks` gained a new``kind`` parameter to specify which of multiple landmark sets to retrieve, by `Alexandre Gramfort`_ and `Richard Höchenberger`_ (:gh:`955`, :gh:`957`)
 
-- Add an explanation in :ref:`ieeg-example` why it is better to have intracranial data (especially but really MEG and EEG too) in individual rather than template coordinates, by `Alex Rockhill`_ (:gh:`975`)
+- Add an explanation in :ref:`ieeg-example` of why it is better to have intracranial data in individual rather than template coordinates, by `Alex Rockhill`_ (:gh:`975`)
 
 API and behavior changes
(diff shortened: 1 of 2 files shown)


doc/whats_new.rst
```diff
@@ -45,6 +45,8 @@ Enhancements
 
 - :func:`mne_bids.get_head_mri_trans` and :func:`mne_bids.update_anat_landmarks` gained a new``kind`` parameter to specify which of multiple landmark sets to retrieve, by `Alexandre Gramfort`_ and `Richard Höchenberger`_ (:gh:`955`, :gh:`957`)
 
+- Add an explanation in :ref:`ieeg-example` why it is better to have intracranial data (especially but really MEG and EEG too) in individual rather than template coordinates, by `Alex Rockhill`_ (:gh:`975`)
```
agramfort, line 48:
```suggestion
- Add an explanation in :ref:`ieeg-example` of why it is better to have intracranial data in individual rather than template coordinates, by `Alex Rockhill`_ (:gh:`975`)
```

examples/convert_ieeg_to_bids.py
```diff
@@ -270,8 +273,15 @@
 # Here we'll use the MNI Talairach transform to get to ``fsaverage`` space
 # from "mri" aka surface RAS space.
 # ``fsaverage`` is very useful for group analysis as shown in
-# `Working with SEEG
-# <https://mne.tools/stable/auto_tutorials/misc/plot_seeg.html>`_.
+# :ref:`tut-working-with-seeg` Note, this is only a linear transform and so
```
agramfort, line 276:
```suggestion
# :ref:`tut-working-with-seeg`. Note, this is only a linear transform and so
```

examples/convert_ieeg_to_bids.py
```diff
@@ -270,8 +273,15 @@
...
+# loses quite a bit of accuracy relative to the needs of intracranial
```
agramfort, line 277:
```suggestion
# one loses quite a bit of accuracy relative to the needs of intracranial
```

examples/convert_ieeg_to_bids.py
```diff
@@ -282,6 +292,7 @@
 raw = mne.io.read_raw_fif(op.join(
     misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
 raw.info['line_freq'] = 60  # specify power line frequency as required by BIDS
+raw.set_channel_types({ch: 'ecog' for ch in raw.ch_names})  # fake intracranial
```
agramfort, line 295: why do you need this?

alexrockhill: It seems like it makes sense for it to be intracranial data in the raw file. This really should be checked by mne-bids that the data type matches too

alexrockhill: In the BIDSPath matches the raw file

alexrockhill:
Okay to merge?

examples/convert_ieeg_to_bids.py
hoechenberger: Why isn't the channel type set correctly automatically when reading the file with MNE? Is there a problem with the data? Or a bug with MNE?

alexrockhill:
I should read that more carefully I thought it was the meg and eeg sample data, I'll fix in asap.

agramfort (approved):

agramfort:
thx @alexrockhill

PR #977, opened 2022-03-04
MRG: Make update_anat_landmarks() accept fiducials file

### LINKED ISSUE #968: Converting fiducials read via mne.io.read_fiducials() to BIDS anatomical landmarks
MNE stores fiducials in the `mri` (surface RAS) coordinate system; however, BIDS requires MRI voxel coordinates. @alexrockhill has created the helpful function `get_anat_landmarks()` that generates landmarks in the correct coordinate system based on digitization points in an `Info` and a transformation matrix.

However, I now have a file created via the coregistration GUI (which internally calls `mne.io.write_fiducials()`), and there is no easy way to store these `mri` landmarks in my BIDS MRI's `AnatomicalLandmarkCoordinates` field.

I suppose we could modify `get_anat_landmarks()` to accept existing fiducials (or a path to a fiducials file), skip the first bits of processing it currently does with the `Info` and transfomation matrix, and convert the fiducials into the correct coordinate system.

WDYT?

cc @agramfort

hoechenberger:
Fixes #968

Merge checklist
---------------

- [x] All comments resolved
- [ ] This is not your own PR
- [x] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -916,9 +916,9 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
             f'{mri_coords_dict}'
         )
 
-    # The MRI landmarks are in "voxels". We need to convert the to the
-    # neuromag RAS coordinate system in order to compare the with MEG landmarks
-    # see also: `mne_bids.write.write_anat`
+    # The MRI landmarks are in "voxels". We need to convert them to the
+    # Neuromag RAS coordinate system in order to compare them with MEG
+    # landmarks. See also: `mne_bids.write.write_anat`
     if fs_subject is None:
         warn('Passing "fs_subject=None" has been deprecated and will raise '
              'an error in future versions. Please explicitly specify the '
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -796,8 +796,10 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
         session than the MEG. It is even possible to point to a T1 image stored
         in an entirely different BIDS dataset than the MEG data.
     fs_subject : str | None
-        The subject identifier used for FreeSurfer. If ``None``, defaults to
-        the ``subject`` entity in ``bids_path``.
+        The subject identifier used for FreeSurfer.
+
+        .. versionchanged:: 0.10
+           Deprecated defaulting to ``bids_path.subject`` if ``None``.
     fs_subjects_dir : path-like | None
         The FreeSurfer subjects directory. If ``None``, defaults to the
         ``SUBJECTS_DIR`` environment variable.
@@ -917,7 +919,11 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
     # neuromag RAS coordinate system in order to compare the with MEG landmarks
     # see also: `mne_bids.write.write_anat`
     if fs_subject is None:
+        warn('Passing "fs_subject=None" has been deprecated and will raise '
+             'an error in future versions. Please explicitly specify the '
+             'FreeSurfer subject name.', DeprecationWarning)
         fs_subject = f'sub-{meg_bids_path.subject}'
+
     fs_subjects_dir = get_subjects_dir(fs_subjects_dir, raise_error=False)
     fs_t1_path = Path(fs_subjects_dir) / fs_subject / 'mri' / 'T1.mgz'
     if not fs_t1_path.exists():
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -795,11 +795,12 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
         Use this parameter e.g. if the T1 scan was recorded during a different
         session than the MEG. It is even possible to point to a T1 image stored
         in an entirely different BIDS dataset than the MEG data.
-    fs_subject : str | None
+    fs_subject : str
+    
         The subject identifier used for FreeSurfer.
 
         .. versionchanged:: 0.10
-           Deprecated defaulting to ``bids_path.subject`` if ``None``.
+           Does not default anymore to ``bids_path.subject`` if ``None``.
     fs_subjects_dir : path-like | None
         The FreeSurfer subjects directory. If ``None``, defaults to the
         ``SUBJECTS_DIR`` environment variable.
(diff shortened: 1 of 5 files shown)


mne_bids/sidecar_updates.py
```diff
@@ -152,14 +158,30 @@ def update_anat_landmarks(
...
                   reliable way to generate the landmark coordinates in the
                   required coordinate system.
+
+        .. note:: If ``path-like``, ``fs_subject`` and ``fs_subjects_dir``
+                  must be provided as well.
+
+        .. versionchanged:: 0.10
+           Added support for ``path-like`` input.
+    fs_subject : str
+        The subject identifier used for FreeSurfer. This If ``None``, defaults
```
hoechenberger, line 178:
```suggestion
        The subject identifier used for FreeSurfer. If ``None``, defaults
```

mne_bids/sidecar_updates.py
```diff
@@ -231,6 +255,17 @@ def update_anat_landmarks(
             f'bids_path. Tried the following filenames: '
             f'{", ".join([p.name for p in tried_paths])}')
 
+    if not isinstance(landmarks, DigMontage):
```
hoechenberger, line 258:
```suggestion
    if not isinstance(landmarks, DigMontage):  # it's pathlike
```

hoechenberger:
@agramfort Turns out we **didn't** have an issue with zooming … I just used `sample` fiducials on testing-data MRI 🤦

mne_bids/sidecar_updates.py
```diff
@@ -231,6 +255,17 @@ def update_anat_landmarks(
             f'bids_path. Tried the following filenames: '
             f'{", ".join([p.name for p in tried_paths])}')
 
+    if not isinstance(landmarks, DigMontage):  # it's pathlike
+        if fs_subject is None:
+            fs_subject = bids_path.subject
```
agramfort, line 260: don't do such magic and do what the docstring says. You must provided the fs_subject in this case

mne_bids/write.py
```diff
@@ -580,9 +580,10 @@ def _mri_landmarks_to_mri_voxels(mri_landmarks, t1_mgh):
         The MRI voxel-space landmark data.
     """
     # Get landmarks in voxel space, using the T1 data
-    vox2ras_tkr = t1_mgh.header.get_vox2ras_tkr()
-    ras2vox_tkr = linalg.inv(vox2ras_tkr)
-    vox_landmarks = apply_trans(ras2vox_tkr, mri_landmarks)  # in vox
+    vox2ras_tkr_t = t1_mgh.header.get_vox2ras_tkr()
+    ras_tkr2vox_t = linalg.inv(vox2ras_tkr_t)
+    vox_landmarks = apply_trans(ras_tkr2vox_t, mri_landmarks)
```
agramfort, line 585: the changes in this function are purely cosmetic. We should avoid this to save review time

mne_bids/sidecar_updates.py
hoechenberger: I'm actually **doing** what the docstring says, I just copied this exact behavior from ``get_head_mri_trans()`` for consistency:

https://github.com/mne-tools/mne-bids/blob/b26f062a4e1c68fc83ec69df44cd345841a5aed1/mne_bids/read.py#L919-L920

I don't like it either, but I'd say we either do this in both functions or in neither.

Ok to  remove the magic here and deprecate this behavior in `get_head_mri_trans()?

mne_bids/write.py
hoechenberger: Sorry, I had originally altered the implementation of this function and later reverted to the previous behavior without reverting the cosmetic changes too. Should I revert?

mne_bids/read.py
```diff
@@ -796,8 +796,10 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
         session than the MEG. It is even possible to point to a T1 image stored
         in an entirely different BIDS dataset than the MEG data.
     fs_subject : str | None
-        The subject identifier used for FreeSurfer. If ``None``, defaults to
-        the ``subject`` entity in ``bids_path``.
+        The subject identifier used for FreeSurfer.
+
+        .. versionchanged:: 0.10
+           Deprecated defaulting to ``bids_path.subject`` if ``None``.
```
agramfort, line 802:
```suggestion
           Does not default anymore to ``bids_path.subject`` if ``None``.
```

mne_bids/read.py
```diff
@@ -917,7 +919,11 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
     # neuromag RAS coordinate system in order to compare the with MEG landmarks
     # see also: `mne_bids.write.write_anat`
     if fs_subject is None:
+        warn('Passing "fs_subject=None" has been deprecated and will raise '
```
agramfort, line 922: I see a typo above "We need to convert the to the" -> "We need to convert to the"

same with "compare the with MEG landmarks"

mne_bids/read.py
```diff
@@ -796,8 +796,10 @@ def get_head_mri_trans(bids_path, extra_params=None, t1_bids_path=None,
         session than the MEG. It is even possible to point to a T1 image stored
         in an entirely different BIDS dataset than the MEG data.
     fs_subject : str | None
```
agramfort, line 798:
```suggestion
    fs_subject : str
    
```

mne_bids/read.py
hoechenberger: Resolved

hoechenberger:
@agramfort I've addressed all your comments, this should be good to merge once CI comes back green

agramfort:
ok to merge this as is as it already fixes a practical problem but it seems the code only works if the landmarks passed as DigMontage are already in mri voxels. This should not be needed. We allow the same thing as what we allow with write_anat

@hoechenberger if you want we can merge this and please create an issue for the next stuff to do

hoechenberger:
@agramfort 
> ok to merge this as is as it already fixes a practical problem but it seems the code only works if the landmarks passed as DigMontage are already in mri voxels. This should not be needed. We allow the same thing as what we allow with write_anat

Very interesting observation, I just looked at the docstring and implementation of `write_anat()`; it calls `_get_landmarks()`, which only works with coordinates in scanner RAS and voxel space (and raises an exception otherwise).

The docstring suggests to use `get_anat_landmarks()` to produce landmark coordinates in the required coordinate frame (the function returns coords in voxel space); scanner RAS is never mentioned.

So this is at least a documentation bug / shortcoming.

But it also means that `update_anat_landmarks()` should support scanner RAS coordinates as well.

I will look into this.

hoechenberger:
@agramfort I've decided to document the inconsistency in #978 and will go ahead here and merge for now!

PR #973, opened 2022-03-02
[ENH, WIP] Add support for BIDS template standard coordinates

alexrockhill:
Fixes #961
Fixes #962

So the issue was that groups wanted to use the standard template coordinate frames but they are currently just mapped to other and then we ask them to change it manually. We have these frames already in the config file so the solution with this PR is to set them to other in the `raw` object returned by `read_raw_bids` and then explain to them that MNE does not recognize these coordinate frames so they will be treated as `unknown` but you can basically treat it like `mri` and find the fiducials and make a head->template trans.

The channels and coordinate system writing and reading is a bit of a jalopy at the moment with a bunch of things added and it's really hard to follow so I simplified but this will undoubtably break all the tests which I haven't fixed yet.

I'm pushing this now to ask for advice on what's going on with the ieeg example. In the last section, I set a transform, it gets saved in those transformed coordinates, I add back the fiducials and `compute_native_head_t` but it's way off:

```
Original: <Transform | head->unknown>
[[  1.    0.1   0.2  10.1]
 [ -0.1   1.    0.1 -20.3]
 [  0.2  -0.1   1.  -30.7]
 [  0.    0.    0.    1. ]]
Recovered: <Transform | unknown->head>
[[ 0.97590007 -0.09759001  0.19518001 -5.84352394]
 [ 0.11789673  0.98844615 -0.09526055 15.95020087]
 [-0.18362846  0.11597587  0.97613021 34.17615497]
 [ 0.          0.          0.          1.        ]]
```

No idea what's going on there, maybe someone who wrote montage code could help a bit.

Merge checklist
---------------

Maintainer, please confirm the following before merging:

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] PR description includes phrase "closes <#issue-number>"

PR #980, opened 2022-03-07
[ENH, MRG] Allow template coordinate frame writing

alexrockhill:
Fixes last part from #973

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -67,7 +67,7 @@ API and behavior changes
 
 - Passing ``fs_subject=None`` to :func:`get_head_mri_trans` has been deprecated. Please pass the FreeSurfer subject name explicitly, by Richard Höchenberger`_ (:gh:`977`)
 
-- Corrupted or missing fiducials in 'head' coordinates now raise an error instead of warning in :func:`mne_bids.write_raw_bids` by `Alex Rockhill`_ (:gh:`980`)
+- Corrupted or missing fiducials in ``head`` coordinates now raise an error instead of warning in :func:`mne_bids.write_raw_bids` by `Alex Rockhill`_ (:gh:`980`)
 
 Requirements
 ^^^^^^^^^^^^
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -56,6 +56,8 @@ Enhancements
 
 - All non-MNE-Python BIDS coordinate frames are now set to ``'unknown'`` on reading, by `Alex Rockhill`_ (:gh:`979`)
 
+- :func:`mne_bids.write_raw_bids` can now write to template coordinates by `Alex Rockhill`_ (:gh:`980`)
+
 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -64,6 +66,8 @@ API and behavior changes
 - :func:`mne_bids.get_head_mri_trans` now raises a warning if ``datatype`` or ``suffix`` of the provided electrophysiological :class:`mne_bids.BIDSPath` are not set. In the future, this will raise an exception, by `Richard Höchenberger`_(:gh:`969`)
 
 - Passing ``fs_subject=None`` to :func:`get_head_mri_trans` has been deprecated. Please pass the FreeSurfer subject name explicitly, by Richard Höchenberger`_ (:gh:`977`)
+
+- Corrupted or missing fiducials in 'head' coordinates now raise an error instead of warning in :func:`mne_bids.write_raw_bids` by `Alex Rockhill`_ (:gh:`980`)
 
 Requirements
 ^^^^^^^^^^^^
--- a/examples/convert_ieeg_to_bids.py
+++ b/examples/convert_ieeg_to_bids.py
@@ -20,6 +20,9 @@ data. Specifically, we will follow these steps:
 
 5. Repeat the process for the ``fsaverage`` template coordinate frame.
 
+6. Repeat the process for one of the other standard template coordinate frames
+   allowed by BIDS.
+
 The iEEG data will be written by :func:`write_raw_bids` with
 the addition of extra metadata elements in the following files:
 
@@ -50,6 +53,7 @@ refer to the `iEEG part of the BIDS specification`_.
 # %%
 
 import os.path as op
+import numpy as np
 import shutil
 
 from nilearn.plotting import plot_anat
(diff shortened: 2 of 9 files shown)


doc/whats_new.rst
```diff
@@ -65,6 +67,8 @@ API and behavior changes
 
 - Passing ``fs_subject=None`` to :func:`get_head_mri_trans` has been deprecated. Please pass the FreeSurfer subject name explicitly, by Richard Höchenberger`_ (:gh:`977`)
 
+- Corrupted or missing fiducials in 'head' coordinates now raise an error instead of warning in :func:`mne_bids.write_raw_bids` by `Alex Rockhill`_ (:gh:`980`)
```
hoechenberger, line 70:
```suggestion
- Corrupted or missing fiducials in ``head`` coordinates now raise an error instead of warning in :func:`mne_bids.write_raw_bids` by `Alex Rockhill`_ (:gh:`980`)
```

examples/convert_ieeg_to_bids.py
```diff
@@ -331,3 +335,133 @@
 # ``montage.add_estimated_fiducials(template, os.environ['FREESURFER_HOME'])``
 # where ``template`` maybe be ``cvs_avg35_inMNI152`` for instance
 raw.set_montage(montage)
+
+# %%
+# Step 6: Store coordinates in another template space accepted by BIDS
+# --------------------------------------------------------------------
+# As of thid writing, BIDS accepts channel coordinates in reference to the
```
hoechenberger, line 342:
```suggestion
# As of this writing, BIDS accepts channel coordinates in reference to the
```

examples/convert_ieeg_to_bids.py
```diff
@@ -331,3 +335,133 @@
...
+# .. code-block:: python
+#
+#      import nibabel as nib
+#      template_T1 = nib.load(op.join(subjects_dir, 'MNI152NLin2009bAsym',
+#                                     'mri', 'T1.mgz'))
+#      trans = mne.transforms.Transform(  # use vox to surface RAS transform
+#          fro='unknown', to='mri', trans=template_T1.header.get_vox2ras_tkr())
+#      brain = mne.viz.Brain(
+#          'sample_seeg', subjects_dir=subjects_dir, **brain_kwargs)
+#      brain.add_sensors(raw.info, trans=trans)
```
hoechenberger, line 449: Can you move the closing parens to a new line please? I think that's what we're mostly using now for new code in MNE-Python and also in MNE-BIDS.

e.g.,
```python
Foo(
    bar=baz
)
```

mne_bids/config.py
```diff
@@ -289,7 +285,7 @@
 MNE_FRAME_TO_STR = {val: key for key, val in MNE_STR_TO_FRAME.items()}
 
 # see BIDS specification for description we copied over from each
-BIDS_COORD_FRAME_DESCRIPTIONS = {
+BIDS_COORD_FRAME_DESCRIPTIONS = dict(**{
```
hoechenberger, line 288: uh there's gotta be a better way to do this. In the worst case, simply first create a dict and then append to it / update it, please.

alexrockhill:
@adam2392 these should all come back green, I think this would be nice to have in the release tomorrow but they should be reviewed. Do you have some time today to do a bit more thorough review. I've tested and it should be good but ideally it wouldn't break anything on the stable version!

mne_bids/dig.py
```diff
@@ -396,25 +387,38 @@ def _write_dig_bids(bids_path, raw, montage=None, acpc_aligned=False,
...
-        # prevent transformation back to "head", only should be used
-        # in this specific circumstance
-        if bids_path.datatype == 'ieeg':
-            montage.remove_fiducials()
-        raw.set_montage(montage)
+        # transform back to native montage coordinates if given
+        with warnings.catch_warnings():
+            warnings.filterwarnings(action='ignore', category=RuntimeWarning,
+                                    message='.*nasion not found', module='mne')
+            raw.set_montage(montage)
```
hoechenberger, line 394: Have you considered setting the verbosity level of `set_montage()` to `'error'`? (I didn't take a look at this method, not sure if it even supports verbosity!)

mne_bids/dig.py
```diff
@@ -396,25 +387,38 @@ def _write_dig_bids(bids_path, raw, montage=None, acpc_aligned=False,
...
-        return
+        raise RuntimeError("Not all digpoints have the same coordinate frame.")
 
     # get the accepted mne-python coordinate frames
     coord_frame_int = int(digpoint['coord_frame'])
     mne_coord_frame = MNE_FRAME_TO_STR.get(coord_frame_int, None)
     coord_frame = MNE_TO_BIDS_FRAMES.get(mne_coord_frame, None)
 
+    if coord_frame == 'CapTrak' and \
+            bids_path.datatype == 'eeg' or bids_path.datatype == 'nirs':
```
hoechenberger, line 414: Better readability.

But please add parens to group the logical operations. I'm not sure which of these belong together / which one(s) should take precedence?
```suggestion
    if (
        coord_frame == 'CapTrak' and
        bids_path.datatype == 'eeg' or
        bids_path.datatype == 'nirs'
    ):
```

mne_bids/dig.py
```diff
@@ -575,7 +540,10 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
     # XXX: Starting with mne 0.24, this will raise a RuntimeWarning
     #      if channel types are included outside of
     #      (EEG/sEEG/ECoG/DBS/fNIRS). Probably needs a fix in the future.
-    raw.set_montage(montage, on_missing='warn')
+    with warnings.catch_warnings():
+        warnings.filterwarnings(action='ignore', category=RuntimeWarning,
+                                message='.*nasion not found', module='mne')
+        raw.set_montage(montage, on_missing='warn')
```
hoechenberger, line 546: see above: can you maybe change the verbosity level?

mne_bids/tests/test_dig.py
```diff
@@ -0,0 +1,119 @@
...
+raw.drop_channels(raw.info['bads'])
+raw.info['line_freq'] = 60
+montage = raw.get_montage()
+
+
+def test_dig_io(tmp_path):
+    """Test passing different coordinate frames give proper warnings."""
+    bids_root = tmp_path / 'bids1'
+    for datatype in ('eeg', 'ieeg'):
+        os.makedirs(op.join(bids_root, 'sub-01', 'ses-01', datatype))
```
hoechenberger, line 46:
```suggestion
        (bids_root / 'sub-01' / 'ses-01' / datatype).mkdir(parent=True)
```

mne_bids/tests/test_dig.py
```diff
@@ -0,0 +1,119 @@
...
+        root=bids_root, datatype='meg', space='CTF')
+    with pytest.raises(ValueError, match='conflicts'):
+        write_raw_bids(raw, bids_path)
+
+
+def test_dig_template(tmp_path):
+    """Test that eeg and ieeg dig are stored properly."""
+    bids_root = tmp_path / 'bids1'
+    for datatype in ('eeg', 'ieeg'):
+        os.makedirs(op.join(bids_root, 'sub-01', 'ses-01', datatype))
```
hoechenberger, line 79:
```suggestion
        (bids_root / 'sub-01' / 'ses-01' / datatype).mkdir(parent=True)
```

mne_bids/tests/test_read.py
```diff
@@ -693,7 +693,8 @@ def test_handle_eeg_coords_reading(tmp_path):
     montage = mne.channels.make_dig_montage(ch_pos=ch_pos,
                                             coord_frame="unknown")
     raw.set_montage(montage)
-    with pytest.warns(RuntimeWarning, match="Skipping EEG electrodes.tsv"):
+    with pytest.raises(RuntimeError, match="'head' coordinate frame "
```
hoechenberger, line 696: Wait, what? So this is a breaking behavior change? If yes, we might need a deprecation cycle for this?

mne_bids/dig.py
alexrockhill: There could be other warnings though... wouldn't want to miss those

hoechenberger:
@alexrockhill We can also postpone the release to Wednesday or so, so you have more time to get this here cleaned up & tested

alexrockhill:
> @alexrockhill We can also postpone the release to Wednesday or so, so you have more time to get this here cleaned up & tested

Okay, trying to respond to all comments and test right away not to hold things up

alexrockhill:
There was one part of the coverage I wasn't able to get to: when `orient == 'n/a'` which is when MEG isn't in one of the recognized formats. According to BIDS, you should be able to write MEG in any of the EEG formats, edf for instance, but we don't currently allow that. The behavior would be that it would write with the given `BIDSPath.space` coordinate frame if one was provided, if not the behavior wouldn't change.

alexrockhill:
Ah that was an annoying bug, I modified the raw when passing it to `write_raw_bids`, this should be good to go now. Sorry this is a bit later in the day East Coast.

alexrockhill:
Ok to merge @hoechenberger and @adam2392 ?

agramfort (approved):
looks good https://6077-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_ieeg_to_bids.html#step-6-store-coordinates-in-another-template-space-accepted-by-bids

but I realize we don't support pyvista and 3D plots in the doc of mne-bids... to be done later I guess

hoechenberger:
Thanks, @alexrockhill!

alexrockhill:
> looks good https://6077-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_ieeg_to_bids.html#step-6-store-coordinates-in-another-template-space-accepted-by-bids
> 
> but I realize we don't support pyvista and 3D plots in the doc of mne-bids... to be done later I guess

Those are block quotes, not actual executed code since we don't have that template in the sample data. I wouldn't doubt that the 3D scraper is also not implemented but they shouldn't render.

agramfort:
arfff should we add this in mne-bids to make sure this code does not rot?

alexrockhill:
Sure, actually while I was writing my response, I thought that `fsaverage` is actually a template that we do have so we could use that and make it executable. It would also be a lot longer because you'd have to convert the coordinates to voxel and scanner RAS. I'll make a PR and we can see if it looks reasonable.

mne_bids/tests/test_read.py
alexrockhill: I don't think there is anything to cycle, it's raising an error where before there was a warning. I think this is appropriate and needed; I really don't like when among all the things that are printed out you get a warning that the channels weren't written which is kind of a big deal if you think they are supposed to be. This is a pretty edge case anyway, when you have a montage in `head` with no fiducials.

alexrockhill:
There were a few issues with this PR but I'm almost done fixing and rewriting the example (I think the `mne.viz.Brain` would be nice but isn't needed). I will make another PR ASAP.

PR #982, opened 2022-03-08
[BUG, MRG] Fix coordinate frames

alexrockhill:
Fixes leftovers from #980.

- The fiducials were not removed before setting the montage, causing an unwanted transformation, now they are removed and then put back which is good because we want them in the sidecar
- There was inefficiency in `mne_bids.utils._get_landmarks` because the tests for the same coordinate frame are done for `raw.get_montage` so it's better not to duplicate these
- Whew, the end of the tutorial that I finished is long winded, but I think if you are given a template, there is no way to know if it's in surface RAS, voxels or scanner RAS, it's not in the definition, so if we are supporting this, I think it's a requirement to walk users through it

I think this cleans up a lot of loose threads but there is a lot of added narrative so if @hoechenberger or @agramfort or both want to read through it and make sure it's all necessary that would be really helpful. Other than that, I am much more confident that everything works now.

cc @adam2392 also because you work with this a lot too

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/examples/convert_ieeg_to_bids.py
+++ b/examples/convert_ieeg_to_bids.py
@@ -115,7 +115,7 @@ montage.apply_trans(trans)  # head->mri
 # - `background on FreeSurfer`_
 # - `MNE-Python coordinate frames`_
 #
-# MNE-Python supports uses ``mni_tal`` and ``mri`` coordinate frames,
+# MNE-Python supports using ``mni_tal`` and ``mri`` coordinate frames,
 # corresponding to the ``fsaverage`` and ``ACPC`` (for an ACPC-aligned T1) BIDS
 # coordinate systems respectively. All other coordinate coordinate frames in
 # MNE-Python, if written with :func:`mne_bids.write_raw_bids`, must have
@@ -608,9 +608,9 @@ print('Recovered coordinate: {recovered}\n'
 # In summary, as we saw, these standard template spaces that are allowable by
 # BIDS are quite complicated. We therefore only cover these cases because
 # datasets are allowed to be in these coordinate systems, and we want to be
-# able to analyze them with MNE-Python. Because the coordinate space doesn't
-# specify a coordinate frame within that space for the standard templates
-# and because saving the raw data in the individual's ACPC space allows the
-# person analyzing the data to transform the positions to whatever template
-# they want, we recommend if at all possible, saving BIDS iEEG data in ACPC
-# space.
+# able to analyze them with MNE-Python. The template coordinate spaces
+# don't specify a coordinate frame, so it is better to save the raw data in
+# the individual's ACPC space allows the person analyzing the data to transform
+# the positions to whatever template they want. Thus, we recommend if at all
+# possible, saving BIDS iEEG data in ACPC coordinate space corresponding to the
+# individual subject's brain, not in a template coordinate frame.
--- a/examples/convert_ieeg_to_bids.py
+++ b/examples/convert_ieeg_to_bids.py
@@ -233,13 +233,7 @@ search_folder_for_text('n/a', bids_root)
 # :func:`read_raw_bids` to read in the data.
 
 # read in the BIDS dataset to plot the coordinates
-raw = read_raw_bids(bids_path=bids_path)
-
-# compare with standard
-print('Recovered coordinate: {recovered}\n'
-      'Saved coordinate:     {saved}'.format(
-          recovered=raw.info['chs'][0]['loc'][:3],
-          saved=montage.dig[0]['r']))
+raw2 = read_raw_bids(bids_path=bids_path)
 
 # %%
 # Now we have to go back to "head" coordinates with the head->mri transform.
@@ -247,11 +241,31 @@ print('Recovered coordinate: {recovered}\n'
 # .. note:: If you were downloading this from ``OpenNeuro``, you would
 #           have to run the Freesurfer ``recon-all`` to get the transforms.
 
-montage = raw.get_montage()
+montage2 = raw2.get_montage()
+
 # this uses Freesurfer recon-all subject directory
-montage.add_estimated_fiducials('sample_seeg', subjects_dir=subjects_dir)
+montage2.add_estimated_fiducials('sample_seeg', subjects_dir=subjects_dir)
+
+# get head->mri trans, invert from mri->head
+trans2 = mne.transforms.invert_transform(
+    mne.channels.compute_native_head_t(montage2))
+
 # now the montage is properly in "head" and ready for analysis in MNE
-raw.set_montage(montage)
+raw2.set_montage(montage2)
+
+# get the monage, apply the trans and make sure it's the same
+# note: the head coordinates may differ because they are defined by
+# the fiducials which are estimated; as long as the head->mri trans
+# is computed with the same fiducials, the coordinates will be the same
+# in ACPC space which is what matters
+montage2 = raw.get_montage()  # get montage in 'head' coordinates
+montage2.apply_trans(trans2)
+
+# compare with standard
+print('Recovered coordinate: {recovered}\n'
+      'Saved coordinate:     {saved}'.format(
+          recovered=montage2.get_positions()['ch_pos']['LENT 1'],
+          saved=montage.get_positions()['ch_pos']['LENT 1']))
 
 # %%
 # Step 4: Cite mne-bids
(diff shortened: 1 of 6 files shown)


alexrockhill:
Ok, everything should pass and that's all I wanted to change.

I just added an error for consistency where previously there would have been a cryptic empty dictionary error if there were not meg landmarks in the `raw` file since I was changing that part of the code already.

examples/convert_ieeg_to_bids.py
```diff
@@ -115,13 +115,12 @@
 # - `background on FreeSurfer`_
 # - `MNE-Python coordinate frames`_
 #
-# Currently, MNE-Python supports the ``mni_tal`` and ``mri`` coordinate frames,
+# MNE-Python supports uses ``mni_tal`` and ``mri`` coordinate frames,
```
adam2392, line 118:
```suggestion
# MNE-Python supports the usage of ``mni_tal`` and ``mri`` coordinate frames,
```

examples/convert_ieeg_to_bids.py
```diff
@@ -115,13 +115,12 @@
...
 # corresponding to the ``fsaverage`` and ``ACPC`` (for an ACPC-aligned T1) BIDS
 # coordinate systems respectively. All other coordinate coordinate frames in
-# MNE-Python if written with :func:`mne_bids.write_raw_bids` are written with
-# coordinate system ``'Other'``. Note, then we suggest using
-# :func:`mne_bids.update_sidecar_json` to update the sidecar
-# ``*_coordsystem.json`` file to add additional information.
+# MNE-Python, if written with :func:`mne_bids.write_raw_bids`, must have
+# an :attr:`mne_bids.BIDSPath.space` specified, and will be read in with
+# the montage channel locations set to the coordinate frame 'unknown'.
```
adam2392, line 123: Is this true? Or perhaps it can just be made a little bit more clear.

Maybe it would also help to provide a direct link to the specification for that matter?

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#electrode-description-_electrodestsv

examples/convert_ieeg_to_bids.py
```diff
@@ -603,3 +603,14 @@
...
+# %%
+# In summary, as we saw, these standard template spaces that are allowable by
+# BIDS are quite complicated. We therefore only cover these cases because
+# datasets are allowed to be in these coordinate systems, and we want to be
+# able to analyze them with MNE-Python. Because the coordinate space doesn't
+# specify a coordinate frame within that space for the standard templates
+# and because saving the raw data in the individual's ACPC space allows the
+# person analyzing the data to transform the positions to whatever template
+# they want, we recommend if at all possible, saving BIDS iEEG data in ACPC
+# space.
```
adam2392, line 616:
```suggestion
# able to analyze them with MNE-Python. The coordinate space in MNE-Python
# doesn't specify a coordinate frame for the standard templates, but saving 
# the raw data in the individual's ACPC space allows the
# person analyzing the data to transform the positions to whatever template
# they want. Therefore, we recommend if at all possible, saving BIDS iEEG
# data in ACPC coordinate space corresponding to the individual subject's
# brain, not in a template coordinate frame.
```

mne_bids/dig.py
```diff
@@ -281,24 +280,24 @@ def _write_coordsystem_json(*, raw, unit, hpi_coord_system,
...
-    dig = raw.info['dig']
-    if dig is None:
-        dig = []
-
-    coord_frame = set([dig[ii]['coord_frame'] for ii in range(len(dig))])
-    if len(coord_frame) > 1:  # noqa E501
-        raise ValueError('All HPI, electrodes, and fiducials must be in the '
-                         'same coordinate frame. Found: "{}"'
-                         .format(coord_frame))
+    if raw.get_montage() is None or raw.get_montage().get_positions() is None:
```
adam2392, line 283: Why would montage not be None, but positions are None?

mne_bids/dig.py
```diff
@@ -281,24 +280,24 @@ def _write_coordsystem_json(*, raw, unit, hpi_coord_system,
...
+        coords = dict(
+            NAS=list() if pos['nasion'] is None else pos['nasion'].tolist(),
+            LPA=list() if pos['lpa'] is None else pos['lpa'].tolist(),
+            RPA=list() if pos['rpa'] is None else pos['rpa'].tolist())
 
     # get the coordinate frame description
     sensor_coord_system_descr = (BIDS_COORD_FRAME_DESCRIPTIONS
                                  .get(sensor_coord_system.lower(), "n/a"))
 
-    coords = _extract_landmarks(dig)
```
adam2392, line 298: What's wrong with `_extract_landmarks`?

adam2392 (changes requested):
Minor changes, but otw lgtm

examples/convert_ieeg_to_bids.py
alexrockhill: If it's not in `MNE_TO_BIDS_FRAMES` and there is no `BIDSPath.space` then it won't be written. It could default to 'Other' that's a -1 from me, I don't think `Other` should be in BIDS.

alexrockhill: You get

```
warn("Coordinate frame could not be inferred from the raw object "
                 "and the BIDSPath.space was none, skipping the writing of "
                 "channel positions")
```

agramfort:
tell me when I should look here again.

examples/convert_ieeg_to_bids.py
alexrockhill: It's linked at the top, we could add it there, but the whole narrative is trying to get them not to save data in a template coordinate frame if possible so handing a list of template coordinate frames that close to the beginning of the example seems counterproductive

mne_bids/dig.py
alexrockhill: It's redundant with `raw.get_montage()` which checks the same things. Seems wise to use the checks there rather than reimplementing them.

alexrockhill:
Ok, I added @adam2392 's comments, if you could review @agramfort that would be really helpful

mne_bids/dig.py
alexrockhill: I got an error in the tests that positions were none I thought, let me double check

adam2392: Okay I might be wrong. Maybe it's a downstream consequence of mne

alexrockhill:
The narrative and code is really lengthy at the end to describe all the coordinate frame transforms. I think in the future we could refactor `mne_bids.read.get_head_mri_trans` to work with iEEG data but that would require the template being in BIDS format with fiducials in the sidecar. Since we don't have `fsaverage` that way in the sample data, maybe this isn't worth doing.

What might be worth doing eventually is to download all the templates, find fiducials on their T1s and save them somewhere in `mne_bids`. That way, montage could be transformed to `mri` by `mne_bids.dig._read_dig_bids` and then they could just use `mne.compute_native_head_t` to get the `trans`. The only thing is that we'd have to do the same circumventing the transform to `head` by `raw.set_montage` so that the data could be returned in `mri` so that they could compute the `trans`. That or `read_raw_bids` could return a `trans` as well which actually might be a nice convenience for EEG and MEG too.

adam2392:
> What might be worth doing eventually is to download all the templates, find fiducials on their T1s and save them somewhere in `mne_bids`. 

Seems like a good GSoC issue :p. Would be a pretty valuable contribution.

alexrockhill:
I guess it wouldn't be too hard but the only thing is that you'd have to infer whether the channels.tsv coordinates were in surface RAS, voxels or scanner RAS. You could also have a keyword to specific with `auto` as the default.

alexrockhill:
@adam2392 look okay now?

alexrockhill:
I opened an issue on `bids-specification` which is really the most upstream point where this should be solved, maybe we'll get a bit of help

examples/convert_ieeg_to_bids.py
```diff
@@ -325,152 +331,286 @@
...
-#      )
-#      brain = mne.viz.Brain(
-#          'sample_seeg', subjects_dir=subjects_dir, **brain_kwargs)
-#      brain.add_sensors(raw.info, trans=trans)
+# In summary, as we saw, these standard template spaces that are allowable by
+# BIDS are quite complicated. We therefore only cover these cases because
+# datasets are allowed to be in these coordinate systems, and we want to be
+# able to analyze them with MNE-Python. The template coordinate spaces
+# don't specify a coordinate frame, so it is better to save the raw data in
+# the individual's ACPC space allows the person analyzing the data to transform
```
adam2392, line 613:
```suggestion
# the individual's ACPC space allowing the person analyzing the data to transform
```

adam2392 (approved):

hoechenberger (approved):
Anyone, feel free to merge once CI goes green!

alexrockhill:
Ok, all green

hoechenberger:
Thanks, @alexrockhill!

PR #983, opened 2022-03-09
[ENH, MRG] Add BIDS Standard Template Transformations

alexrockhill:
I went ahead and downloaded all the T1s, marked their fiducials in voxels in `Freeview` and then computed the transforms. I think I'll wait until https://github.com/mne-tools/mne-bids/pull/982 merges first and then add these with a `mne_bids.get_template_trans` which will dramatically simplify the end. It was kind of cool to see all the different templates, there are definitely some interesting ones, maybe that will be something to incorporate in the future.

Also, I have no idea where to store these but they are 264KB total so I don't think it's too big of a deal where they go. It would be nice to store the T1s as well but then we'd probably have to make sure with their licenses.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


alexrockhill:
Oh also the T1s are here: https://drive.google.com/drive/folders/1rN-Rb7D6ZTyVKOc2WZ-AM3f3dspIhZtQ?usp=sharing and the script is in the first commit but then I deleted it and committed again so it doesn't accidentally end up merged.

larsoner:
You only transformed from one template to another using fiducials? This seems much less accurate than using flirt or ants to compute a proper full some

adam2392:
> Oh also the T1s are here: https://drive.google.com/drive/folders/1rN-Rb7D6ZTyVKOc2WZ-AM3f3dspIhZtQ?usp=sharing and the script is in the first commit but then I deleted it and committed again so it doesn't accidentally end up merged.

Unrelated to the PR I suppose:

If BIDS doesn't take it, I would still be inclined to store it somewhere easily accessible; maybe `mne-bids-data`? I can imagine a future where leveraging the template T1s, one can verify their individual coordinates "look right". Then they can do any series of transformation from template to template and plot the corresponding coordinates and verify that it still "looks right" without having to leave their codebase (i.e. T1s are accessible programmatically).

alexrockhill:
There are a lot of them and the largest is 180 MB. I think that could be in a data repository though.

alexrockhill:
I think because you need to anonymize and upload an individual T1 per subject for subject-specific coordinates and you don't need any anatomical data for the templates the lower bar to entry will make people want to share in template coordinates whether we recommend against it in `mne-bids` or not. So IMO, probably better to get on board and support it while still recommending to share in ACPC.

alexrockhill:
744MB total for the T1s as I in `nii.gz`, they could be `mgz` too and that might save a bit of space.

alexrockhill:
> You only transformed from one template to another using fiducials? This seems much less accurate than using flirt or ants to compute a proper full affine

This doesn't transform between templates, it just defines a `head` coordinate frame so that you can use MNE functions that require a `trans` and will facilitate transforming between scanner RAS, voxels and surface RAS depending on which coordinate frame the template channel locations are stored in. Sorry I missed this comment somehow @larsoner!

alexrockhill:
Any thoughts on where to put the transforms, is this okay?

larsoner:
Okay as long as it's just to properly define a head coord system this seems fine!

alexrockhill:
Ok, this drastically reduces the complexity of the ieeg tutorial, so I think it's really good. I tried a few version (e.g. just providing the `trans`) but this is all so complicated when you start mixing coordinate frames and units that I think it's better to just abstract everything. I'm not sure how everyone else feels about `auto` but I think because Freesurfer coordinate frames have `scanner RAS` == `surface RAS` this isn't in big danger of a mixup, which I added to the text. I think the ideal thing is someone grabs a dataset off OpenNeuro and it just works (that's kind of the whole point of BIDS) so I think this is the best solution until https://github.com/bids-standard/bids-specification/issues/1025 gets resolved and hopefully they just pick a coordinate frame, probably voxels, that the electrodes.tsv data has to be in.

hoechenberger:
@alexrockhill Is this something we should try to squeeze into 0.10 as well or..?

alexrockhill:
> @alexrockhill Is this something we should try to squeeze into 0.10 as well or..?

It would probably be nice, it's mostly just cleaning up and making things nicer. Last PR though before the release definitely

alexrockhill:
I think it should all be green now, I tested it thoroughly on my side, I just didn't configure accessing the data directly in the repository

alexrockhill:
Ok, looks good to go, here's the artifact: https://6099-89170358-gh.circle-artifacts.com/0/dev/auto_examples/convert_ieeg_to_bids.html

alexrockhill:
Ok just a few typos, ready for review if someone has a chance

adam2392:
Looks pretty good to me. Something I wished for when I was still a phd student Haha.

alexrockhill:
> Looks pretty good to me. Something I wished for when I was still a phd student Haha.

Yeah, it took a bit of doing but I think well worth it.

doc/api.rst
```diff
@@ -31,6 +31,7 @@ mne_bids
    get_anat_landmarks
    update_anat_landmarks
    get_head_mri_trans
+   template_to_head
```
hoechenberger, line 34: There's a bit of an inconsistency in the naming here. Maybe the new function should be called `template_to_head_trans()`?

mne_bids/dig.py
```diff
@@ -543,6 +546,104 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+            d['r'] /= 1000  # mm -> m
+    # now add fiducials (in mri coordinates)
+    fids = mne.io.read_fiducials(
+        data_dir.joinpath(f'space-{space}_fiducials.fif'))[0]
+    montage.dig = fids + montage.dig  # add fiducials
+    for fid in fids:  # ensure also in mri
+        fid['coord_frame'] = MNE_STR_TO_FRAME['mri']
+    raw.set_montage(montage)  # transform to head
+    # finally return montage
+    return mne.read_trans(data_dir.joinpath(f'space-{space}_trans.fif'))
```
hoechenberger, line 649:
```suggestion
    return mne.read_trans(data_dir / f'space-{space}_trans.fif')
```

mne_bids/dig.py
```diff
@@ -543,6 +546,104 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+@verbose
+def template_to_head(raw, space, coord_frame='auto', unit='auto',
+                     verbose=None):
+    """Transform a BIDS standard template montage to the head coordinate frame.
+
+    Parameters
+    ----------
+    raw : mne.io.Raw
+        The MNE-Python Raw object with a montage in one of the BIDS standard
+        template coordinate frames (modified in place).
```
hoechenberger, line 563: Passing an `Info` or even simply a `DigMontage` should suffice to get the `trans`, right? I think we should support this… or do you think it doesn't make sense, because usually the user will have a `Raw` (or `Epochs` or …) that they want to "convert", and they're not looking for **just** a `trans`?

doc/api.rst
alexrockhill: It transforms the instance (could be an epoch but it's associated with `read_raw_bids` so I think `raw` is fine) so I think it shouldn't include the word `trans` because then someone might miss that it transforms the object, the `trans` is just secondary

mne_bids/dig.py
alexrockhill: I tried it with the montage originally but it was really cumbersome because you ended up getting, setting and then re-getting the montage in the examples. I'll change it to `info` though.

alexrockhill: I tried to use info but I forgot it doesn't have the `ContainsMixin` so you can't `get_montage` although you can set it. @larsoner or @agramfort, is there a reason `info` can't `get_montage`? I'd be happy to add it in a quick PR if not but I assume so.

alexrockhill: Or maybe `get_montage` should moved from the `ContainsMixin` to the `MontageMixin`, I'm not sure if that would break everything though

alexrockhill: Ok I opened a PR and since there is about to be a release, this will immediately work with the stable branch of MNE (after the release). Since it's a new function, it won't break for anyone but will be a great reminder to upgrade to 1.0.

alexrockhill:
Ok, I realized even if we merge the mixins PR, then it's a pain to depend on 1.0 so soon so I just copied it over. Is there a way to set a reminder to do something at the next release? At 0.11, we should switch it to just `info.get_montage`.

larsoner:
I would make a:
```
# Remove once we depend on MNE-Python 1.0+
def _get_montage(info):
    try:
        info.get_montage
    except AttributeError:
        return RawArray(np.zeros((info['nchan'], 1)), info).get_montage()
    else:
        return info.get_montage()
```

mne_bids/dig.py
```diff
@@ -543,6 +546,104 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+            ch['coord_frame'] = MNE_STR_TO_FRAME['unknown']
         for d in raw.info['dig']:
-            d['coord_frame'] = _str_to_frame['unknown']
+            d['coord_frame'] = MNE_STR_TO_FRAME['unknown']
+
+
+@verbose
+def template_to_head(raw, space, coord_frame='auto', unit='auto',
+                     verbose=None):
+    """Transform a BIDS standard template montage to the head coordinate frame.
```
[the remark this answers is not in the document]
alexrockhill: The template montage is getting transformed to head and you need a head->mri transform to get it to work with functions that require a `raw` and a `trans`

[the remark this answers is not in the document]
adam2392: Ah yeah I see. Hard to do the mental gymnastics sometimes...

In that case, I wonder if it's worth adding a `Notes` section that explicitly warns the user there is literally no way of checking if you do something wrong (e.g. putting in the incorrect name in `space` parameter). Say someone puts `fsaverage5` instead of `fsaverage6`.

adam2392:
Looks good to me. Can't do a thorough review of the code, but the details in the example are nice cuz I remember always having a million questions about these kinds of transformations.

alexrockhill:
Ok, I think that's a bit uglier with all the `template_to_head(...)[1]` but if that was confusing you both @hoechenberger and @adam2392, then better to make it clear that the info is being modified in the return type. That's how functions like `filter` do it anyway so probably best.

mne_bids/dig.py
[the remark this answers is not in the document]
alexrockhill: Neither `fsaverage5` or `fsaverage6` are supported, those are deprecated haha.

You can plot the alignment with the T1 of the template space, if you put the wrong one, it will look off... We can add that later, there's an issue open because `mne-bids` doesn't support 3D plotting yet so I would have but not this PR

alexrockhill:
Ok to merge @adam2392?

adam2392 (approved):

agramfort (approved):
I also checked that the new files are also shipped with the source dist.

mne_bids/dig.py
```diff
@@ -543,6 +546,116 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+
+# Remove once we depend on MNE-Python 1.0+
+def _get_montage(info):
+    try:
+        info.get_montage
+    except AttributeError:
+        return mne.io.RawArray(np.zeros((info['nchan'], 1)),
+                               info).get_montage()
+    else:
+        return info.get_montage()
```
hoechenberger, line 562:
```suggestion
    try:
        return info.get_montage()
    except AttributeError:
        return mne.io.RawArray(np.zeros((info['nchan'], 1)),
                               info).get_montage()        
```

mne_bids/dig.py
```diff
@@ -543,6 +546,116 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+            so be very careful not to assume a BIDS dataset's coordinates are
+            in one when they are actually in the other. The only way to tell
+            for template coordinate systems, currently, is if it is specified
+            in the dataset documentation.
+
+    unit : str
+        The unit that was used in the coordinate system specification; either
+        ``m`` or ``mm``. If ``auto``, ``m`` will be inferred if the montage
+        spans less than -1 to 1 in units (meters) and ``mm`` otherwise. If the
+        ``coord_frame`` is 'mri_voxel', ``unit`` will be ignored.
```
hoechenberger, line 597:
```suggestion
    unit : 'm' | 'mm' | 'auto'
        The unit that was used in the coordinate system specification.
        If ``'auto'``, ``'m'`` will be inferred if the montage
        spans less than ``-1`` to ``1`` in units, and ``'mm'`` otherwise. If the
        ``coord_frame`` is ``'mri_voxel'``, ``unit`` will be ignored.
```

mne_bids/dig.py
```diff
@@ -543,6 +546,116 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+        The name of the BIDS standard template. See
+        https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#standard-template-identifiers
+        for a list of acceptable spaces.
+    coord_frame : str
+        BIDS template coordinate systems do not specify a coordinate frame
+        so this must be determined by inspecting the documentation for the
+        dataset or the electrodes.tsv file. Must be ``mri``, ``mri_voxel``
+        or ``ras``. If ``auto``, the montage will be assumed to be in
+        ``mri_voxel`` if the coordinates are strictly positive and
+        ``scanner RAS`` otherwise.
```
hoechenberger, line 583:
```suggestion
    coord_frame : 'mri' | 'mri_voxel' | 'ras'
        BIDS template coordinate systems do not specify a coordinate frame,
        so this must be determined by inspecting the documentation for the
        dataset or the ``electrodes.tsv`` file. Must be.
        If ``'auto'``, assumes ``'mri_voxel'`` if the coordinates are strictly
        positive, and ``'ras'`` ("scanner RAS") otherwise.
```

mne_bids/dig.py
```diff
@@ -543,6 +546,116 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+        ``m`` or ``mm``. If ``auto``, ``m`` will be inferred if the montage
+        spans less than -1 to 1 in units (meters) and ``mm`` otherwise. If the
+        ``coord_frame`` is 'mri_voxel', ``unit`` will be ignored.
+    %(verbose)s
+
+    Returns
+    -------
+    %(info_not_none)s The modified ``Info`` object.
+    trans : mne.transforms.Transform
+        The data transformation matrix from ``head`` to ``mri`` coordinates.
```
hoechenberger, line 604:
```suggestion
        The data transformation matrix from ``'head'`` to ``'mri'`` coordinates.
```

mne_bids/dig.py
```diff
@@ -543,6 +546,116 @@ def _read_dig_bids(electrodes_fpath, coordsystem_fpath,
...
+        vox_mri_trans = mne.read_trans(
+            data_dir / f'space-{space}_vox-mri_trans.fif')
+        montage.apply_trans(vox_mri_trans)
+    assert montage.get_positions()['coord_frame'] == 'mri'
+    if not (unit == 'm' and coord_frame == 'mri'):  # if so, already in m
+        for d in montage.dig:
+            d['r'] /= 1000  # mm -> m
+    # now add fiducials (in mri coordinates)
+    fids = mne.io.read_fiducials(
+        data_dir / f'space-{space}_fiducials.fif')[0]
```
hoechenberger, line 655:
```suggestion
    fids = mne.io.read_fiducials(
        data_dir / f'space-{space}_fiducials.fif'
    )[0]
```

hoechenberger (commented):
Otherwise LGTM!

hoechenberger:
Doc build is failing, apparently due to an issue with MNE-NIRS?

```python
/home/circleci/project/examples/convert_nirs_to_bids.py failed leaving traceback:
  File "/home/circleci/project/examples/convert_nirs_to_bids.py", line 31, in <module>
    import mne_nirs  # For convenient downloading of example data
  File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/__init__.py", line 10, in <module>
    from . import statistics
  File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/statistics/__init__.py", line 7, in <module>
    from ._glm_level_first import (RegressionResults, ContrastResults,
  File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/statistics/_glm_level_first.py", line 19, in <module>
    from mne.channels.channels import ContainsMixin
ImportError: cannot import name 'ContainsMixin' from 'mne.channels.channels' (/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne/channels/channels.py)
```

cc @rob-luke 


@alexrockhill I'm going ahead and merging this one, as the issue with the doc build seems unrelated to your changes. Thank you!

alexrockhill:
> Doc build is failing, apparently due to an issue with MNE-NIRS?
> 
> ```python
> /home/circleci/project/examples/convert_nirs_to_bids.py failed leaving traceback:
> Traceback (most recent call last):
>   File "/home/circleci/project/examples/convert_nirs_to_bids.py", line 31, in <module>
>     import mne_nirs  # For convenient downloading of example data
>   File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/__init__.py", line 10, in <module>
>     from . import statistics
>   File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/statistics/__init__.py", line 7, in <module>
>     from ._glm_level_first import (RegressionResults, ContrastResults,
>   File "/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne_nirs/statistics/_glm_level_first.py", line 19, in <module>
>     from mne.channels.channels import ContainsMixin
> ImportError: cannot import name 'ContainsMixin' from 'mne.channels.channels' (/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne/channels/channels.py)
> ```
> 
> cc @rob-luke 
> 
> 
> @alexrockhill I'm going ahead and merging this one, as the issue with the doc build seems unrelated to your changes. Thank you!

Should be a quick fix, I just moved that mixin to mne.io.meas_info.

hoechenberger:
> Should be a quick fix, I just moved that mixin to mne.io.meas_info.

Would be awesome if you could have a look at this if you have a few minutes to spare! Once the doc build is working again, I can release MNE-BIDS 0.10.

PR #990, opened 2022-03-29
[BUG, MRG] Fix ACPC Coordinates in surface RAS when they should be in scanner RAS

### LINKED ISSUE #989: [BUG] ACPC in surface RAS, should be in Scanner RAS
I realized that in https://mne.tools/mne-bids/dev/auto_examples/convert_ieeg_to_bids.html the ACPC coordinates are in surface RAS but in https://mne.tools/stable/auto_tutorials/clinical/10_ieeg_localize.html we tell users (correctly) to align scanner RAS to ACPC. I can make a PR, it might be nice to use the Freesurfer T1 in a convenience function that converts a montage to scanner RAS. Although someone could share a T1 that isn't 256 x 256 x 256 and then the Freesurfer T1 wouldn't be the same size, but, I think, since RAS is in mm/m, it should be compensated for by the affine and should be fine to use the Freesurfer T1 even in that instance.

alexrockhill:
Fixes #989.

It's a bit inelegant but I encapsulated the ugly parts behind helper functions. MNE uses m and surface RAS and nibabel/BIDS uses mm (generally) and scanner RAS which makes a lot of headaches. I think this is about as clean as possible but let me know if there are things that should be fixed. This is a pretty high priority also because `main` is wrong will cause incorrectly formatted datasets.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -57,7 +57,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- Fix ACPC in ``surface RAS`` instead of ``scanner RAS`` in :ref:`ieeg-example` and add convence functions :func:`mne_bids.get_ras_mri_trans` and :func:`mne_bids.get_mri_ras_trans` to help, by `Alex Rockhill`_ :gh:`XXX`
+- Fix ACPC in ``surface RAS`` instead of ``scanner RAS`` in :ref:`ieeg-example` and add convence functions :func:`mne_bids.convert_montage_to_ras` and :func:`mne_bids.convert_montage_to_mri` to help, by `Alex Rockhill`_ :gh:`XXX`
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -57,7 +57,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- Fix ACPC in ``surface RAS`` instead of ``scanner RAS`` in :ref:`ieeg-example` and add convence functions :func:`mne_bids.convert_montage_to_ras` and :func:`mne_bids.convert_montage_to_mri` to help, by `Alex Rockhill`_ :gh:`990`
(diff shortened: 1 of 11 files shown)


doc/whats_new.rst
```diff
@@ -57,7 +57,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- ...
+- Fix ACPC in ``surface RAS`` instead of ``scanner RAS`` in :ref:`ieeg-example` and add convence functions :func:`mne_bids.convert_montage_to_ras` and :func:`mne_bids.convert_montage_to_mri` to help, by `Alex Rockhill`_ :gh:`990`
```
agramfort, line 60: convence?

hoechenberger:
I'm wondering if this shouldn't better be implemented as methods of DigMontage upstream in MNE? What am I missing?

alexrockhill:
> I'm wondering if this shouldn't better be implemented as methods of DigMontage upstream in MNE? What am I missing?

Yes, that would be nice if they were but then we'd have to require the development version of MNE. I can do a PR now to add to MNE-Python and then we can change it next MNE-Python release.

hoechenberger:
@larsoner If you have a few minutes, could you please share your thoughts on this one?

alexrockhill:
LGTM?

mne_bids/tests/test_dig.py
```diff
@@ -36,6 +37,8 @@
 data_path = testing.data_path()
 raw_fname = op.join(data_path, 'MEG', 'sample',
                     'sample_audvis_trunc_raw.fif')
+trans = mne.read_trans(op.join(data_path, 'MEG', 'sample',
+                               'sample_audvis_trunc-trans.fif'))
```
larsoner, line 41: It's better not to `read_*` at the root level because this happens during test *collection*. Better to leave as `str` and use it below instead. I would make this `trans_fname` and create `trans` below

larsoner (commented):
I would also 1) avoid the `scale` stuff by being careful about units, instead, and either way 2) use `combine_transforms` and apply once. But the given approach seems okay as is so +1 for merge after fixing the test collection bloat issue

mne_bids/tests/test_dig.py
alexrockhill: How do you avoid the scale? Divide the `T1.header.get_ras2vox()` by 1000?

larsoner (approved):
LGTM +1 for merge

PR #999, opened 2022-04-14
WIP: Allow storage of crosstalk and fine-calibration via write_raw_bids()

hoechenberger:
@agramfort WDYT?

The idea is that with these changes and the changes to empty-room handling we just merged into `main`, users would only have to call a single function -- `write_raw_bids()` -- once to get all their MEG data stored: experimental data, empty-room, fine-calibration and crosstalk.

Currently, this requires 3 function calls:
- `write_raw_bids()`
- `write_meg_calibration()`
- `write_meg_crosstalk()`

With these changes here, it would come down to `write_raw_bids(raw, bids_path, empty_room=..., meg_calibration=..., meg_crosstalk=...)` and you'd be all set.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1207,11 +1207,15 @@ def make_dataset_description(*, path, name, hed_version=None,
 
 
 @verbose
-def write_raw_bids(raw, bids_path, events_data=None, event_id=None,
-                   anonymize=None, format='auto', symlink=False,
-                   empty_room=None, allow_preload=False,
-                   montage=None, acpc_aligned=False,
-                   overwrite=False, verbose=None):
+def write_raw_bids(
+    raw, bids_path, *,
+    events_data=None, event_id=None,
+    anonymize=None, format='auto', symlink=False,
+    empty_room=None, meg_calibration=None, meg_crosstalk=None,
+    allow_preload=False,
+    montage=None, acpc_aligned=False,
+    overwrite=False, verbose=None
+):
     """Save raw data to a BIDS-compliant folder structure.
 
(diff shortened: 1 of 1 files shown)


agramfort:
I am not convinced. It's 3 function calls but with the same bids_path so it's really not much pain. Much less that emptyroom I would say. Now since you always write fine cal and cross talk maybe this could be a single function... but it's an API change so...

my 2c

hoechenberger:
@agramfort If your main concern is the API change:
Since we're merely adding two now parameters relatively "late" in (or, at the end of) the function signature (at positions 8 and 9!), I don't think this will cause any trouble …

And to make things more "future-proof", I'm switching to kw-only args after the 4th parameter (so that raw, bids_path, events_data, and event_id can still be passed by position)

Not convinced..?

sappelhoff:
Also: isn't fine calibration and crosstalk only needed for meg data recorded from Elekta/neuromag/Megin hardware?

agramfort:
yes exactly it's only for one MEG vendor/system

jasmainak:
I would have been more convinced if you could get these files from the BIDS structure without having to provide separate paths :)

hoechenberger:
> I would have been more convinced if you could get these files from the BIDS structure without having to provide separate paths :)

That's what we already support when reading, but this PR is about writing … or did I misunderstand you?

jasmainak:
oh indeed, yin and yang :) Sorry for the noise!

agramfort:
discussed with @hoechenberger

Issue #1005, opened 2022-05-12
Manufacturers list does not include an extension for the EGI device


anandsaini024:
While using write_raw_bids() for the .mff data file (EGI Device), it throws an error because ``mne-bids`` infers the manufacturer information from the file extension. In my case the when it looks for extension in the files, it is '.bin'.

The manufactures [list](https://github.com/mne-tools/mne-bids/blob/14dde8535d0538537b04fde1c9f1b4dc854239ea/mne_bids/config.py#:~:text=eeg_manufacturers%20%3D%20%7B,%27Neuroscan%27%7D) does not include the manufacturer's details in the dictionary for EGI devices.

Although If the file format is non-standard for the manufacturer, it is required to update the manufacturer field in the sidecars manually. However, reporting this as you probably would want to include the EGI in the manufacturers list as well.

The error:

---------------------------------------------------------------------------
KeyError    Traceback (most recent call last)
Input In [12], in <cell line: 1>()
----> 1 write_raw_bids(raw, bids_path, format='BrainVision', allow_preload=True, overwrite=True)

File <decorator-gen-593>:24, in write_raw_bids(raw, bids_path, events_data, event_id, anonymize, format, symlink, empty_room, allow_preload, montage, acpc_aligned, overwrite, verbose)

File c:\users\asaini\desktop\..\venv\lib\site-packages\mne_bids\write.py:1668, in write_raw_bids(***failed resolving arguments***)
   1663 make_dataset_description(bids_path.root, name=" ", overwrite=False)
   1665 _sidecar_json(raw, task=bids_path.task, manufacturer=manufacturer,
   1666               fname=sidecar_path.fpath, datatype=bids_path.datatype,
   1667               emptyroom_fname=associated_er_path, overwrite=overwrite)
-> 1668 _channels_tsv(raw, channels_path.fpath, overwrite)
   1670 # create parent directories if needed
   1671 _mkdir_p(os.path.dirname(data_path))

File c:\users\asaini\desktop\..\venv\lib\site-packages\mne_bids\write.py:110, in _channels_tsv(raw, fname, overwrite)
    108 # get the manufacturer from the file in the Raw object
    109 _, ext = _parse_ext(raw.filenames[0])
--> 110 manufacturer = MANUFACTURERS[ext]
    112 ignored_channels = IGNORED_CHANNELS.get(manufacturer, list())
    114 status, ch_type, description = list(), list(), list()

KeyError: '.bin'

[...]

PR #1006, opened 2022-05-13
Update config.py

anandsaini024:
Added EGI as a manufacturer in the eeg_manufacturers dict for extension '.bin' and '.mff'. This PR closes issue #1005.

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -52,7 +52,8 @@ eeg_manufacturers = {'.vhdr': 'Brain Products', '.eeg': 'Brain Products',
                      '.set': 'n/a', '.fdt': 'n/a',
                      '.lay': 'Persyst', '.dat': 'Persyst',
                      '.EEG': 'Nihon Kohden',
-                     '.cnt': 'Neuroscan', '.CNT': 'Neuroscan'}
+                     '.cnt': 'Neuroscan', '.CNT': 'Neuroscan'
+                     '.bin': 'EGI', '.mff': 'EGI'}
 
 ieeg_manufacturers = {'.vhdr': 'Brain Products', '.eeg': 'Brain Products',
                       '.edf': 'n/a', '.EDF': 'n/a', '.set': 'n/a',


mne_bids/config.py
```diff
@@ -52,7 +52,8 @@
                      '.set': 'n/a', '.fdt': 'n/a',
                      '.lay': 'Persyst', '.dat': 'Persyst',
                      '.EEG': 'Nihon Kohden',
-                     '.cnt': 'Neuroscan', '.CNT': 'Neuroscan'}
+                     '.cnt': 'Neuroscan', '.CNT': 'Neuroscan'
```
mscheltienne, line 55:
```suggestion
                     '.cnt': 'Neuroscan', '.CNT': 'Neuroscan',
```

adam2392:
Can you add a change log entry and a unit test for writing?

adam2392:
Hey just checking in to see if you hit any snags.

I suspect if you add necessary dataset to here it should test the writing of this EGI data format:
https://github.com/mne-tools/mne-bids/blob/b4250ec348a80ac14ec8a14f1a51b0a30afdec8b/mne_bids/tests/test_write.py#L103-L107

Not entirely sure tho. The tests definitely need a bit of organizing :p.

anandsaini024:
> y just checking in to see if you hit any snags.
> 
> I suspect if you add necessary dataset to here it

Hey, was busy with something else in between. I will look at this tomorrow.

mne_bids/tests/test_write.py
```diff
@@ -107,13 +107,15 @@ def fn(fname, *args, **kwargs):
...
 _read_raw_snirf = _wrap_read_raw(mne.io.read_raw_snirf)
+_read_raw_egi = _wrap_read_raw(mne.io.read_raw_egi)
 
 # parametrized directory, filename and reader for EEG/iEEG data formats
 test_eegieeg_data = [
     ('EDF', 'test_reduced.edf', _read_raw_edf),
     ('Persyst', 'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay', _read_raw_persyst),  # noqa
     ('NihonKohden', 'MB0400FU.EEG', _read_raw_nihon),
     ('CNT', 'scan41_short.cnt', _read_raw_cnt),
+	('EGI', 'test_egi.mff', _read_raw_egi),
```
hoechenberger, line 118:
```suggestion
    ('CNT', 'scan41_short.cnt', _read_raw_cnt),
    ('EGI', 'test_egi.mff', _read_raw_egi),
```

hoechenberger:
@anandsaini024 It appears you'll have to amend `allowed_extensions_meg` in `config.py`

mscheltienne:
It's still going to fail because with this test file: `'/Users/scheltie/mne_data/MNE-testing-data/EGI/test_egi.mff'`
when doing: `events, _ = mne.events_from_annotations(raw, event_id=None)`
events is returned as an empty array `array([], shape=(0, 3), dtype=int64)`
while after conversion and re-reading with `raw2 = read_raw_bids(bids_path=bids_output_path)`
the events returned with `events2, _ = mne.events_from_annotations(raw2)`
do have a single element.. `array([[    0,     0, 99999]])`.

adam2392:
Yes I think it will involve some refactoring of the unit tests

mscheltienne:
If you have time to finish this PR during the code sprint next week, it would be super appreciated! I can give you a hand if you want. For now, the issue I ran into 3 weeks ago is related to the conversion to the brain vision format. Take the code snippet below:

```
from mne.io import read_raw_brainvision, read_raw_egi
from mne_bids import BIDSPath, write_raw_bids


directory = testing.data_path() / "EGI"
raw = read_raw_egi(directory / "test_egi.mff")

root="C:/Users/Mathieu/Downloads/tests/bids1"
bids_path = BIDSPath(
    subject="01", session="01", run="01", acquisition="01",
    task="testing", root=root)
write_raw_bids(raw, bids_path)
raw2 = read_raw_brainvision(bids_path.fpath)
```

The `raw.annotations` is empty while the loaded `raw2.annotations` contains one segment. Up to the call to `pybv.write_brainvision`, the `raw.annotations` is empty. So it's either the writer from `pybv` or the reader from `mne` that is responsible for this segment. I am not very familiar with the brainvision format, maybe @sappelhoff or @cbrnr you know where this segment is coming from?

scott-huberty:
Hey @mscheltienne , thanks! It would be great to briefly meet with you this week if you have some time, just so I can catch myself up and get a hold of the work that's been done already.

scott-huberty:
@mscheltienne sorry for the belated reply, I am available now and for the rest of my afternoon!

adam2392:
Wow that's a weird bug. What's the extra segment being appended?

mne_bids/tests/test_write.py
```diff
@@ -1228,6 +1230,11 @@ def test_eegieeg(dir_name, fname, reader, _bids_validate, tmp_path):
                           match='Encountered data in "int" format. '
                           'Converting to float32.'):
             bids_output_path = write_raw_bids(**kwargs)
+    elif dir_name == 'EGI':
+        with pytest.warns(RuntimeWarning,
+                          match=r'Encountered data in "(float|single)" '
+                          'format. Converting to float32.'):
```
mscheltienne, line 1236: I'm not entirely sure if this is the correct way to use the match with a regex pattern, but it's required after https://github.com/mne-tools/mne-python/pull/10851 which fixes the `orig_format` for EGI datasets.

drammock: I think that will work?  You can also include a `.*` after the disjunct if you want the pattern to fit on one line.

mscheltienne: I didn't like the idea to match everything of whatever length ;)

scott-huberty:
> Wow that's a weird bug. What's the extra segment being appended?

Sorry that I didn't get to finish this during the code sprint - I've been in another intensive workshop this last week that runs into next, but can return to this when I'm finished. I'll let you know the details of the extra segment when I do. From what I remember it's called something like `New Segment` - I don't think it's a bug, it seems like it might be a common annotation that is added when reading brain vision files? (at least i've seen it [mentioned](https://github.com/mne-tools/mne-python/blob/6921ed20975ae0b37be447bb5d2a185262f08ae7/mne/io/brainvision/brainvision.py#L255-L258) in the `read_raw_brainvision()` code). That would explain why this annotation wasn't present in the `rawMFF` object, but is present in the BIDS Compliant `RawBrainVision` object. But I'll need to investigate it further.

adam2392:
> Sorry that I didn't get to finish this during the code sprint - I've been in another intensive workshop this last week that runs into next, but can return to this when I'm finished. I'll let you know the details of the extra segment when I do. From what I remember it's called something like `New Segment` - I don't think it's a bug, it seems like it might be a common annotation that is added when reading brain vision files? (at least i've seen it [mentioned](https://github.com/mne-tools/mne-python/blob/6921ed20975ae0b37be447bb5d2a185262f08ae7/mne/io/brainvision/brainvision.py#L255-L258) in the `read_raw_brainvision()` code). That would explain why this annotation wasn't present in the `rawMFF` object, but is present in the BIDS Compliant `RawBrainVision` object. But I'll need to investigate it further.

I see. Then it seems that perhaps we might not want to add the "New Segment" annotation to the BIDS files? Let us know when you're back and the team can see if we have any thoughts...

cc @sappelhoff who wrote pybv. and the brainvision part.

scott-huberty:
> > Sorry that I didn't get to finish this during the code sprint - I've been in another intensive workshop this last week that runs into next, but can return to this when I'm finished. I'll let you know the details of the extra segment when I do. From what I remember it's called something like `New Segment` - I don't think it's a bug, it seems like it might be a common annotation that is added when reading brain vision files? (at least i've seen it [mentioned](https://github.com/mne-tools/mne-python/blob/6921ed20975ae0b37be447bb5d2a185262f08ae7/mne/io/brainvision/brainvision.py#L255-L258) in the `read_raw_brainvision()` code). That would explain why this annotation wasn't present in the `rawMFF` object, but is present in the BIDS Compliant `RawBrainVision` object. But I'll need to investigate it further.
> 
> I see. Then it seems that perhaps we might not want to add the "New Segment" annotation to the BIDS files? Let us know when you're back and the team can see if we have any thoughts...
> 
> cc @sappelhoff who wrote pybv. and the brainvision part.

Sounds good. Btw, I was reading a [bit more into it,](https://pressrelease.brainproducts.com/markers/), and indeed `New Segment` markers are a BV convention, they represent discontinuities in EEG data (akin to boundary events in EEGlab) and the first marker in a BV recording is apparently always a `New Segment` marker. I think It will help once we better understand how `read_raw_brainvision` handles these, and I agree with you that we don't want these `New Segment` annotations being added to BIDS compliant files especially in cases where the original EEG recordings were not brain vision.

sappelhoff:
Adding the `New Segment` to **BrainVision files** markers at the beginning of a recording is required to encode the recording datetime into the file format. I think this is a benign change that is just one of many changes when converting from one data format into another. However if you all decide that this absolutely breaks everything, then you can skip writing that `New Segment` marker by passing `meas_date=None` in pybv, see the API: https://pybv.readthedocs.io/en/stable/generated/pybv.write_brainvision.html#pybv.write_brainvision

mscheltienne:
I'm surprised this issue popped up only now with the conversion from EGI to BV, and not earlier for all the other conversions.

scott-huberty:
> Adding the `New Segment` to **BrainVision files** markers at the beginning of a recording is required to encode the recording datetime into the file format. I think this is a benign change that is just one of many changes when converting from one data format into another. However if you all decide that this absolutely breaks everything, then you can skip writing that `New Segment` marker by passing `meas_date=None` in pybv, see the API: https://pybv.readthedocs.io/en/stable/generated/pybv.write_brainvision.html#pybv.write_brainvision

Thanks @sappelhoff , this was all so much more simple than I had gathered!

What do you all think? If the `New Segment` annotation is an intended part of the `pybv` writer, do you foresee it being a problem, if this annotation is included in the events sidecar file? 

On the one hand, we probably want the characteristics of the EEG file returned by `MNE-BIDS` to resemble (as much as possible) the raw EEG object the user provided (the same events,  etc), on the other hand, if this `New Segment` annotation  is necessary to assign the `meas_date`, I imagine that not having a `meas_date` in the BIDS compliant EEG file presents it's own issues.

mscheltienne:
I'm not super familiar with BIDS yet, is there not a sidecar field for the measurement date?

sappelhoff:
There is, which is why it would be okay to not encode the meas date in the raw data on top.

scott-huberty:
Just following up on this.

@mscheltienne confirming that anytime you use `write_raw_bids` to write a BV file, and then use `mne.io.read_raw_brainvision` to read the file back in, the raw object will have the `New Segment` annotation as the first annotation. This is not unique to EGI files that are converted to BV, even though this is the first time we noticed that `New Segment` was added.

However.. this `New_Segment` annotation only appears when you read in the BV file with `mne.io.read_raw_brainvision`. If you read in the BV file with `mne_bids.read_raw_bids`, the `New_Segment` Annotation isn't there.. I haven't figured out why.

Also, I followed @sappelhoff 's advice and tried setting `meas_date=None` in the `write_raw_brainvision` call inside the `mne_bids._write_raw_brainvision` code. With this change, the `New Segment` Annotation isn't appearing in the loaded raw object even when using `mne.io.read_raw_brainvision`.

This change brings the downside, that the RawBrainVision object will not have a `meas_date` set. However,  again, if you use `read_raw_bids` instead of `mne.io.read_raw_brainvision`, the `meas_date` will be properly set, by extracting the `meas_date` from the `scans.tsv` sidecar file.

mscheltienne:
Thanks for investigating. IMO, it makes more sense to set `meas_date=None`. This conversion is here to be compliant with the BIDS standard in the first place, thus a sidecar with the measurement date should be present and the use of `read_raw_bids` should be preferred.

mscheltienne:
I updated with the branch with the latest main, let's see if the tests are green.

adam2392:
> Thanks for investigating. IMO, it makes more sense to set `meas_date=None`. This conversion is here to be compliant with the BIDS standard in the first place, thus a sidecar with the measurement date should be present and the use of `read_raw_bids` should be preferred.

Agreed. BIDS standard is to prioritize first the BIDS sidecar metadata over that which is present in the raw recording file.

mscheltienne:
Next issue:

```
RuntimeWarning: Not setting positions of 7 stim channels found in montage:
E           ['DIN1', 'DIN2', 'DIN3', 'DIN4', 'DIN5', 'DIN7', 'STI 014']
E           Consider setting the channel types to be of EEG/sEEG/ECoG/DBS/fNIRS using inst.set_channel_types before calling inst.set_montage, or omit these channels when creating your montage.
```

Why does `read_raw_bids` emit this warning? In the `raw` written and it the `raw` read, the channels are correctly set as Stimulus in the `Info`. Does the BIDS has a different way to handle the `stim` channels and they should not be present in the written raw?

mscheltienne:
@scott-huberty I fixed other failures in the same test, please pull the changes if you keep digging into this :)

scott-huberty:
> Next issue:
> 
> ```
> RuntimeWarning: Not setting positions of 7 stim channels found in montage:
> E           ['DIN1', 'DIN2', 'DIN3', 'DIN4', 'DIN5', 'DIN7', 'STI 014']
> E           Consider setting the channel types to be of EEG/sEEG/ECoG/DBS/fNIRS using inst.set_channel_types before calling inst.set_montage, or omit these channels when creating your montage.
> ```
> 
> Why does `read_raw_bids` emit this warning? In the `raw` written and in the `raw` read, the channels are correctly set as `'stim'` in the `Info`. Does BIDS has a different way to handle the `stim` channels and they should not be present in the written raw?

Looks like you fixed it in the end? What was it? The channels that were listed in the warning, were of type "TRIG" in the channels sidecar file (for the BIDS compliant folder made from `test_egi.mff`, which was raising the warning?).

mscheltienne:
No, I did not, I fixed other issues 😅

scott-huberty:
> @scott-huberty I fixed other failures in the same test, please pull the changes if you keep digging into this :)

Ok Thanks! I'll keep hacking away at it this week

sappelhoff:
> Thanks for investigating. IMO, it makes more sense to set meas_date=None. This conversion is here to be compliant with the BIDS standard in the first place, thus a sidecar with the measurement date should be present and the use of read_raw_bids should be preferred.

Agreed :+1:

scott-huberty:
> Why does `read_raw_bids` emit this warning?

What is the BIDS standard for the `electrodes.tsv` sidecar file?  In this test case, stim channels are being written to `electrodes.tsv` during `write_raw_bids`. Then, during `read_raw_bids` `mne` warns (correctly, IMO) that these channels do not have any associated location and will be ignored during `set_montage`.

Isn't it the case that only eeg/meg channels should be written to `electrodes.tsv`?



When I use `mne-bids`, I usually do something like:
```
raw.pick_types(eeg=True, meg=True)  # exclude stim chans
```

Now, only EEG/MEG channels will be written to the electrodes sidecar file. Should mne-bids be doing something like this under the hood during `write_raw_bids`?

sappelhoff:
Only **electrodes** and NOT **channels** should be written to `electrodes.tsv`. For a disambiguation between these two, you can read the intro section here: https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#eeg-recording-data


> Isn't it the case that only eeg channels should be written to electrodes.tsv?

nope, all **electrode** should we written. This is also means that for bipolar channels, TWO rows should be written.

Having that said, I think STIM **channels** should not be written, because these do not correspond to electrodes (unless there is a veeeery very special setup being used somewhere).

hoechenberger:
> Having that said, I think STIM channels should not be written, because these do not correspond to electrodes 

Yes.

mscheltienne:
So, should we monkey patch the tests (either dropping the stim channel, or ignoring the warning) or should we prevent `write_raw_bids` from writing stimulation channels directly?
It looks to me like to save a `raw` to BIDS, you should extract events first and provide them as `events_data, event_id`:

```
raw = Raw()  # my raw with stim channels
events = mne.find_events()
write_raw_bids(raw, events_data=events, event_id=dict(...))
```

scott-huberty:
> So, should we monkey patch the tests (either dropping the stim channel, or ignoring the warning) or should we prevent `write_raw_bids` from writing stimulation channels directly?

If everyone agrees that stim channels shouldn't be written to `electrodes.tsv`, probably prevent `write_raw_bids` from writing the stim channels?

maybe something like:
```
if ch['kind'] == FIFF.FIFFV_STIM_CH:
    continue
```
at the beginning of this for loop?:
https://github.com/mne-tools/mne-bids/blob/54d50f61e01424081793db7ef3f68659319056cc/mne_bids/dig.py#L140-L152

sappelhoff:
looks like a reasonable suggestion to me! Though maybe it warrants its own issue and accompanying PR.

scott-huberty:
> looks like a reasonable suggestion to me! Though maybe it warrants its own issue and accompanying PR.

Sounds good! I will open an issue and begin working on a PR.

mscheltienne:
@hoechenberger @sappelhoff @adam2392 Alright, this PR is (finally) green!
I had to add a version check for MNE 1.1 in the tests following https://github.com/mne-tools/mne-python/pull/10851
Have a look and merge if happy :)

Thanks @scott-huberty for working on this one!

sappelhoff:
> And by the way, test_eegieeg was a pain to work with. IMO, It's way too long and testing too many aspects in the same function.

Yes :disappointed: 

- #212
- #678

sappelhoff (approved):
@anandsaini024 can you please leave a comment on #1028 with:

- your full name
- your ORCID (if you have one)
- your affiliation (if you have one)

mne_bids/tests/test_write.py
```diff
@@ -1491,10 +1506,13 @@ def test_eegieeg(dir_name, fname, reader, _bids_validate, tmp_path):
...
     elif dir_name == 'EGI':
-        with pytest.warns(RuntimeWarning,
-                          match=r'Encountered data in "(float|single)" '
-                          'format. Converting to float32.'):
+        if check_version("mne", "1.1"):
             write_raw_bids(**kwargs)
+        else:
+            with pytest.warns(RuntimeWarning,
+                              match=r'Encountered data in "float" '
+                              'format. Converting to float32.'):
```
hoechenberger, line 1514:
```suggestion
            with pytest.warns(
                RuntimeWarning,
                match=r'Encountered data in "float" format. '
                      r'Converting to float32.'
            ):
```

PR #1008, opened 2022-05-16
allow copyfile_brainvision to work with .dat extension

dominikwelke:
PR Description
--------------

hi all. 
the datafile in brainvision exchange format  either has the extension `.eeg` or `.dat`.
mne-bids' `copyfile_brainvision()` function can only handle `.eeg`, though, and throws an error otherwise.

this PR fixes this.

so far I left the new filenames with `.eeg` extension, because I didn't find out whether one option is preferred.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -73,7 +73,7 @@ def _get_brainvision_encoding(vhdr_file):
 
 
 def _get_brainvision_paths(vhdr_path):
-    """Get the .eeg and .vmrk file paths from a BrainVision header file.
+    """Get the .eeg/.dat and .vmrk file paths from a BrainVision header file.
 
     Parameters
     ----------
@@ -83,7 +83,7 @@ def _get_brainvision_paths(vhdr_path):
     Returns
     -------
     paths : tuple
-        Paths to the .eeg file at index 0 and the .vmrk file at index 1 of
+        Paths to the .eeg/.dat file at index 0 and the .vmrk file at index 1 of
         the returned tuple.
 
     """
@@ -100,10 +100,11 @@ def _get_brainvision_paths(vhdr_path):
     with open(vhdr_path, 'r', encoding=enc) as f:
         lines = f.readlines()
 
-    # Try to find data file .eeg
-    eeg_file_match = re.search(r'DataFile=(.*\.eeg)', ' '.join(lines))
+    # Try to find data file .eeg/.dat
+    eeg_file_match = re.search(r'DataFile=(.*\.(eeg|dat))', ' '.join(lines))
+    
     if not eeg_file_match:
-        raise ValueError('Could not find a .eeg file link in'
+        raise ValueError('Could not find a .eeg or .dat file link in'
                          f' {vhdr_path}')
     else:
         eeg_file = eeg_file_match.groups()[0]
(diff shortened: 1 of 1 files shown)


sappelhoff:
Hey @dominikwelke I have always wondered where the .dat files are coming from. Do you know?

dominikwelke:
hi @sappelhoff 

I know that at least eeglab creates `.dat` files if you export to brainvision format (also their BIDS tools).

..and I found this probably outdated brainvision spec sheet on the fieldtrip webpage that mentions `.dat` files: http://old.fieldtriptoolbox.org/_media/getting_started/brainvisioncorefileformat_1.0_2018-08-02.pdf

sappelhoff:
Okay, thanks - we should clarify this with eeglab and brain products then. I'll send them a message.

No objection against this PR though.

And I think unless there is a good reason, eeglab should write .eeg and not .dat

hoechenberger:
@dominikwelke Could you please add a changelog entry?
Otherwise this is looking good to me!

dominikwelke:
sure @hoechenberger  

bug fix or enhancement?

sappelhoff (approved):

agramfort:
thx a lot @dominikwelke

dominikwelke:
yes, sorry for being unprecise.

yes, the `pop_writebva()` function from the bva-io plugin stores data in `.dat` format.
this seems to be the standard plugin, though, if you want to export data in brain vision format.

sappelhoff:
okay thanks. I just emailed Arno to clarify this ... in the end it doesn't matter too much, but it's always good to know.

sappelhoff:
Here is their answer:

TLDR: `.dat` is a BrainVision analyzer export format ... I don't think it should be used too much

> Dear Stefan,
Thank you for contacting the Scientific Support team. I am glad to help you with the questions.
When recording using our hardware, BrainVision Recorder automatically assigns .eeg as extension for the EEG data file according to the [BrainVision Core file format](https://www.brainproducts.com/productdetails.php?id=21&tab=5), which has been officially adopted by BIDS-EEG. Among the three important points in compliance with BIDS-EEG standards are:
>- The BrainVision Core file format refers to raw data only (i.e. unprocessed data).
>- The EEG data file format is binary while the metadata (header and marker) are stored as text files.
>- The data file must have one of the following file extensions: “eeg”, “avg” or “seg”.
>
>On the other hand, when exporting data in BrainVision Analyzer, one has the choice in Generic Data Export module - the field specifying the extension is editable and one can decide if one wants to use .dat or .eeg. In addition to Generic Data Export, one could also use Create New Dataset module to export data. But, this creates EEG and metadata files that are not complying to BIDS-EEG standards (i.e., .dat, .xhdr and .xmrk). 
So, .dat extension is one of the BrainVision format, it is not outdated, it is simply a freely selectable option for the user when exporting the data, but it does not meet the specification of BIDS-EEG standards. 
We hope the provided information clarifies your doubts. But if you have further concerns, please feel free to get back to us.

hoechenberger:
So we cannot guarantee that .dat is actually readable by our common BV readers?

sappelhoff:
indeed, that's how I read the message. And it reinforces my opinion to handle .dat with care ...

hoechenberger:
My concern is that we're allowing the creation of something that is not BIDS-compliant. We should at least emit a warning, don't you think?

sappelhoff:
we are still writing as eeg vhdr vmrk with the changes here. We basically simply allow that a "dat" is passed (but it'll be converted to eeg)

A warning wouldn't hurt. And perhaps a quick check whether reading the file with mne is possible.

hoechenberger:
Ah ok. Sorry about the confusion then.

hoechenberger:
Yes okay, since it's meant for derivatives, maybe we don't even need to warn (yet). No clue. :)

sappelhoff:
> Yes okay, since it's meant for derivatives, maybe we don't even need to warn (yet). No clue. :)

only for the case when somebody wants to read `.dat` and write to `.eeg` for raw bids data :shrug: but yes, if we assert that reading the data is possible (`preload=False`), then that may be enough of a verification that we don't get really weird data

PR #1010, opened 2022-05-24
follow up: allow copyfile_brainvision to work with .dat extension

dominikwelke:
PR Description
--------------

this PR asserts that brainvision files can be read by MNE

follow up for #1008 

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -313,8 +313,8 @@ def copyfile_brainvision(vhdr_src, vhdr_dest, anonymize=None, verbose=None):
     fname_src, ext_src = _parse_ext(vhdr_src)
     fname_dest, ext_dest = _parse_ext(vhdr_dest)
     if ext_src != ext_dest:
-        raise ValueError('Need to move data with same extension, '
-                         ' but got "{ext_src}" and "{ext_dest}"')
+        raise ValueError(f'Need to move data with same extension, '
+                         f' but got "{ext_src}" and "{ext_dest}"')
 
     eeg_file_path, vmrk_file_path = _get_brainvision_paths(vhdr_src)
 
@@ -322,8 +322,8 @@ def copyfile_brainvision(vhdr_src, vhdr_dest, anonymize=None, verbose=None):
     try:
         _ = read_raw_brainvision(vhdr_src, preload=False, verbose=None)
     except BaseException:
-        raise ImportError(f'The specified EEG data cannot be loaded by MNE '
-                          f'- it may be corrupted')
+        raise ImportError('The specified EEG data cannot be loaded by MNE '
+                          '- it may be corrupted')
 
     # extract encoding from brainvision header file, or default to utf-8
     enc = _get_brainvision_encoding(vhdr_src)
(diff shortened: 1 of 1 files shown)


dominikwelke:
hi @sappelhoff @hoechenberger 
here is a draft for the requested assertion.

if you prefer some other place let me know. 
e.g. it could also become a higher level assertion applied to all files (not only brainvision)

I am not convinced though, that this will catch all cases. for example when writing the code i renamed a random `.json` file to replace the binary `.eeg` file in an example dataset, and `mne` read if without any errors

..and i didnt raise a warning so far, let me know if you want that

sappelhoff (commented):
The `.eeg` file remains untouched by MNE, if you pass `preload=False` --> so this is to be expected in this case.

I think this check is good to have, it will give us some safety that at least the vhdr and vmrk can be properly read.

I think a warning if ".dat" is copied may be good -- especially because we are rewriting it as ".eeg", right? And then maybe some info along the lines: ".dat" is not necessarily raw data, please double check that everything is fine.

not sure if we should add this to all data formats -- but for BrainVision we have a good reason to do so IMHO

thanks!

dominikwelke:
> The `.eeg` file remains untouched by MNE, if you pass `preload=False` --> so this is to be expected in this case.

i tried with `preload=True`

sappelhoff:
oh, that's weird, did you also rename the file references in VHDR and VMRK (what bv_copyfile does)?

dominikwelke:
yes, weird.. you can reproduce if you want :)

[Archive.zip](https://github.com/mne-tools/mne-bids/files/8764355/Archive.zip)
(it's simply the `*_eeg.json` renamed to `*_eeg.eeg`)

it reads the textfile as if it were an eeg file, and `raw.get_data()` returns an array with one datapoint per channel.
this is obviously a problem of base mne, if a *problem* at all (in the end one should trust the users that they provide actual data, and dont simply rename random files 😃  )



in this context it's relevant, though:
if we dont really assert the integrity of the data, i dont really see the point in including an assertion at all.. basically the way it is (`preload=False`) only checks if the header is meaningfull, but not the whole set..

what do you think?

sappelhoff:
wow, I can reproduce -- that's crazy. probably because MNE just opens that json file as if it was binary and is then reading those bytes interpreting them as numbers, or something like it?

Maybe not an important problem, but something that would be nice to have fixed with a good error message. :shrug:

I think you are right then, we might as well close this PR and hope for the best / wait for user reports if people run into issues. (I hope not)

dominikwelke:
> wow, I can reproduce -- that's crazy. probably because MNE just opens that json file as if it was binary and is then reading those bytes interpreting them as numbers, or something like it?

this is what i thought, too. 
i didnt do a lot of testing with other files.. not surprisingly an empty textfile renamed to `.eeg` doesnt work. (some general error along "couldnt read line 1")

> Maybe not an important problem, but something that would be nice to have fixed with a good error message. 🤷

I'll create an issue in the MNE repo soon, at least people are aware then
not sure whether it's really possible to catch this, though.  

> I think you are right then, we might as well close this PR and hope for the best / wait for user reports if people run into issues. (I hope not)

ok with me.. 
but actually I only meant the assertion. raising a warning if `.dat` is provided would still make sense IMO (high likelihood that the files have been manipulated, and might not adhere to BIDS standards). i didnt have much time last week, but will create a commit soon

dominikwelke:
i changed it to the warning @sappelhoff 

dont think the ubuntu CIs are unhappy due to my PR?

sappelhoff (approved):

sappelhoff:
thanks!

PR #1030, opened 2022-07-30
MRG: In BIDSPath, don't infer the datatype from the suffix

hoechenberger:
I needed to set datatype to None to make `find_emptyroom()` deal with the inheritance principle correctly, but turns out it never took effect because `BIDSPath.update()` would silently replace my `datatype=None` with an inferred `datatype='meg'`.

This PR stops this automated inference.

This will likely break user code, but it's the right thing to do as we've been trying to be too clever in the past.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -48,7 +48,7 @@ Detailed list of changes
 🧐 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- In many places, we used to infer the `datatype` of a :class:`~mne_bids.BIDSPath` from the `suffix`, if not explicitly provided. However, this has lead to trouble in certain edge cases. In an effort to reduce the amount of implicit behavior in MNE-BIDS, we now require users to explicitly specify a `datatype` whenever the invoked functions or methods expect one, by `Richard Höchenberger`_ (:gh:`1030`)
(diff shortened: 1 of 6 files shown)


hoechenberger:
This PR is a blocker for https://github.com/mne-tools/mne-bids-pipeline/pull/576

doc/whats_new.rst
```diff
@@ -48,6 +48,8 @@ Detailed list of changes
 🧐 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- In many places, we used to infer the `datatype` of a :class:`~mne_bids.BIDSPath` from the `suffix`, if not explicitly provided. However, this has lead to trouble in certain edge cases. In an effort to reduce the amount of implicit behavior in MNE-BIDS, we now require users to explicitly specify a `datatype` whenever the invoked functions or methods expect one, by `Richard Höchenberger`_ (:gh:`1030`)
```
hoechenberger, line 51:
```suggestion
- In many places, we used to infer the ``datatype`` of a :class:`~mne_bids.BIDSPath` from the ``suffix``, if not explicitly provided. However, this has lead to trouble in certain edge cases. In an effort to reduce the amount of implicit behavior in MNE-BIDS, we now require users to explicitly specify a ``datatype`` whenever the invoked functions or methods expect one, by `Richard Höchenberger`_ (:gh:`1030`)
```

mne_bids/path.py
```diff
@@ -1237,14 +1233,15 @@ def _parse_ext(raw_fname):
     return fname, ext
 
 
-def _infer_datatype_from_path(fname):
+def _infer_datatype_from_path(fname: Path):
```
hoechenberger, line 1236: this function is used by `get_bids_path_from_fname()`

IMHO it's doing too much magic too, but here I only ensured it keeps on working

agramfort (approved):

agramfort:
Thanks @hoechenberger

PR #1023, opened 2022-07-28
amended `_write_electrodes_tsv` to exclude writing stim channels to electrodes.tsv

scott-huberty:
closes #1022 

`make tests` and `make pep` passed on my local machine, fingers crossed!

Issue #485, opened 2020-07-27
Bipolar electrode recordings and "electrodes.tsv"


sappelhoff:
When working on the impedances PR, I was working with data that included EOG and ECG *channels*.

Each of these *channels* is coming from a set of *two* electrodes with bipolar referencing scheme.

E.g. `ECG` = `ECG+` - `ECG-`.

In BIDS, the `electrodes.tsv` file is there to record *electrodes*, not *channels*. Yet, MNE-BIDS writes bipolar channels as a single row to `electrodes.tsv`.

Furthermore, Ground and Reference would each need a row in `electrodes.tsv` as well.

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv

I am aware that this is really hard to automate ... the reasons for that are twofold as far as I can see:

1. MNE-Python has no representation of how each channel was referenced
1. Many data formats "drop" information about Ground and Reference electrodes before saving the data

agramfort:
some hints to see how mne-python still can help.

https://mne.tools/dev/generated/mne.set_eeg_reference.html
https://mne.tools/dev/generated/mne.set_bipolar_reference.html
<https://mne.tools/dev/generated/mne.set_bipolar_reference.html#mne.set_bipolar_reference>

I am pretty sure that some references choices can be undone.

in chs['loc'] the first three value is the position of the electrode and
the next 3 the position
of the reference electrode.

In the FIF format EEG000 is the dig is the reference electrode

maybe this can help

hoechenberger:
@sappelhoff In BIDS, is there a way to define which electrodes were combined (and how) to form a channel in `channels.tsv`, and to do that in an automatable?

For a bipolar channel created from two unipolar electrodes, how you would specify the reference in `channels.tsv`?

The spec is very unclear on this:
<img width="821" alt="Screen Shot 2021-02-10 at 22 14 09" src="https://user-images.githubusercontent.com/2046265/107573239-8bb13600-6bed-11eb-8d40-c9e686023f9d.png">

In your example with the two EOG channels, one would need to create an entry in `channels.tsv` that probably looks something like this:

```
name  type  units  reference              status  status_description
ECG   ECG   mV     bipolar: ECG- - ECG+   good    n/a
```
But this is not mentioned in the specs.
-> As I see it, there is currently no way in BIDS to specify a bipolar reference scheme per channel.

hoechenberger:
[...]

PR #1032, opened 2022-07-30
[MRG, DOC] Fix whats_new.rst link

mscheltienne:
Spotted in #1006

```
RuntimeWarning: Not setting position of 1 emg channel found in montage:
E           ['EMG']
E           Consider setting the channel types to be of EEG/sEEG/ECoG/DBS/fNIRS using inst.set_channel_types before calling inst.set_montage, or omit these channels when creating your montage.
```

If I read the [BIDS specification](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#eeg-recording-data) correctly, a bipolar channel such as EOG, ECG, EMG should not be written to the sidecar:

> Channel: [...] This can be connected to two electrodes (to measure the potential difference between them)

I amended the fix for `stim` channels in #1023 to include also `eog`, `ecg` and `emg` channels.

Also, I'm surprised the doc-build did not complain in #1023, in `whats_new.rst` the `_` was badly placed. I fixed it here and added myself and @scott-huberty to the `authors.rst` file.

### DIFF
--- a/doc/authors.rst
+++ b/doc/authors.rst
@@ -33,3 +33,5 @@
 .. _Simon Kern: https://github.com/skjerns
 .. _Yorguin Mantilla: https://github.com/yjmantilla
 .. _Swastika Gupta: https://swastyy.github.io
+.. _Mathieu Scheltienne: https://github.com/mscheltienne
+.. _Scott Huberty: https://github.com/scott-huberty
--- a/doc/authors.rst
+++ b/doc/authors.rst
@@ -33,5 +33,4 @@
 .. _Simon Kern: https://github.com/skjerns
 .. _Yorguin Mantilla: https://github.com/yjmantilla
 .. _Swastika Gupta: https://swastyy.github.io
-.. _Mathieu Scheltienne: https://github.com/mscheltienne
 .. _Scott Huberty: https://github.com/scott-huberty
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -72,7 +72,9 @@ Detailed list of changes
 
 - :func:`~mne_bids.print_dir_tree` now correctly expands ``~`` to the user's home directory, by `Richard Höchenberger`_ (:gh:`1013`)
 
-- :func:`~mne_bids.write_raw_bids` now correctly excludes stim channels when writing to electrodes.tsv, by `Scott Huberty_` (:gh:`1023`)
+- :func:`~mne_bids.write_raw_bids` now correctly excludes stim channels when writing to electrodes.tsv, by `Scott Huberty`_ (:gh:`1023`)
+
+- :func:`~mne_bids.write_raw_bids` now correctly excludes eog, ecg, emg channels when writing to electrodes.tsv, by `Mathieu Scheltienne`_ (:gh:`1032`)
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -76,8 +76,6 @@ Detailed list of changes
 
 - :func:`~mne_bids.write_raw_bids` now correctly excludes stim channels when writing to electrodes.tsv, by `Scott Huberty`_ (:gh:`1023`)
 
-- :func:`~mne_bids.write_raw_bids` now correctly excludes eog, ecg, emg channels when writing to electrodes.tsv, by `Mathieu Scheltienne`_ (:gh:`1032`)
-
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
 .. include:: authors.rst
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -48,6 +48,8 @@ Detailed list of changes
 🧐 API and behavior changes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- In many places, we used to infer the ``datatype`` of a :class:`~mne_bids.BIDSPath` from the ``suffix``, if not explicitly provided. However, this has lead to trouble in certain edge cases. In an effort to reduce the amount of implicit behavior in MNE-BIDS, we now require users to explicitly specify a ``datatype`` whenever the invoked functions or methods expect one, by `Richard Höchenberger`_ (:gh:`1030`)
+
 - :func:`mne_bids.make_dataset_description` now accepts keyword arguments only, and can now also write the following metadata: ``HEDVersion``, ``EthicsApprovals``, ``GeneratedBy``, and ``SourceDatasets``, by `Stefan Appelhoff`_ (:gh:`406`)
 
 - The deprecated function ``mne_bids.mark_bad_channels`` has been removed in favor of :func:`mne_bids.mark_channels`, by `Richard Höchenberger`_ (:gh:`1009`)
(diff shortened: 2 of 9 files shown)


hoechenberger:
> Also, I'm surprised the doc-build did not complain

I had an issue with my docs in #1030 as well and the build passed. Something is off.

mscheltienne:
Yep, in #1023 it rendered [this page](https://output.circle-artifacts.com/output/job/52a106e2-5670-4a60-bf97-ec4f86b6b82f/artifacts/0/dev/whats_new.html?highlight=what+s+new
) without complaining about the syntax or the missing link.

image

hoechenberger:
> If I read the [BIDS specification](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#eeg-recording-data) correctly, a bipolar channel such as EOG, ECG, EMG should not be written to the sidecar:

Where did you read this?

Electrode positions should be written to `electrodes.tsv`. Omitting an EOG electrode just because we don't handle it properly during reading is not the right solution here … if we do have EOG electrode positions, they should be written to `electrodes.tsv`

mscheltienne:
The distinction between `Electrodes` or `Channel`:

> 
>     Electrode = A single point of contact between the acquisition system and the recording site (for example, scalp, neural tissue, ...). Multiple electrodes can be organized as caps (for EEG), arrays, grids, leads, strips, probes, shafts, and so on.
> 
>     Channel = A single analog-to-digital converter in the recording system that regularly samples the value of a transducer, which results in the signal being represented as a time series in the digitized data. This can be connected to two electrodes (to measure the potential difference between them), a magnetic field or magnetic gradient sensor, temperature sensor, accelerometer, and so on.

An EOG / ECG / EMG is not comprised of a single point of contact between the acquisition system and the recording site. Usually, it's a bipolar channel connected to 2 electrodes. And it does not have a location associated with it. IMO, those 2 points make it fall under the `Channel` category.

hoechenberger:
> An EOG / ECG / EMG is not comprised of a **single** point of contact between the acquisition system and the recording site. Usually, it's a bipolar channel connected to 2 electrodes.

I disagree; a common scheme is to e.g. place an EOG electrode below the eye and use Fp1/Fp2 as the bipolar counterpart. If you have measured a location for those electrodes, they should end up in electrodes.tsv

I think you're confusing electrodes and channels here.

hoechenberger:
The specs also specifically mention the bipolar case:

> Note that electrode positions SHOULD NOT be added to [channels.tsv], but to [*_electrodes.tsv](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv). Furthermore, the entries in *_electrodes.tsv and *_channels.tsv do not have to match exactly, as for example in the case of recording a single EOG channel from a bipolar referencing scheme of two electrodes, or a data channel originating from an auxiliary, non-electrode device. That is, in most cases *_electrodes.tsv will have more entries than *_channels.tsv. See the examples for *_channels.tsv below, and for *_electrodes.tsv in ["Electrodes description"](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv).

hoechenberger:
In fact, their example even suggests to add EOG electrodes to electrodes.tsv even if you **don't** have the locations:

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#example-electrodestsv

Screen Shot 2022-07-30 at 13 48 56


So I'm -1 on this change here

mscheltienne:
Then what about ECG and EMG? I think all 3 should be treated in the same way: they measure muscle activity via a bipolar layout. 

With ECG / EMG, an electrode from the EEG layout will (probably?) never be used as a bipolar counterpart, while for EOG it is indeed classic to use either Fp1/Fp2 as one of the counterparts, or even other electrodes. e.g. the ANT Neuro recent caps have a droplead electrode for EOG which is referenced to the same electrode as all the EEG electrodes: CPz (you can re-reference that channel to Fp1/Fp2 in your processing).

But in the end, it is still a channel with a bipolar layout, which does not measure brain activity but muscle activity, and which does not have one specific position. If for instance my EOG channel is not using Fp1/Fp2, but 2 electrodes on a bipolar aux channel, we end up in the case where it should not be included in the `electrodes.tsv` because the channel is comprised of 2 electrodes with 2 distinct locations. While if you use Fp1/Fp2, you can cheat and include the EOG channel because the bipolar counterpart is part of the EEG layout. 

My point of view as a light-BIDS user ;) Let me know and I'll adapt accordingly. I'll return to this tomorrow or on Monday.

sappelhoff:
this is related to https://github.com/mne-tools/mne-bids/issues/485

hoechenberger:
> this is related to:
> 
> * [Bipolar electrode recordings and "electrodes.tsv" #485](https://github.com/mne-tools/mne-bids/issues/485)

Hah, I already had strong opinions on this entire mess a year ago:

https://github.com/mne-tools/mne-bids/issues/485#issuecomment-777374523

hoechenberger:
@mscheltienne 
> If for instance my EOG channel is not using Fp1/Fp2, but 2 electrodes on a bipolar aux channel, we end up in the case where it should not be included in the `electrodes.tsv` because the channel is comprised of 2 electrodes with 2 distinct locations.

I don't see why it shouldn't be included?

I don't think the specs say, "must only include electrodes that measure brain activity". It's **any** electrodes that go into this file…

sappelhoff:
without having read everything in this thread, my opinion is:

- a raw data file contains data from **channels**, these **channels** should all be listed in `channels.tsv` ... with the corresponding order in sidecar and raw file
- a given **channel** may correspond conceptually to
    - a single electrode (e.g., when referenced to a common reference, we often name the electrode and channel the same)
    - two electrodes (e.g., we often name an EOG channel "HEOG" if it's referenced between an electrode to the outer 
canthus of the left eye ("HEOG-") and another electrode to the outer canthus of the right eye ("HEOG+"))
    - note: I know that conceptually, each channel *always* corresponds to two electrodes (this is how we measure voltages), but it's common practice to "ignore" this when there is a common reference for all channels (such as a single electrode, or the common average reference)
- all **electrodes** should be listed in `electrodes.tsv`
    - this includes ground, ref, bipolar electrodes, etc.
    - thus, `electrodes.tsv` most likely has more entries than `channels.tsv`

--> writing `channels.tsv` in an automated fashion is easy.
--> writing `electrodes.tsv` in an automated fashion is hard, due to the way we represent data in our analysis software (at least in mne-python ... don't know about fieldtrip, eeglab, etc.)

We should do our best in mne-bids, and then write guides for the stuff we cannot easily automate.

hoechenberger:
Idea on how to move forward here:

In addition to all (i)EEG **channels**, Write all **channels** for which we have locations to electrodes.tsv. 

My thinking is that if a **channel** has a location, then it has a unique physical location and hence it's safe to assume it's an **electrode**

Thoughts?

mscheltienne:
Thank you for the clarification. That distinction between `channels.tsv` and `electrodes.tsv` is clear now (I knew well the `electrodes.tsv` file, but had forgotten the existence of `channels.tsv`). It does look difficult to automate the writing of the `electrodes.tsv` file from an `mne.Info` alone.

I'll revert the changes in this PR and will keep the doc fix for #1023.
And for #1006, I'll add an ignore for the warning.

sappelhoff (approved):

adam2392 (approved):

hoechenberger:
Thanks @mscheltienne

PR #1034, opened 2022-08-02
[WIP] Improve check of scans.tsv for BrainVision files

### LINKED ISSUE #1033: [BUG] parsing scans.tsv is currently too restrictive for BrainVision files
#### Describe the bug
After reading through the BIDS spec for the `_scans.tsv` to check that this is indeed an error, it doesn't specify which filename should be used in the tsv. This is particularly ambiguous when it comes to BrainVision where the data are stored across three different files: `.vhdr`, `.vmrk`, and `.eeg`. By default, mne-bids assumes the file in the `_scans.tsv` will be `.vhdr`, but it could be `.eeg` or `.vmrk` as well. We currently only test for whether it's `.vhdr` [here](https://github.com/mne-tools/mne-bids/blob/main/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/sub-01_ses-eeg_scans.tsv#L2). 

However, after reading a BIDS dataset, we found that it could also have the `.eeg` as seen [here](https://openneuro.org/datasets/ds003688/versions/1.0.7/file-display/sub-45:ses-iemu:sub-45_ses-iemu_scans.tsv).

teonbrooks:
PR Description
--------------

This PR closes #1033.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -92,7 +92,7 @@ Detailed list of changes
 
 - Instead of deleting files and raising cryptic errors, an intentional error message is now sent when calling :func:`~mne_bids.write_raw_bids` with the source file identical to the destination file, unless ``format`` is specified, by `Adam Li`_ and `Stefan Appelhoff`_ (:gh:`889`)
 
-- Internal helper function to :func:`~mne.read_raw_bids` would reject BrainVision data if `_scans.tsv` listed a `.eeg` file instead of `.vhdr`, by `Teon Brooks`_ (:gh:`1034`)
+- Internal helper function to :func:`~mne_bids.read_raw_bids` would reject BrainVision data if ``_scans.tsv`` listed a ``.eeg`` file instead of ``.vhdr``, by `Teon Brooks`_ (:gh:`1034`)
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -92,6 +92,8 @@ Detailed list of changes
 
 - Instead of deleting files and raising cryptic errors, an intentional error message is now sent when calling :func:`~mne_bids.write_raw_bids` with the source file identical to the destination file, unless ``format`` is specified, by `Adam Li`_ and `Stefan Appelhoff`_ (:gh:`889`)
 
+- Internal helper function to :func:`~mne.read_raw_bids` would reject BrainVision data if `_scans.tsv` listed a `.eeg` file instead of `.vhdr`, by `Teon Brooks`_ (:gh:`1034`)
+
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
 .. include:: authors.rst
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -261,7 +261,7 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
                              r'split-\d+_' + bids_path.datatype +
                              bids_path.fpath.suffix)
         split_fnames = list(filter(
-            lambda x: pattern.match(str(x)), fnames
+            lambda x: pattern.match(x.as_posix()), fnames
         ))
         split_acq_times = []
         for split_f in split_fnames:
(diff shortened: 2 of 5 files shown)


mne_bids/read.py
```diff
@@ -234,8 +234,8 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
     scans_tsv = _from_tsv(scans_fname)
     fname = bids_path.fpath.name
 
-    if fname.endswith('.pdf'):
-        # for BTI files, the scan is an entire directory
+    # for BTI and BrainVision files, the scan is an entire directory
+    if fname.split('.')[1] in ('pdf', 'vhdr'):
```
agramfort, line 238: I would use os.path.splitext or Path.suffix with pathlib

mne_bids/read.py
```diff
@@ -248,6 +248,10 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
     else:
         acq_times = ['n/a'] * len(fnames)
 
+    # There are three possible extensions for BrainVision
+    if fname.split('.')[1] in ('vhdr', 'eeg', 'vmrk'):
```
agramfort, line 252: you don't like my splitext suggestion @teonbrooks ?

teonbrooks: yes yes, I missed the review when I pushed. needed to do some other maneuvering. let me know what you think

teonbrooks: also, i just learned great deal about pathlib :)

agramfort (commented):
would it be possible to add a test or is it too much to ask?

teonbrooks:
> would it be possible to add a test or is it too much to ask?

@agramfort, yep, i can do that. i renamed it to wip, and i'll write a test in the morning. noticed this at neurohackademy when working through an ieeg example

mne_bids/tests/test_read.py
```diff
@@ -56,6 +57,10 @@
 # Data with cHPI info
 raw_fname_chpi = op.join(data_path, 'SSS', 'test_move_anon_raw.fif')
 
+# Tiny BIDS testing dataset
+mne_bids_root = os.sep.join(mne_bids.__file__.split("/")[:-2])
```
sappelhoff, line 61:
```suggestion
mne_bids_root = os.sep.join(mne_bids.__file__.split(os.sep)[:-2])
```

doc/whats_new.rst
```diff
@@ -92,6 +92,8 @@ Detailed list of changes
 
 - Instead of deleting files and raising cryptic errors, an intentional error message is now sent when calling :func:`~mne_bids.write_raw_bids` with the source file identical to the destination file, unless ``format`` is specified, by `Adam Li`_ and `Stefan Appelhoff`_ (:gh:`889`)
 
+- Internal helper function to :func:`~mne.read_raw_bids` would reject BrainVision data if `_scans.tsv` listed a `.eeg` file instead of `.vhdr`, by `Teon Brooks`_ (:gh:`1034`)
```
sappelhoff, line 95:
```suggestion
- Internal helper function to :func:`~mne_bids.read_raw_bids` would reject BrainVision data if ``_scans.tsv`` listed a ``.eeg`` file instead of ``.vhdr``, by `Teon Brooks`_ (:gh:`1034`)
```

sappelhoff (commented):
looks like you have some Windows issue:

`ValueError: 'eeg\\sub-01_ses-01_task-testing_acq-01_run-01_eeg.vhdr' is not in list`

probably `as_posix` needs to be called on the Path somewhere, to normalize to `/` instead of `\` on win.

teonbrooks:
> looks like you have some Windows issue:
> 
> `ValueError: 'eeg\\sub-01_ses-01_task-testing_acq-01_run-01_eeg.vhdr' is not in list`
> 
> probably `as_posix` needs to be called on the Path somewhere, to normalize to `/` instead of `\` on win.

so i'm looking into this and it seems like we read in the tsv file but we don't change filename from str to a filepath obj that would be insensitive to the os. should we be working with filepath all over the place?

also, if you see an easy fix, please let me know, i keep bumping into different windows errors on the testing front. not sure if i have any more time to debug it at the moment

agramfort:
ok all good !

CI failure with doc is also in main branch

agramfort:
thanks @teonbrooks !

PR #1052, opened 2022-08-19
MRG: Set channel type to misc if we cannot perform a proper conversion from BIDS to MNE channel types

### LINKED ISSUE #1048: Don't silently apply the BIDS datatype's channel type to channels for which we don't have a proper BIDS -> MNE mapping
_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids/issues/1046#issuecomment-1210231693_

We currently use the following channel type mapping from BIDS to MNE:
https://github.com/mne-tools/mne-bids/blob/16a7b0c1a4348256f6781f539cd82c0acf444de3/mne_bids/utils.py#L77-L83

We need to add `TEMP` and `GSR` here.

I don't believe MNE has "proper" support for these, so we should map them to `misc`.
@larsoner @agramfort WDYT?

@sappelhoff I don't think we should silently apply the BIDS datatype's channel type to channels for which we don't have a proper BIDS -> MNE mapping defined; instead, we should default to `misc`. WDYT?

hoechenberger:
Previously, e.g. GSR and temperature channels could end up as "eeg" channels, for example.

Fixes #1048

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [x] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -94,6 +94,8 @@ Detailed list of changes
 
 - Internal helper function to :func:`~mne_bids.read_raw_bids` would reject BrainVision data if ``_scans.tsv`` listed a ``.eeg`` file instead of ``.vhdr``, by `Teon Brooks`_ (:gh:`1034`)
 
+- Whenever :func:`~mne_bids.read_raw_bids` encounters a channel type that currently doesn't translate into an appropriate MNE channel type, the channel type will now be set to ``'misc``. Previously, seemingly arbitrary channel types would be applied, e.g. ``'eeg'`` for GSR and temperature channels, by `Richard Höchenberger`_ (:gh:`1052`)
+
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
 .. include:: authors.rst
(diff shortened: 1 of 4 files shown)


doc/whats_new.rst
```diff
@@ -94,6 +94,8 @@ Detailed list of changes
 
 - Internal helper function to :func:`~mne_bids.read_raw_bids` would reject BrainVision data if ``_scans.tsv`` listed a ``.eeg`` file instead of ``.vhdr``, by `Teon Brooks`_ (:gh:`1034`)
 
+- Whenever :func:`~mne_bids.read_raw_bids` encounters a channel type that currently doesn't translate into an appropriate MNE channel type, the channel type will now be set to ``'misc``. Previously, seemingly arbitrary channel types would be applied, e.g. ``'eeg'`` for GSR and temperature channels, by `Richard Höchenberger`_ (:gh:`xxx`)
```
hoechenberger, line 97:
```suggestion
- Whenever :func:`~mne_bids.read_raw_bids` encounters a channel type that currently doesn't translate into an appropriate MNE channel type, the channel type will now be set to ``'misc``. Previously, seemingly arbitrary channel types would be applied, e.g. ``'eeg'`` for GSR and temperature channels, by `Richard Höchenberger`_ (:gh:`1052`)
```

agramfort (approved):

PR #1059, opened 2022-08-28
MRG: Add support for GSR and temperature channels

### LINKED ISSUE #1046: BIDS channel type issue
bids_path = BIDSPath(root='../ds003061')

Channels 65 to 79 are not EEG (and marked as such in the BIDS repo). Yet MNE thinks 2 of these channels are EEG.

https://openneuro.org/datasets/ds003061

hoechenberger:
Also updates tiny_bids

Will fail until https://github.com/mne-tools/mne-python/pull/11108 has been merged

Fixes #1046

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [x] All CIs are happy


hoechenberger:
@sappelhoff I don't think I'm using tiny_bids correctly. I don't understand why the idea is that it should overwrite itself (?) upon (re)-creation

Could you take a look please? I only wanted to add GSR and temperature channels…

sappelhoff:
I don't see the problem :thinking: You'd have to edit `tiny_bids/code/make_tiny_bids_dataset.py` (like you did) and then re-run it, to update the dataset. And then you'd simply commit the changes. This makes it much more easy to keep `tiny_bids` in line with what mne-bids currently produces -- instead of manually editing the data files in `tiny_bids`.

hoechenberger:
@sappelhoff But look at the diff for the VHDR file, it seems
we're actually losing entire sections, no?

If that's ok with you, then there is no problem for me either :)

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_channels.tsv
```diff
@@ -67,4 +67,4 @@ ECG	ECG	µV	0.015915494309189534	1000.0	ElectroCardioGram	5000.0	good	n/a
 HEOG	EOG	µV	0.015915494309189534	1000.0	ElectroOculoGram	5000.0	good	n/a
 VEOG	EOG	µV	0.015915494309189534	1000.0	ElectroOculoGram	5000.0	good	n/a
 GSR	GSR	S	0.015915494309189534	1000.0	Galvanic skin response (electrodermal activity, EDA)	5000.0	good	n/a
-Temperature	TEMP	C	0.015915494309189534	1000.0	Other type of channel	5000.0	good	n/a
+Temperature	TEMP	oC	0.015915494309189534	1000.0	Other type of channel	5000.0	good	n/a
```
hoechenberger, line 70: need to add a test for this

sappelhoff (commented):
ahh yes, this is because apparently when the code was run earlier, we used `copyfile_brainvision` ... and now we use `pybv` (see also diff of first line in vhdr) so some metadata gets lost.

hoechenberger:
Wo should I "repair" this manually? WDYT?

sappelhoff:
I don't think the "lost" metadata is needed, so no problem from my side.

I am rather interested in why now pybv is used instead of copyfile_brainvision. :-) 

Maybe I find some time later to check.

hoechenberger:
> I am rather interested in why now pybv is used instead of copyfile_brainvision. :-)

Oh this one is easy, because I load the data, then append the GSR and temperature channels, then save the modified file. So it cannot be copied

Which now makes me wonder … if we run the script for tiny_bids creation multiple times, will we start to accumulate GSR and temperature sensors? :D

sappelhoff:
ahh yes, so you need some check like

`if my_gsr_ch_ not in channels: append channel ... else not`

hoechenberger:
@agramfort @sappelhoff This is good to be merged from my end. CI still running but all should work.

You can try it out:

```python
# %%
from pathlib import Path
import openneuro
import mne
import mne_bids

dataset_name = 'ds003061'
bids_root = Path('/tmp') / dataset_name

# %%
openneuro.download(
    dataset=dataset_name,
    include='sub-001',
    target_dir=bids_root,
)

# %%
bp = mne_bids.BIDSPath(
    subject='001',
    task='P300',
    run='1',
    datatype='eeg',
    suffix='eeg',
    extension='.set',
    root=bids_root,
)
raw = mne_bids.read_raw_bids(bp)

for ch_name in raw.ch_names:
    ch_type = raw.get_channel_types(ch_name)[0]
    ch_idx = raw.ch_names.index(ch_name)
    print(f'{ch_idx:<5} {ch_name:<6} {ch_type}')
```
Screen Shot 2022-08-29 at 17 43 47

doc/whats_new.rst
```diff
@@ -51,6 +51,8 @@ Detailed list of changes
 
 - TSV files that are empty (i.e., only a header row is present) are now handled more robustly and a warning is issued, by `Stefan Appelhoff`_ (:gh:`1038`)
 
+- Added support for ``GSR`` (galvanic skin response / electrodermal activity, EDA) and ``TEMP`` (temperaute) channel types, by `Richard Höchenberger`_ (:gh:`xxx`)
```
sappelhoff, line 54:
```suggestion
- Added support for ``GSR`` (galvanic skin response / electrodermal activity, EDA) and ``TEMP`` (temperature) channel types, by `Richard Höchenberger`_ (:gh:`xxx`)
```

sappelhoff (approved):

hoechenberger:
@agramfort feel free to merge / enable automerge if happy

hoechenberger:
This is because I update our testing dataset and we have added weight and height writing since that dataset was originally created. So when I now re-write the exact same dataset, suddenly weight and height columns pop up. These are undesired side effects, yes. But the only way to remove those would be through manual edits

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_eeg.json
```diff
@@ -1,7 +1,7 @@
 {
     "TaskName": "rest",
     "Manufacturer": "Brain Products",
-    "PowerLineFrequency": 60,
+    "PowerLineFrequency": 50.0,
```
hoechenberger, line 4: No idea why this is changing… @sappelhoff could you maybe take a look?

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_channels.tsv
```diff
@@ -66,3 +66,5 @@ PO8	EEG	µV	0.015915494309189534	1000.0	ElectroEncephaloGram	5000.0	good	n/a
 ECG	ECG	µV	0.015915494309189534	1000.0	ElectroCardioGram	5000.0	good	n/a
 HEOG	EOG	µV	0.015915494309189534	1000.0	ElectroOculoGram	5000.0	good	n/a
 VEOG	EOG	µV	0.015915494309189534	1000.0	ElectroOculoGram	5000.0	good	n/a
+GSR	GSR	S	0.015915494309189534	1000.0	Galvanic skin response (electrodermal activity, EDA)	5000.0	good	left index finger
+Temperature	TEMP	oC	0.015915494309189534	1000.0	Other type of channel	5000.0	good	left ear
```
hoechenberger, line 70: The channel description is incorrect here ("Other type of channel")
This needs to be fixed

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_eeg.vmrk
```diff
@@ -1,13 +1,15 @@
...
 ; Each entry: Mk<Marker number>=<Type>,<Description>,<Position in data points>,
-; <Size in data points>, <Channel number (0 = marker is related to all channels)>
+;             <Size in data points>, <Channel number (0 = marker is related to all channels)>
+;             <Date (YYYYMMDDhhmmssuuuuuu)>
 ; Fields are delimited by commas, some fields might be omitted (empty).
 ; Commas in type or description text are coded as "\1".
-Mk1=New Segment,,1,1,0,20000101120000000000
-Mk2=Comment,ControlBox is not connected via USB,1,1,0
+Mk1=Stimulus,S  1,1,1,0
+Mk2=Stimulus,S  2,1001,1,0
```
hoechenberger, line 15: This change seems to indicate that manual the events were never written in the previous version of the script, @sappelhoff 

Meaning we have a bug in MNE-BIDS, where the events passed to write_raw_bids() are not written in case the original data is just being copied over.

Here in this PR, where we enforce writing the data anew (because the raw is modified), the events show up.

@sappelhoff Can you follow what I mean?

hoechenberger:
I'm seeing

> Reading channel info from /Users/hoechenberger/Development/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_channels.tsv.
> /Users/hoechenberger/Development/mne-bids/mne_bids/read.py:586: RuntimeWarning: The unit for channel(s) GSR has changed from NA to S.
>   raw.set_channel_types(channel_type_bids_mne_map_available_channels)

This needs a fix

All good for the temperature channel

hoechenberger:
> This needs a fix
> 
> All good for the temperature channel

I think it's due to a limitation in the MNE-Python BV reader, which by default classifies the GSR and temperature channels as `misc`. Only after reading, MNE-BIDS sets the channel types

But read_raw_brainvision() also offers to provide a list of EOG channels. We currently dont make use of this; and we could expand support to GSR and temperature as well. But not in this PR :)

https://mne.tools/stable/generated/mne.io.read_raw_brainvision.html?highlight=read_raw_brainvision#mne.io.read_raw_brainvision

So I'd say for now we'll just have to live with the warning re channel unit for the GSR channel changes

hoechenberger:
@agramfort I've manually removed changes to the testing dataset that are not immediately related to this PR. This will keep the diff small and clean.

Once merged, we should open a followup PR where we re-run `make_tiny_bids_dataset.py` to get everything up-to-date, I suppose

mne_bids/write.py
```diff
@@ -120,12 +124,37 @@ def _channels_tsv(raw, fname, overwrite=False):
         ch_type.append(map_chs[_channel_type])
         description.append(map_desc[_channel_type])
     low_cutoff, high_cutoff = (raw.info['highpass'], raw.info['lowpass'])
-    if raw._orig_units:
+    if (
+        raw._orig_units is not None and
+        len(raw._orig_units) == len(raw.ch_names)
+    ):
         units = [raw._orig_units.get(ch, 'n/a') for ch in raw.ch_names]
```
agramfort, line 131: I don’t know why this line is not enough to deal with missing units in _orig_units and why you need to do this limbo below

hoechenberger: The handling we have in `main` relies on `raw._orig_units`, yes.

But I loaded an existing raw and appended the GSR and temperature channels via `raw.add_channels()`. This leads to a situation where `raw._orig_units` does **not** contain the two new channels (IDK if this is maybe a bug in MNE?)

So when I try to write this new raw, MNE-BIDS `main` discovers, "oh nice, there's `raw._orig_units`!` and uses those. But since the two newly-added channels are missing there, they'll get the unit `n/a` assigned, which is not correct.

This is not relevant for our tests here because we're loading the testing dataset and not modifying it "on the fly", so I can revert this change for now.

But there's certainly a bug here somewhere, either in MNE-BIDS, which shouldn't use `raw._orig_units`, or in MNE-Python, which should update `_orig_units` on `raw.add_channels()`

hoechenberger: For context, I've reverted the changes @agramfort complained about in [ece3ebf](https://github.com/mne-tools/mne-bids/pull/1059/commits/ece3ebff0221d731dfe49589dc1fa5034631ef23)

But we definitely need to look into this again, I very much believe we have a bug somewhere

sappelhoff: > or in MNE-Python, which should update _orig_units on raw.add_channels()

to me it's this one that should be fixed :point_up:

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_eeg.json
sappelhoff: no idea why it was 60 before ... 50 is correct (this was recorded in Berlin) and expected based on the code.

mne_bids/write.py
hoechenberger: @sappelhoff Could you maybe try to produce an MWE and open an issue with MNE? I'm rather busy right now with this PR and other things on my plate :(

hoechenberger:
I adjusted the `test_search_folder_for_text()` test slightly so it's more robust and it now passes the test suite here.

mne_bids/write.py
sappelhoff: https://github.com/mne-tools/mne-python/issues/11119

hoechenberger: <3 thanks!!

hoechenberger:
@agramfort I hope everything is clearer (and cleaner :)) now!

agramfort (approved):
🙏 @hoechenberger

Issue #1065, opened 2022-09-10
BUG: Handling subjects that don't start with sub- ?


larsoner:
For this dataset:

https://openneuro.org/datasets/ds001971/versions/1.1.1/file-display/participants.tsv

<img width="470" alt="Screen Shot 2022-09-10 at 2 20 43 PM" src="https://user-images.githubusercontent.com/2365790/189496648-be16858a-b137-4af8-82fc-122f36abbef2.png">

I can't load it with MNE-BIDS (via MNE-BIDS-Pipeline):
```
  File "/Users/larsoner/python/mne-bids-pipeline/config.py", line 3453, in _load_data
    raw = read_raw_bids(bids_path=bids_path,

  File "<decorator-gen-612>", line 12, in read_raw_bids

  File "/Users/larsoner/python/mne-bids/mne_bids/read.py", line 793, in read_raw_bids
    raw = _handle_participants_reading(

  File "/Users/larsoner/python/mne-bids/mne_bids/read.py", line 200, in _handle_participants_reading
    row_ind = _safe_index(subjects, subject, allow_all=True)

  File "/Users/larsoner/python/mne-bids/mne_bids/read.py", line 194, in _safe_index
    raise ValueError(f'{exc}{extra}') from None

ValueError: 'sub-001' is not in list. Did you mean one of ['CQ2', 'BT4', 'CQ1', 'CQ3', 'CP9', 'CA7', 'CB6', 'BV2', 'CP8', 'CP1', 'CP2', 'CO9', 'CP5', 'CL1', 'BL7', 'CO3', 'CE2', 'CV6', 'CJ2', 'CY8']?
```

Because it seems like there is a mismatch between the `participant_id` and the filenames:

<img width="612" alt="Screen Shot 2022-09-10 at 2 21 38 PM" src="https://user-images.githubusercontent.com/2365790/189496713-4e832f4d-1765-4a19-8d18-fec4ab08c640.png">

Is this a bug with MNE-BIDS, or with the dataset? If it's with the dataset, it's also a bug with the validator since it's green on OpenNeuro (other than a warning about not all subjects having the same files), right?

hoechenberger:
It's a bug with the dataset, IMHO it doesn't conform to BIDS this way

hoechenberger:
Our error message isn't very helpful though ....

larsoner:
So we should open an issue/comment on OpenNeuro, and also on the BIDS-Validator, then?

[...]

PR #1066, opened 2022-09-10
[MRG] Suggest alternatives when value not found in list

larsoner:
PR Description
--------------

When `lst.index(val)` fails (e.g., `subjects.index(subject)`), try to help people out with suggestions. I was on this PR when I got the traceback in #1065:

> ValueError: 'sub-001' is not in list. Did you mean one of ['CQ2', 'BT4', 'CQ1', 'CQ3', 'CP9', 'CA7', 'CB6', 'BV2', 'CP8', 'CP1', 'CP2', 'CO9', 'CP5', 'CL1', 'BL7', 'CO3', 'CE2', 'CV6', 'CJ2', 'CY8']?

I figured subject lists are generally short enough to `allow_all` to be given (if there are no close matches), but wasn't sure about the others. But they can easily be added elsewhere or tweaked in these cases by follow-up PRs by heavier/more experienced users and devs of MNE-BIDS.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -189,7 +189,7 @@ def _verbose_list_index(lst, val, *, allow_all=False):
     try:
         return lst.index(val)
     except ValueError as exc:
-        extra = get_close_matches(lst, val)
+        extra = get_close_matches(str(val), [str(ll) for ll in lst])
         if allow_all and not extra:
             extra = lst
         extra = f'. Did you mean one of {extra}?' if extra else ''
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -184,11 +184,12 @@ def _read_events(events, event_id, raw, bids_path=None):
 
 
 def _verbose_list_index(lst, val, *, allow_all=False):
-    # try to "return lst.index(val)" but be more informative/verbose when
-    # it fails
+    # try to "return lst.index(val)" for list of str, but be more
+    # informative/verbose when it fails
     try:
         return lst.index(val)
     except ValueError as exc:
+        # Use str cast here to deal with pathlib.Path instances
         extra = get_close_matches(str(val), [str(ll) for ll in lst])
         if allow_all and not extra:
             extra = lst
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -183,7 +183,9 @@ def _read_events(events, event_id, raw, bids_path=None):
     return all_events, all_dur, all_desc
 
 
-def _safe_index(lst, val, *, allow_all=False):
+def _verbose_list_index(lst, val, *, allow_all=False):
+    # try to "return lst.index(val)" but be more informative/verbose when
+    # it fails
     try:
         return lst.index(val)
     except ValueError as exc:
@@ -197,7 +199,7 @@ def _safe_index(lst, val, *, allow_all=False):
 def _handle_participants_reading(participants_fname, raw, subject):
     participants_tsv = _from_tsv(participants_fname)
     subjects = participants_tsv['participant_id']
-    row_ind = _safe_index(subjects, subject, allow_all=True)
+    row_ind = _verbose_list_index(subjects, subject, allow_all=True)
     raw.info['subject_info'] = dict()  # start from scratch
 
     # set data from participants tsv into subject_info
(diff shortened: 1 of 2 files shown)


larsoner:
... on `main` the error message is just:
```
ValueError: 'sub-001' is not in list
```
which I think is much less helpful.

mne_bids/read.py
```diff
@@ -182,10 +183,21 @@ def _read_events(events, event_id, raw, bids_path=None):
     return all_events, all_dur, all_desc
 
 
+def _safe_index(lst, val, *, allow_all=False):
+    try:
```
hoechenberger, line 187: Could you add just a tiny little docstring here? Just so a newcomer would immediately understand how this function makes indexing "safe"?

hoechenberger (commented):
I love this idea. Just one comment

mne_bids/read.py
larsoner: Heh, that's probably a misnomer. It's more of an informative-fail than safe... I'll see if I can make a better name

hoechenberger: That was **excatly** my point ;)

hoechenberger (approved):

larsoner:
I suspect the failures in https://github.com/mne-tools/mne-bids/runs/8286821887?check_suite_focus=true are due to https://github.com/mne-tools/mne-python/pull/11146, where EEG positions in `chs[idx]['loc'][:3]` are now put in `info['dig']` to keep things consistent. I'm not quite sure the right way to deal with this in MNE-BIDS, though. Should we just ignore the warning about not having fiducials in the tests?

agramfort:
doing 

`mne.io.read_raw_cnt(os.path.expanduser("~/mne_data/MNE-testing-data/CNT/scan41_short.cnt")).info`

replicates the problem. You see that there is no fiducials read from the files.

I see two options we make it just a warning and not a ValueError in write_raw_bids
and we filter it in the test or we consider than having dig in head after mne.io.read_raw_cnt while fiducials are not present makes little sense as head is defined from fiducials. I am leaning towards the 2nd solution.

thought?

hoechenberger:
Pinging @sappelhoff

sappelhoff:
in principle I am also leaning towards @agramfort's 2nd solution ... except that one may get coordinates that are already in "head" (which is the same as "captrak", and just in general a sensible /obvious way to define coordinates) without fiducials --> these are still valid, even if no fiducials are present, no?

agramfort:
can you come up with good fiducials automatically for captrack?

sappelhoff:
> can you come up with good fiducials automatically for captrack?

it's the same as head, LPA is at (-1, 0, 0), RPA is at (1, 0, 0), NAS is at (0, 1, 0) ... not sure if it's possible to estimate the positions based on that. probably not

sappelhoff:
@agramfort @larsoner @hoechenberger I think we should agree on a fix for this problem, because our regular CI fails because of it for many days now :-)

Taking @agramfort's [comment](https://github.com/mne-tools/mne-bids/pull/1066#issuecomment-1242976866):

- Is the present problem only an issue for CTF data?
- Taking Alex's point 2 --> how **can** CTF have dig in head when there are no fiducials present?

how do you guys think we should fix it?

larsoner:
> Taking Alex's point 2 --> how can CTF have dig in head when there are no fiducials present?

To be clear we're talking about CNT not CTF. CTF usually has fiducials IIRC because the HPI coils are placed there.

FWIW I'm happy with putting these in UNKNOWN coord frame instead as kind of a bug fix. Previously the behavior was just to put locations in `info['chs'][ii]['loc'][:3]` and not in dig, and `info['chs']` for EEG were (until recently) *always* assumed to be in head coordinates.

I see two options really:

1. A bigger-harder change is to be smarter about it now and do some UNKNOWN coord frame stuff in MNE-Python and here.
2. A smaller and quicker (but less correct) change is just to detect this head-frame-but-no-fiducials case for EEG-only data and strip the `info['dig']` EEG points. This would be equivalent to behavior in MNE-Python 1.1. This would at least allow us to decide to do something smarter later.

My vote is to do (2) for now since MNE-Python 1.2 release is in the next week. Then we can take weeks/months to sort out the correct solution via (1) in follow-up PRs.

PR #1067, opened 2022-09-12
Fix chpi write

### LINKED ISSUE #940: [BUG] cHPI information in meg recording sidecar set incorrectly
#### Describe the bug
Since the last version of mne_bids, the `ContinuousHeadLocalization` field is set to true in my data (recorded with Neuromag Vectorview), even though I have not activated it during the measurement. First I thought I made a mistake and accidentally activated it, but looking at `raw.plot_psd()` there is no above-baseline activity at the specified frequencies. And since this happens to a bunch of files I have recorded (for whom I am certain that I have not activated cHPI every single time), I am rather sure cHPI was indeed inactive.

I had a look at how mne-bids retrieves cHPI information and sets the field. Essentially, it checks (via mne) whether the `raw.info` file includes frequencies at which the HPIs supposedly oscillated. Indeed such frequencies are set in my raw files. However, my hunch is that they are not diagnostic for whether cHPI was active or not, but are set per default, by some general settings or something like that. Therefore, relying on whether frequencies are set or not, might not be a valid criterion here. 

In fact, looking at `mne.chpi.get_chpi_info()`, which is what is used to decide about cHPI, also has a return value that supposedly indicates whether cHPI is on or off:
https://github.com/mne-tools/mne-python/blob/b7e8c7791b000258b2ba158ddd89eecbbb35af30/mne/chpi.py#L310-L318

which might be a better choice to infer whether or not cHPI or not were used. However, in seems that also these information are non-zero in my case. So, I don't know what is going on. Next week I shall try to record some test files with and without cHPI, to see in which way they differ. 

#### Additional information
. I use the latest stable versions of both mne and mne-bids. The relevant part of the code is this: 
https://github.com/mne-tools/mne-bids/blob/0473a3e50deda6db0a0e3fba5feaa62a3d0486ad/mne_bids/write.py#L751

eort:
PR Description
--------------
closes #940 
Given the added functionality in [mne-python](https://github.com/mne-tools/mne-python/pull/11122), I adapted the way the status of the hpi's is determined for neuromag files. Essentially, if any of the hpi's were active for at least one timepoint the field is set. 

I wasn't sure what to do about the `HeadCoilFrequency` field. It seems like it is set regardless of the hpi status, but I think it is confusing (and rather pointless), to have the frequencies listed even though non of them were active. So I set them to None in case of inactive hpi.

Also the functionality would require the latest version of mne-python (1.2)

Merge checklist
---------------

If applicable:

- [x] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -36,6 +36,7 @@ The following authors had contributed before. Thank you for sticking around! �
 * `Robert Luke`_
 * `Stefan Appelhoff`_
 * `Dominik Welke`_
+* `Eduard Ort`_
 
 Detailed list of changes
 ~~~~~~~~~~~~~~~~~~~~~~~~
@@ -74,6 +75,7 @@ Detailed list of changes
 
 - Until now, :class:`mne_bids.BIDSPath` prepends extensions with a period "." automatically. We intend to remove this undocumented side-effect and now emit a ``FutureWarning`` if an ``extension`` that does not start with a ``.`` is provided. Starting with MNE-BIDS 0.12, an exception will be raised in this case, by `Richard Höchenberger`_ (:gh:`1061`)
 
+
 🛠 Requirements
 ^^^^^^^^^^^^^^^
 
@@ -107,6 +109,9 @@ Detailed list of changes
 - Internal helper function to :func:`~mne_bids.read_raw_bids` would reject BrainVision data if ``_scans.tsv`` listed a ``.eeg`` file instead of ``.vhdr``, by `Teon Brooks`_ (:gh:`1034`)
 
 - Whenever :func:`~mne_bids.read_raw_bids` encounters a channel type that currently doesn't translate into an appropriate MNE channel type, the channel type will now be set to ``'misc``. Previously, seemingly arbitrary channel types would be applied, e.g. ``'eeg'`` for GSR and temperature channels, by `Richard Höchenberger`_ (:gh:`1052`)
+
+- Fix the incorrect setting of the fields `ContinuousHeadLocalization` and `HeadCoilFrequency` for Neuromag meg recordings, by `Eduard Ort`_ (:gh:`1067`)
+
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
 
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -731,7 +731,6 @@ def test_fif(_bids_validate, tmp_path):
 def test_chpi(_bids_validate, tmp_path, format):
     """Test writing of cHPI information."""
     data_path = testing.data_path()
-    kit_data_path = op.join(base_path, 'kit', 'tests', 'data')
 
     if format == 'fif_no_chpi':
         fif_raw_fname = op.join(data_path, 'MEG', 'sample',
@@ -744,6 +743,7 @@ def test_chpi(_bids_validate, tmp_path, format):
         ctf_raw_fname = op.join(data_path, 'CTF', 'testdata_ctf.ds')
         raw = _read_raw_ctf(ctf_raw_fname)
     elif format == 'kit':
+        kit_data_path = op.join(base_path, 'kit', 'tests', 'data')
         kit_raw_fname = op.join(kit_data_path, 'test.sqd')
         kit_hpi_fname = op.join(kit_data_path, 'test_mrk.sqd')
         kit_electrode_fname = op.join(kit_data_path, 'test.elp')
@@ -763,14 +763,24 @@ def test_chpi(_bids_validate, tmp_path, format):
     if parse_version(mne.__version__) <= parse_version('0.23'):
         assert 'ContinuousHeadLocalization' not in meg_json_data
         assert 'HeadCoilFrequency' not in meg_json_data
-    elif format in ['fif_no_chpi', 'kit']:
+
+    elif format in ['fif_no_chpi', 'fif']:
+        if parse_version(mne.__version__) <= parse_version('1.1'):
+            assert meg_json_data['ContinuousHeadLocalization'] is None
+            assert meg_json_data['HeadCoilFrequency'] is None
+        else:
+            if format == 'fif_no_chpi':
+                assert meg_json_data['ContinuousHeadLocalization'] is False
+                assert meg_json_data['HeadCoilFrequency'] == []
+            elif format == 'fif':
+                assert meg_json_data['ContinuousHeadLocalization'] is True
+                assert_array_almost_equal(meg_json_data['HeadCoilFrequency'],
+                                          [83., 143., 203., 263., 323.])
+
+    elif format == 'kit':
         # no cHPI info is contained in the sample data
         assert meg_json_data['ContinuousHeadLocalization'] is False
         assert meg_json_data['HeadCoilFrequency'] == []
-    elif format == 'fif':
-        assert meg_json_data['ContinuousHeadLocalization'] is True
-        assert_array_almost_equal(meg_json_data['HeadCoilFrequency'],
-                                  [83., 143., 203., 263., 323.])
     elif format == 'ctf':
         assert meg_json_data['ContinuousHeadLocalization'] is True
         assert_array_equal(meg_json_data['HeadCoilFrequency'],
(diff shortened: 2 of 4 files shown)


eort:
> Also the functionality would require the latest version of mne-python (1.0.2)

which is at least partially responsible for the tests failing

agramfort:
@eort can you make your PR draft if it’s not ready yet ? 🙏

sappelhoff:
> I wasn't sure what to do about the HeadCoilFrequency field. It seems like it is set regardless of the hpi status, but I think it is confusing (and rather pointless), to have the frequencies listed even though non of them were active. So I set them to None in case of inactive hpi.

agreed

> Also the functionality would require the latest version of mne-python (1.2)

can you please add some logic such that this new feature is only used when an adequate version of mne-python is present, and else the old code is used? You can also add a comment like `XXX: Remove this when support for mne <1.2 is dropped`

Same for a test: just use the "skipif" decorator that you can already find in several other tests.

eort:
> can you please add some logic such that this new feature is only used when an adequate version of mne-python is present, and else the old code is used? 

Yes, I can do that. I don't think leaving the old code makes a lot of sense though, because the old code gives incorrect information (always claiming chpi was present). But like Eric pointed out, the `get_chpi_info` method doesn't provide any info about whether the hpi were active or not. So for versions < 1.2 I would rather just not set those fields. This would then also have consequences for the tests, as neuromag files wouldn't be tested if the mne version is <1.2. I added similar version checks to the tests. 

> Same for a test: just use the "skipif" decorator that you can already find in several other tests.

I couldn't find any 'live' usage of this decorators. Regardless, can they be used if those version checks are only relevant for neuromag files? The behavior for CTF and KIT files hasn't change.

sappelhoff:
The changes you pushed LGTM, thanks! If you could also add a changelog entry, I think this becomes mergeable. We just need to inspect the CI log very well to make sure no "new issues" are introduced, and errors are all due to expected shortcomings (see discussion in #1066)

eort:
Talking about introducing new issues... What is preferred behavior in case it cannot be determined whether a field is True or False? Is it setting it to `None` or leaving it out altogether? So far, it was left out, but my understanding was setting it to `None`

sappelhoff:
in the case of missing data, if it is metadata that BIDS considers **optional**, leaving it out completely is preferred IMHO.

If it is **recommended** and we don't have info, we need to set it to `n/a`, I think.

eort:
Both `HeadCoilFrequency` and `ContinuousHeadLocalization` are recommended, so I set them to `n/a`

eort:
Turns out things are more complicated than I thought. 

1. 'n/a' is not a valid value for `HeadCoilFrequency` and `ContinuousHeadLocalization` (which is probably the reason they were left out in first place). So, we either change the bids specifications such that `"n/a"` will be allowed, or we only use those allowed values (bool for `ContinuousHeadLocalization`, and numbers/lists of numbers for `HeadCoilFrequency`. I lean towards the first option. Opinions?

2. the new method [`get_active_chpi`](https://github.com/mne-tools/mne-python/blob/be6135a7f441581ae71cbe0403ee0b36774b8f69/mne/chpi.py#L1389) calls [`get_chpi_info`]https://github.com/mne-tools/mne-python/blob/be6135a7f441581ae71cbe0403ee0b36774b8f69/mne/chpi.py#L302
in case of missing hpi data this will throw an error. While latter method has `on-missing` argument, so we could add that argument to `get_active_chpi` and pass it through to `get_chpi_info`. Or are there better options?

sappelhoff:
Ah, I see. I think it should be possible to specify `n/a` so I also lean towards your first option. Else we can continue with omitting these metadata.

eort:
mh, after checking again the [specifications](https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#compulsory-optional-and-additional-data-and-metadata), I came across this passage:

> It is RECOMMENDED that non-compulsory metadata fields (like notch in channels.tsv files) and/or files (like events.tsv) are fully omitted when they are unavailable or unapplicable, instead of specified with an n/a value, or included as an empty file (for example an empty events.tsv file with only the headers included).

Given that we are talking about optional fields here, I think adding `n/a` is not ideal, instead the field should just not be set when no information on the status are available.

sappelhoff:
:face_exhaling: shame on me, because I was the one adding that sentence to the spec. And I didn't remember when I wrote what I did above. Thanks for double checking @eort

eort:
no prob. Seems unrealistic to remember all details of a 500 pages doc ;)

So only thing that needs to be done still, is the api change in `mne.chpi.get_active_hpi()`. Once that is merged, I will push the latest version to this PR.

doc/whats_new.rst
```diff
@@ -108,6 +110,9 @@ Detailed list of changes
 
 - Whenever :func:`~mne_bids.read_raw_bids` encounters a channel type that currently doesn't translate into an appropriate MNE channel type, the channel type will now be set to ``'misc``. Previously, seemingly arbitrary channel types would be applied, e.g. ``'eeg'`` for GSR and temperature channels, by `Richard Höchenberger`_ (:gh:`1052`)
 
+- Fix the incorrect setting of the fields `ContinuousHeadLocalization` and `HeadCoilFrequency` for Neuromag meg recordings, by `Eduard Ort`_ (:gh:`1067`)
```
hoechenberger, line 113:
```suggestion
- Fix the incorrect setting of the fields ``ContinuousHeadLocalization`` and ``HeadCoilFrequency`` for Neuromag MEG recordings, by `Eduard Ort`_ (:gh:`1067`)
```

mne_bids/tests/test_write.py
```diff
@@ -763,18 +763,27 @@ def test_chpi(_bids_validate, tmp_path, format):
...
+            assert 'HeadCoilFrequency' not in meg_json_data
+        else:
+            if format == 'fif_no_chpi':
+                assert meg_json_data['ContinuousHeadLocalization'] is False
+                assert meg_json_data['HeadCoilFrequency'] == []
+            elif format == 'fif':
+                assert meg_json_data['ContinuousHeadLocalization'] is True
+                assert_array_almost_equal(meg_json_data['HeadCoilFrequency'],
+                                          [83., 143., 203., 263., 323.])
+
```
hoechenberger, line 779:
```suggestion
```

mne_bids/tests/test_write.py
```diff
@@ -763,18 +763,27 @@ def test_chpi(_bids_validate, tmp_path, format):
     if parse_version(mne.__version__) <= parse_version('0.23'):
         assert 'ContinuousHeadLocalization' not in meg_json_data
         assert 'HeadCoilFrequency' not in meg_json_data
-    elif format in ['fif_no_chpi', 'kit']:
+
```
hoechenberger, line 766:
```suggestion
```

eort:
there are still problems in mne.chpi.get_active_chpi(). Once they are solved, I shall try once again to get this thing to work

sappelhoff (dismissed):
LGTM, thanks @eort for working on this and the added functionality in mne-python!

sappelhoff:
oh but actually, we **do** get one error that seems related, see: https://github.com/mne-tools/mne-bids/actions/runs/3080131213/jobs/4977127367#step:14:5540

> E           NotImplementedError: Identifying active HPI channels is not implemented for other systems than neuromag.

eort:
oh indeed. I missed that among all the matplotlib errors. 

I am confused about the message though. It says that it fails to properly handle a `RawCNT` file, but in the configuration of the test, only `('fif_no_chpi', 'fif', 'ctf', 'kit')` are listed. Where does the file come from?

In any case, the problem seems to be that all system that are not KIT or CTF are lumped into one category, assuming those must be neuromag then. So, we could change it to an elif, and throw and notimplementederror for the else case?

sappelhoff:
> In any case, the problem seems to be that all system that are not KIT or CTF are lumped into one category, assuming those must be neuromag then. So, we could change it to an elif, and throw and notimplementederror for the else case?

yes, seems like it -- I don't know how (would need to read more carefully) but it seems like you need to fix exactly that logic.

eort:
I used the same logic as in https://github.com/mne-tools/mne-python/blob/de546e2188761de69eede103b7c5a6cc09542a4d/mne/chpi.py#L1392 and restructured the entire if-else section accordingly.

eort:
I don't see anything wrong with the report problems, but better check for yourself

sappelhoff:
Thank you @eort :tada:

PR #1073, opened 2022-09-29
MAINT: Ensure we dont download without permission

larsoner:
Lots of failures locally with this.

Okay for me to put `download=False` in a bunch of `data_path()` calls and add `requires_testing_data` calls where necessary?

- [x] Show failures: https://github.com/mne-tools/mne-bids/actions/runs/3152298983/jobs/5127340379
- [x] Fix failures
- [x] Revert change to use `larsoner` branch in favor of MNE-Python `main` once https://github.com/mne-tools/mne-python/pull/11213 is merged

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -213,7 +213,7 @@ jobs:
     # Only run on a limited set of jobs
     - name: Run pytest without testing data
       run: make test
-      if: ${{ matrix.mne-version == 'mne-main' }}
+      if: ${{ matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest' && matrix.mne-version == 'mne-main' }}
 
     # Get testing data
     - run: ./tools/get_testing_version.sh
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -213,7 +213,7 @@ jobs:
     # Only run on a limited set of jobs
     - name: Run pytest without testing data
       run: make test
-      if: matrix.python_version == '3.10' && matrix.os == 'ubuntu-latest' && matrix.mne-version == 'mne-main'
+      if: ${{ matrix.python_version == '3.10' && matrix.os == 'ubuntu-latest' && matrix.mne-version == 'mne-main' }}
 
     # Get testing data
     - run: ./tools/get_testing_version.sh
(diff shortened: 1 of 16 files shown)


hoechenberger:
Good enough for me!

larsoner:
A few things I found and fixed along the way:

1. Don't create `data = np.empty(...)` for `RawArray` in tests. The array can contain junk and lead to casting errors (e.g., to `float32` when wrining).
2. Don't use bare `np.random.random(...)` for the same reason. It can contain numbers that fit in `float64` but not `float32` leading to overflow errors when casting.
3. Added a `bad_frame_xfail` that we can remove once https://github.com/mne-tools/mne-python/pull/11212 is merged and any necesary changes are made here, in the meantime it allows tests here to be green
4. Added to CIs a run that executes tests *without* the testing data. This should be quick (locally takes 6 sec) and pass, and ensure that `requires_testing_data` decorators are used properly.

Pushing a commit now that hopefully will pass!

larsoner:
... I also added dataset caching, and removed one use of `sample` data where we could just as easily use `testing`.

.github/workflows/unit_tests.yml
```diff
@@ -158,13 +158,13 @@ jobs:
...
       run: |
-        git clone --depth 1 https://github.com/mne-tools/mne-python.git -b maint/1.0
+        git clone --depth 1 --single-branch --branch maint/1.0 https://github.com/mne-tools/mne-python.git -b maint/1.0
         python -m pip install -e ./mne-python
 
     - name: Install MNE (main)
       if: "matrix.mne-version == 'mne-main'"
       run: |
-        git clone --depth 1 https://github.com/mne-tools/mne-python.git -b main
+        git clone --depth 1 --single-branch --branch main https://github.com/mne-tools/mne-python.git -b main
```
agramfort, line 167: shall we keep this?

larsoner: Yes it's just more specific/explicit and a tiny bit faster

larsoner:
All green, ready for review/merge from my end

.github/workflows/unit_tests.yml
sappelhoff: but we should probably remove the trailing `-b main`, right? that's just a single flag specified twice?

sappelhoff:
Thanks!

PR #1076, opened 2022-10-07
MAINT: Tolerate new MNE behavior

larsoner:
@hoechenberger should fix the failing test following https://github.com/mne-tools/mne-python/pull/11212 where fiducials are estimated and added when implicitly converting to the head coordinate frame

### DIFF
--- a/mne_bids/dig.py
+++ b/mne_bids/dig.py
@@ -10,7 +10,6 @@ from collections import OrderedDict
 from pathlib import Path
 import re
 import warnings
-from importlib_resources import files
 
 import mne
 import numpy as np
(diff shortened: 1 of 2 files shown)


larsoner:
I also just pushed a commit to avoid using `importlib_resources`. It's causing a conda-forge build error:

https://github.com/conda-forge/mne-bids-feedstock/pull/21

Keeping it would mean 1) adding to setup.py deps, probably dependent on Python version, 2) changing the import, probably with a try/except, 3) updating the feedstock. What I've implemented here instead is just `data_dir = Path(__file__).parent / 'data'` which requires none of these gymnastics, and hopefully is good enough.

hoechenberger:
Sounds good to me. Some tests are still failing though!

larsoner:
Those should be fixed by https://github.com/mne-tools/mne-python/pull/11230

larsoner:
Ready for review/merge from my end

sappelhoff (approved):

sappelhoff:
Thanks @larsoner

sappelhoff:
we should probably have removed this line as well with this PR, right?

https://github.com/mne-tools/mne-bids/blob/ee27b43acd7fede44a384fe5762f90c70a460789/setup.cfg#L41

larsoner:
Yes we should have, thanks @sappelhoff !

PR #1079, opened 2022-10-09
avoid creating many Annotations objects in for loop

agramfort:
without this fix read_raw_bids on a subject from https://openneuro.org/datasets/ds003825 takes 600s on my laptop. With this fix it's < 3s.


hoechenberger:
@agramfort Care to add a short changelog entry?
Otherwise LGTM

sappelhoff (approved):
nice

hoechenberger:
Thanks @agramfort!

PR #1086, opened 2022-10-17
Write all annotations even if event_id was passed

### LINKED ISSUE #1084: We're not writing all annotations if events are passed to write_raw_bids() as well

This problem came up in https://mne.discourse.group/t/mne-bids-structure-one-subject-is-different/5687/

The user concatenated raws, which creates an `edge` annotation. However, after writing the data via `write_raw_bids()`, the annotation didn't show up in `events.tsv`. Turns out the user passed `events` and `event_id` to `write_raw_bids()` too, which caused this issue to surface.

### Steps to reproduce

In the following code, the concatenated raw has an edge and a bad annotation, but they don't end up in the `events.tsv` sidecar:

```Python
# %%
import pandas as pd
import mne
import mne_bids

sample_dir = mne.datasets.sample.data_path()
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname)
raw_concat = mne.concatenate_raws([raw.copy(), raw])
events = mne.find_events(raw_concat)
event_id = {
    'Auditory/Left': 1,
    'Auditory/Right': 2,
    'Visual/Left': 3,
    'Visual/Right': 4,
    'Smiley': 5,
    'Button': 32
}

# %%
bp = mne_bids.BIDSPath(
    subject='sample',
    task='sample',
    suffix='meg',
    extension='.fif',
    datatype='meg',
    root='/tmp/bids-root'
)
mne_bids.write_raw_bids(
    raw_concat,
    bids_path=bp,
    events=events,
    event_id=event_id,
    overwrite=True,
    allow_preload=True,
    format='FIF'
)

# %%
events_tsv_path = (
    bp
    .copy()
    .update(suffix='events', extension='.tsv')
)
events_tsv = pd.read_csv(events_tsv_path, sep='\t')
print(events_tsv['trial_type'].unique())
```

### Expected results

We should see 642 events in `events.tsv`, including the edge and bad annotations.

### Actual results

The edge and bad annotations are missing from the `events.tsv` sidecar.

### Additional information

The problem occurs here:
https://github.com/mne-tools/mne-bids/blob/9231836b227dceecec5f73b498a32591ef2f31ef/mne_bids/read.py#L159-L163

While in line 162 we explicitly try to include edge and bad annotations, the documentation of `events_from_annotations()` also states that any annotations that do not appear in `event_id` will simply be dropped.

IMHO the cleanest solution here would be to force users to always supply an event_id dictionary. Currently it's optional if no `events` are passed, which causes us to generate arbitrary event codes for annotations – which at least one user complained about not too long ago. Of course this would break lots of existing code.

We could of course also add special handling for all annotations for which no event_id mapping has been supplied, and we simply generate some new (arbitrary) event codes.

@sappelhoff @agramfort WDYT?

hoechenberger:
At this stage, I've only added a test that fails and demonstrates the bug.

Fixes #1084

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -2704,6 +2704,50 @@ def test_annotations(_bids_validate, bad_segments, tmp_path):
     _bids_validate(bids_root)
 
 
+@pytest.mark.filterwarnings(warning_str['channel_unit_changed'])
+@testing.requires_testing_data
+def test_annotations_and_events(_bids_validate, tmp_path):
+    """Test combined writing of Annotations and events."""
+    bids_root = tmp_path / 'bids'
+    bids_path = _bids_path.copy().update(root=bids_root, datatype='meg')
+    raw_fname = data_path / 'MEG' / 'sample' / 'sample_audvis_trunc_raw.fif'
+    events_fname = (
+        data_path / 'MEG' / 'sample' / 'sample_audvis_trunc_raw-eve.fif'
+    )
+
+    events = mne.read_events(events_fname)
+    events = events[events[:, 2] != 0]  # drop unknown "0" events
+    event_id = {'Auditory/Left': 1, 'Auditory/Right': 2, 'Visual/Left': 3,
+                'Visual/Right': 4, 'Smiley': 5, 'Button': 32}
+    raw = _read_raw_fif(raw_fname)
+    annotations = mne.Annotations(
+        # Try to avoid rounding errors.
+        onset=(
+            1 / raw.info['sfreq'] * 600,
+            1 / raw.info['sfreq'] * 600,  # intentional
+            1 / raw.info['sfreq'] * 3000
+        ),
+        duration=(
+            1 / raw.info['sfreq'],
+            1 / raw.info['sfreq'],
+            1 / raw.info['sfreq'] * 200
+        ),
+        description=('BAD_segment', 'EDGE_segment', 'custom'),
+    )
+    raw.set_annotations(annotations)
+
+    # Jointly write annotations and events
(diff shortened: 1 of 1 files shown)


hoechenberger:
New logic:

- If only ``events`` and ``event_id`` are passed, all remains unchanged
- If no ``events`` are passed, but the data has annotations, it's now possible (optional!) to specify ``event_id`` to allow for a custom mapping from annotation descriptions to event codes. If ``event_id`` is passed, though, it **must** contain **all** annoation descriptions. If it is not passed, we will assign arbitrary event codes to each annotation description. This is the old behavior, i.e., this bit remains unchanged; however, we now emit an info log message
- If ``events`` are passed and the data has annotations, we now require that ``event_id`` not only contain a mapping for all events, but also for all annotation descriptions, otherwise we raise an exception

hoechenberger:
If we want to go a different route and allow for incomplete `event_id` mappings for annotations, we'll have to work upstream on MNE's `events_from_annotations()`

hoechenberger:
I know this is complex stuff, but @sappelhoff if you have 20 min to review, it would be great :) because it's actually an annoying bug that ideally should get fixed soon

sappelhoff:
I'll try to do it soon!

sappelhoff (approved):
LGTM, I like the new behavior! Thanks @hoechenberger!

sappelhoff:
I added a versioning section and patch release section to the release wiki: https://github.com/mne-tools/mne-bids/wiki/How-to-make-a-release

and I'll look into filling up some more information there and making a patch release soon @hoechenberger

PR #1093, opened 2022-11-01
BUG: Fix API for candidates

larsoner:
PR Description
--------------

So in #1083 I added `return_candidates=True`, but this doesn't really solve the use case. The "list all candidates" step needs to be *separate* from the "traverse and read all candidates", and it's not currently. This PR splits them. Now instead of doing:
```
fname, candidates = bids_path.find_empty_room(return_candidates=True)
```
you do
```
candidates = bids_path.get_empty_room_candidates()
fname = bids_path.find_empty_room()
```
And the `get_empty_room_candidates` does not read any files, just lists/globs filenames so should be very fast (and negligible I/O overhead).

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_path.py
+++ b/mne_bids/tests/test_path.py
@@ -19,7 +19,7 @@ from mne_bids import (get_datatypes, get_entity_vals, print_dir_tree,
                       BIDSPath, write_raw_bids, read_raw_bids,
                       write_meg_calibration, write_meg_crosstalk)
 from mne_bids.path import (_parse_ext, get_entities_from_fname,
-                           _find_best_candidates, _find_matching_sidecar,
+                           _find_best_candidates,
                            _filter_fnames, search_folder_for_text,
                            get_bids_path_from_fname)
 from mne_bids.config import ALLOWED_PATH_ENTITIES_SHORT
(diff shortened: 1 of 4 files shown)


larsoner:
... can I make `_find_matching_sidecar` public as `bids_path.find_matching_sidecar()`? Then we could actually get the sidecar paths for stuff like this, which is nice.

larsoner:
Okay I pushed a `bids_path.find_matching_sidecar` as this is something else we've wanted for mne-bids-pipeline. If these sidecar files are updated, we need to rerun steps, so it's good to be able to find the same files that `mne-bids` will use / look at...

mne_bids/tests/test_path.py
```diff
@@ -19,7 +19,7 @@
                       BIDSPath, write_raw_bids, read_raw_bids,
                       write_meg_calibration, write_meg_crosstalk)
 from mne_bids.path import (_parse_ext, get_entities_from_fname,
-                           _find_best_candidates, _find_matching_sidecar,
+                           _find_best_candidates,
```
larsoner, line 22: I changed the private `_find_matching_sidecar` calls in this file to use the public interface. There are other places the private function is used like `test_write.py` but it doesn't seem necessary to change them all here (code churn / blame obfuscation with little benefit)

sappelhoff (approved):
seems sensible from a top level view, and code changes look clean as well.

agramfort (approved):

agramfort:
thx @larsoner

PR #1095, opened 2022-11-07
more specific import to maybe avoid issue

### LINKED ISSUE #1094: doc build fails because of mne (main)
> ImportError: cannot import name 'psd_welch' from 'mne.time_frequency' (/home/circleci/mne_bids_env/lib/python3.9/site-packages/mne/time_frequency/__init__.py)

https://app.circleci.com/pipelines/github/mne-tools/mne-bids/4820/workflows/813d1bc7-3b45-463b-af0e-3d5ddab39dc7/jobs/6961

sappelhoff:
maybe fixes #1094 

... but the real fix must be applied in mne-nirs: replacing `mne.time_frequency.psd_welch` with `raw.compute_psd` or similar.

cc @rob-luke

### DIFF
--- a/examples/convert_nirs_to_bids.py
+++ b/examples/convert_nirs_to_bids.py
@@ -29,7 +29,7 @@ import pathlib
 import shutil
 
 import mne
-import mne_nirs  # For convenient downloading of example data
+from mne_nirs import datasets  # For convenient downloading of example data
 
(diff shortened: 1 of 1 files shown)


rob-luke:
Thanks @sappelhoff
I created an issue in MNE-NIRS so we don't forget to address this.

PR #1099, opened 2022-11-15
ENH: Auto warning stacklevel

larsoner:
Make use of the `ignore_namespaces` to make messages like this with an unhelpful `stacklevel`:
```
/home/larsoner/python/mne-bids/mne_bids/path.py:1583: RuntimeWarning: Did not find any events.tsv associated with sub-0001_task-AEF_run-01.

The search_str was "/home/larsoner/mne_data/ds000246/sub-0001/**/meg/sub-0001*events.tsv"
  warn(msg)
/home/larsoner/python/mne-bids/mne_bids/read.py:580: RuntimeWarning: No BIDS -> MNE mapping found for channel type "SYSCLOCK". Type of channel "SCLK01-177" will be set to "misc".
  warn(
/home/larsoner/python/mne-bids/mne_bids/read.py:580: RuntimeWarning: No BIDS -> MNE mapping found for channel type "OTHER". Type of channel "HADC001-4408" will be set to "misc".
```
automagically resolve to a more helpful stacklevel:
```
/home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/_import_data.py:338: RuntimeWarning: Did not find any events.tsv associated with sub-0001_task-AEF_run-01.

The search_str was "/home/larsoner/mne_data/ds000246/sub-0001/**/meg/sub-0001*events.tsv"
  raw = read_raw_bids(bids_path=bids_path,
/home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/_import_data.py:338: RuntimeWarning: No BIDS -> MNE mapping found for channel type "SYSCLOCK". Type of channel "SCLK01-177" will be set to "misc".
  raw = read_raw_bids(bids_path=bids_path,
/home/larsoner/python/mne-bids-pipeline/mne_bids_pipeline/_import_data.py:338: RuntimeWarning: No BIDS -> MNE mapping found for channel type "OTHER". Type of channel "HADC001-4408" will be set to "misc".
```

As an aside, any reason all of mne-bids's imports are absolute, e.g. `from mne_bids.utils import warn` rather than relative like `from .utils import warn`? I think the latter is more standard, worth switching them in a follow-up PR?

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -25,10 +25,10 @@ from scipy.io import loadmat, savemat
 import mne
 from mne.io import (read_raw_brainvision, read_raw_edf, read_raw_bdf,
                     anonymize_info)
-from mne.utils import logger, verbose, warn
+from mne.utils import logger, verbose
 
 from mne_bids.path import BIDSPath, _parse_ext, _mkdir_p
-from mne_bids.utils import _get_mrk_meas_date, _check_anonymize
+from mne_bids.utils import _get_mrk_meas_date, _check_anonymize, warn
 import numpy as np
 
 
--- a/mne_bids/dig.py
+++ b/mne_bids/dig.py
@@ -14,7 +14,7 @@ import warnings
 import mne
 import numpy as np
 from mne.io.constants import FIFF
-from mne.utils import (logger, warn, _validate_type, _check_option,
+from mne.utils import (logger, _validate_type, _check_option,
                        has_nibabel, get_subjects_dir)
 from mne.io.pick import _picks_to_idx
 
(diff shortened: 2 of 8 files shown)


hoechenberger:
> As an aside, any reason all of mne-bids's imports are absolute

I don't know but it's something I've been wanting to change for a long time. I think the absolute imports can even cause trouble in certain edge cases, so I'd like to see them go

larsoner:
All green!

sappelhoff:
@larsoner @hoechenberger let's keep the absolute imports, please. As far as I understood, it's also recommended this way by PEP8 --> https://pep8.org/#imports

Apart from that I personally find them way less confusing and error prone. Nobody can interpret a number of dots before an import, but everybody can understand a chain of semantically meaningful module/package names.

larsoner:
> As far as I understood, it's also recommended this way by PEP8 --> https://pep8.org/#imports

FWIW it seems less cut-and-dry as they equivocate a bit:

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

That being said, pandas, NumPy, and SciPy all appear to use absolute imports, scikit-learn uses relative. So I think I know where our tendency to use relative comes from :) But both do seem valid/used, so :+1: to leaving it as is here

Issue #1100, opened 2022-11-14
MNE-Python channel unit warnings should be suppressed


jo780-full:
The original issue was resolved, but a second issue emerged on this thread. Skip to [this comment](https://github.com/mne-tools/mne-bids/issues/1100#issuecomment-1315893527) and read on from there. I've hidden the original problem in "details".  (signed, @drammock)

<details>

### Description of the problem

Hello, I am new to mne and have a problem with this dataset.

It shows 
ValueError: invalid literal for int() with base 10: ''

### Steps to reproduce

```Python
`datatype='eeg'
subject='024'
task='med1breath'
suffix='eeg'
bids_path=BIDSPath(subject=subject,task=task,suffix=suffix,datatype=datatype,root=bids_root)
sidecar_path=bids_path.copy().update(extension='.json')
raw = read_raw_bids(bids_path=bids_path)
`
```

### Link to data

[doi:10.18112/openneuro.ds003969.v1.0.0](https://doi.org/10.18112/openneuro.ds003969.v1.0.0)]

### Expected results

Reads file

### Actual results

![image](https://user-images.githubusercontent.com/84071740/201730018-600f2071-673b-480b-a8d7-f2b8c2d3be46.png)

### Additional information

Platform:         Windows-10-10.0.22000-SP0
Python:           3.9.7 (default, Sep 16 2021, 16:59:28) [MSC v.1916 64 bit (AMD64)]
Executable:       C:\Users\josep\anaconda3\python.exe
CPU:              AMD64 Family 23 Model 96 Stepping 1, AuthenticAMD: 16 cores
Memory:           15.4 GB

mne:              1.2.2
numpy:            1.20.3 {MKL 2021.4-Product with 8 threads}
scipy:            1.7.1
matplotlib:       3.4.3 {backend=module://matplotlib_inline.backend_inline}

sklearn:          0.24.2
numba:            0.54.1
nibabel:          Not found
nilearn:          Not found
dipy:             Not found
openmeeg:         Not found
cupy:             Not found
pandas:           1.3.4
pyvista:          Not found
pyvistaqt:        Not found
ipyvtklink:       Not found
vtk:              Not found
qtpy:             1.10.0 {PyQt5=5.9.7}
ipympl:           Not found
pyqtgraph:        Not found
pooch:            v1.6.0

mne_bids:         0.11.1
mne_nirs:         Not found
mne_features:     Not found
mne_qt_browser:   Not found
mne_connectivity: Not found
[...]

PR #1106, opened 2022-12-01
Fix CI via catching a warning

sappelhoff:
PR Description
--------------
fixes CI, see: https://github.com/mne-tools/mne-bids/actions/runs/3589485847/jobs/6041964853

I think this is because of https://github.com/mne-tools/mne-python/pull/11344

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/commands/tests/test_cli.py
+++ b/mne_bids/commands/tests/test_cli.py
@@ -41,7 +41,8 @@ task = 'testing'
 datatype = 'meg'
 
 event_id = {'Auditory/Left': 1, 'Auditory/Right': 2, 'Visual/Left': 3,
-            'Visual/Right': 4, 'Smiley': 5, 'Button': 32}
+            'Visual/Right': 4, 'Smiley': 5, 'Button': 32, 'Nothing': 0}
+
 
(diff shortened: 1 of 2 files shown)


larsoner:
Yes that's the PR that caused it, but I think MNE-BIDS should maybe *internally* squash these warnings using a `record_warnings` or so. It's weird to get a warning when using `read_raw_bids` that units have changed when `read_raw_bids` is doing the right thing (right?) by changing the channel types according to the `channels.tsv` or whatever.

For example, I don't like that we have to use `verbose='error'` here:

https://github.com/mne-tools/mne-python/pull/11348

to fix

https://app.circleci.com/pipelines/github/mne-tools/mne-python/17161/workflows/bb0bdc62-4fa4-4b9c-a781-dc73ab855516/jobs/50797

It seems like mne-bids should just "do the right thing" with channel types and units, and the warning makes it seem to the end user like it potentially is not

hoechenberger:
@larsoner alao see https://github.com/mne-tools/mne-bids/issues/1100, I believe that's exactly what you mean

PR #1108, opened 2022-12-05
MAINT: Update CIs and bump minimum to 3.8

### LINKED ISSUE #1107: weird CI failure in schedules run: scipy build broken, "setuptools" missing
This is the run in question: https://github.com/mne-tools/mne-bids/actions/runs/3616926118

We got:

> ImportError: The `scipy` install you are using seems to be broken, (extension modules cannot be imported), please try reinstalling.

And:

> ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.

:shrug: I have no idea where these are coming from, but I have observed them more and more often in occasional runs. Anyone got an idea?

larsoner:
Closes #1107


.github/workflows/unit_tests.yml
```diff
@@ -144,12 +145,12 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ${{ env.pythonLocation }}
-        key: test-0-${{ env.pythonLocation }}-${{ hashFiles('setup.cfg') }}-${{ hashFiles('test_requirements.txt') }}
+        key: test-0-${{ env.pythonLocation }}-${{ env.os }}-${{ hashFiles('setup.cfg') }}-${{ hashFiles('test_requirements.txt') }}
```
larsoner, line 148: I think this was one real problem -- sharing the cache implicitly across Ubuntu versions (20.04 and 22.04)

larsoner:
@sappelhoff ready for CI branch protection rules to be updated and merged

.github/workflows/unit_tests.yml
```diff
@@ -59,12 +59,13 @@ jobs:
...
       with:
         path: ${{ env.pythonLocation }}
-        key: build-1-${{ env.pythonLocation }}-${{ hashFiles('setup.cfg') }}-${{ hashFiles('test_requirements.txt') }}
+        key: build-2-${{ env.pythonLocation }}-${{ hashFiles('setup.cfg') }}-${{ hashFiles('test_requirements.txt') }}
 
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         python -m pip install wheel setuptools build
+        python -m pip install --upgrade scipy
```
sappelhoff, line 68:
```suggestion
```

hoechenberger:
Do we still support MNE 1.0? Latest stable is 1.2, I don't think we're even testing with 1.0 and 1.2 anymore, right?

sappelhoff:
+1 to upgrade to MNE 1.2 ... I tried to remove the scipy upgrade in https://github.com/mne-tools/mne-bids/pull/1108/commits/72bef097e83f347c7581599f45efe5d6f4a7e752 because I felt like it should pass regardless ... apparently I was wrong.

hoechenberger:
Oh I didn't realize we actually still test with MNE 1.0 -- but neither with 1.1 nor 1.2 (which were only implicitly covered for a while when testing with main)

hoechenberger:
Thanks @larsoner, seems I missed quite a few there!!!

hoechenberger (approved):

PR #1119, opened 2023-02-18
update for new version of google analytics

agramfort:
update for new version of google analytics


hoechenberger:
Thanks @agramfort!

PR #1128, opened 2023-03-22
MAINT: Work around MNE change

larsoner:
Don't use `mne.utils.has_nibabel` since it's gone as of https://github.com/mne-tools/mne-python/pull/11578. Instead, use the `_import_nibabel` pattern which gives a nicer traceback (I think) more compactly:
```
$ python -c "import mne_bids.utils; mne_bids.utils._import_nibabel()"
  File "<string>", line 1, in <module>
  File "/Users/larsoner/python/mne-bids/mne_bids/utils.py", line 465, in _import_nibabel
    raise exc.__class__(
ModuleNotFoundError: nibabel is required to work with MRI data but could not be imported, got: No module named 'nibabel'
```

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -29,6 +29,7 @@ The following authors contributed for the first time. Thank you so much! 🤩
 The following authors had contributed before. Thank you for sticking around! 🤘
 
 * `Richard Höchenberger`_
+* `Eric Larson`_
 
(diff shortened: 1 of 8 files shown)


sappelhoff:
Thanks for having an eye on this @larsoner

hoechenberger (approved):

PR #1137, opened 2023-05-10
MRG: Add sessions as official suffix

JonathanVHoecke:
PR Description
--------------
Addition of sessions as a suffix when reading BIDS dataset.
Sessions are a suffix since BIDS v1.8
See:
https://bids-specification.readthedocs.io/en/stable/glossary.html#sessions-suffixes

I tested the code to successfully read in sessions.tsv and sessions.json files.

best wishes
Jonathan Vanhoecke
ICN Lab - Charité Universitätsmedizin Berlin


sappelhoff (commented):
@JonathanVHoecke I restarted the CI and apart from the doc-build, everything passes now. No idea why, but circleci somehow blocks execution for your PR. 

Could you please elaborate on what your changes enable, why you need that, and what other adjustments may be needed (if any)?

Thanks!

hoechenberger:
We will also need to add a test case before merging 👍

JonathanVHoecke:
Hi @sappelhoff and @hoechenberger 

My change was very small by just adding "sessions" into mne_bids/config.py under the allowed suffixes. No other changes were needed.
Why I need that, is because we use the sessions.tsv file to store session-specific information, such as the UPDRS score and other clinical scores as measured on that day.
I did not see any other need for other changes.

For the test-case, I added a sessions.tsv and sessions.json file in your tiny_bids dataset.
Further, I recycled your test_handle_scans_reading to be the same for test_handle_sessions_reading

I hope this is an answer to your question
best wishes
Jonathan

sappelhoff:
Thanks for the explanations :-) I don't really get the test case --> if `sessions.tsv` wasn't even considered a valid file for mne-bids until this PR, how can you add a test case that requires mne-bids to take changes in some `sessions.tsv` file into account? Or do I misunderstand something here?

Furthermore, could you please take care that the CI checks pass? (Except for the doc build, which --as said before-- somehow does not trigger for you)

hoechenberger:
Regarding my proposal to add a test, I was merely thinking of a one-liner where a `BIDSPath` with a `session` suffix is being created :)

JonathanVHoecke:
Dear @sappelhoff and @hoechenberger 

Thank you so much for your feedback. It is the about the 5th time I try to make a contribution to Github (Fieldtrip/BIDSvalidator/mne-bids), so I am still eager to learn about this process. Thanks for your remarks.

Concerning the tests that failed, I now adhered to the style standards.
Further, it is not entirely clear to me, where exactly the test dataset is being read from, as it is missing the sessions.tsv indeed
Hence, I wrote a test that wrote session.tsv, and then read and altered it, just the way that was done with the test  test_handle_scans_reading for scans.tsv

For any questions or feedback, please let me know
Best wishes
Jonathan

sappelhoff:
btw: I think the reason why circleCI is not being triggered is due to this setting: 

image

@JonathanVHoecke can you confirm that you don't have a circleci account? 

@larsoner why do we have this setting?

more info: https://support.circleci.com/hc/en-us/articles/4413968995099-How-to-set-up-the-Prevent-unregistered-user-spend-feature

drammock:
> why do we have this setting?

because we pay for our CircleCI service, and if a user fails to do basic doc build debugging locally (instead pushing one commit at a time trying to use the CIs to debug) it will use up I think 20ish minutes for each normal doc build, and ~2 hours if they push with `[circle full]`.

JonathanVHoecke:
@sappelhoff. Thank you to get back to me. I confirm I do not have a CircleCI account.

sappelhoff:
Thanks for the answer @drammock.

@JonathanVHoecke if you want the circleci check to run for your PR (which would be nice), this means that you'd need to:

1. create a circleci account (you can simply log in with your GitHub credentials)
2. *always* check that your docs build fine locally (see the [contributing guide section](https://github.com/mne-tools/mne-bids/blob/main/CONTRIBUTING.md#building-the-documentation))
3. take care not to push each commit individually, but rather as coherent groups of commits (in order to save time that the CI is running)

As for updating the `tiny_bids` dataset, the way to go is to:

1. update this script: https://github.com/mne-tools/mne-bids/blob/9394395f2d768d73eb6b09f239f65f9721962d24/mne_bids/tests/data/tiny_bids/code/make_tiny_bids_dataset.py#L1
4. run the script with your up-to-date installation of mne-bids
5. commit the resulting changes (confirming that there are no unexpected changes to the `tiny_bids` dataset)

--> this would be the workflow as opposed to just adding a file like you did here.

I don't really see why the `sessions.tsv` file was missing in the CI. **HOWEVER** ...

as @hoechenberger says in his [comment above](https://github.com/mne-tools/mne-bids/pull/1137#issuecomment-1544679537), we only need a much smaller test for the change you propose here.

One potentially good location would be here: https://github.com/mne-tools/mne-bids/blob/9394395f2d768d73eb6b09f239f65f9721962d24/mne_bids/tests/test_path.py#L410-L423

As a hint for you, I found this location by:

1. looking for where the tests for `BIDSPath` are located
2. searching in that file for `suffix` and checking the different tests for something that looked like it checked suffixes in a general way

Practically:

- please follow my instructions above for getting the circleci check to run (unless you really don't want a circleci account, in which case we'll have to see)
- remove your new test and your other changes that we don't need anymore
- add a tiny tiny test at my suggested location, as recommended by Richard

hoechenberger:
I just realized that there's a potential problem with the specification of those `*_sessions` files. A proper BIDS path **must** contain the entities `sub` and `task`. The example provided by @JonathanVHoecke with a path `sub-01_sessions.tsv` doesn't fit into this scheme. Now, how should we proceed?

mne_bids/tests/test_read.py
```diff
@@ -49,7 +49,7 @@
 _bids_path_minimal = BIDSPath(subject=subject_id, task=task)
 
 # Get the MNE testing sample data - USA
-data_path = testing.data_path(download=False)
+data_path = testing.data_path(download=True)
```
hoechenberger, line 52:
```suggestion
data_path = testing.data_path(download=False)
```

mne_bids/tests/test_read.py
```diff
@@ -1288,9 +1345,9 @@ def test_channels_tsv_raw_mismatch(tmp_path):
     raw.save(raw_path, overwrite=True)
 
     with pytest.warns(
-        RuntimeWarning,
-        match=f'Cannot set "bad" status for the following channels, as '
-              f'they are missing in the raw data: {ch_name_orig}'
+            RuntimeWarning,
+            match=f'Cannot set "bad" status for the following channels, as '
+                  f'they are missing in the raw data: {ch_name_orig}'
```
hoechenberger, line 1350:
```suggestion
        RuntimeWarning,
        match=f'Cannot set "bad" status for the following channels, as '
              f'they are missing in the raw data: {ch_name_orig}'
```

mne_bids/tests/test_read.py
```diff
@@ -1255,9 +1312,9 @@ def test_channels_tsv_raw_mismatch(tmp_path):
     raw.save(raw_path, overwrite=True)
 
     with pytest.warns(
-        RuntimeWarning,
-        match='number of channels in the channels.tsv sidecar .* '
-              'does not match the number of channels in the raw data'
+            RuntimeWarning,
+            match='number of channels in the channels.tsv sidecar .* '
+                  'does not match the number of channels in the raw data'
```
hoechenberger, line 1317:
```suggestion
        RuntimeWarning,
        match='number of channels in the channels.tsv sidecar .* '
              'does not match the number of channels in the raw data'
```

sappelhoff:
> Since those files are stored one level up in the folder hierarchy, as compared to the actual recordings, maybe it's okay for them not to bear the task entity.

That's how I read it

hoechenberger:
@sappelhoff I believe this one should be ready for review.

sappelhoff (approved):
LGTM! Thanks :-)

hoechenberger:
thanks @JonathanVHoecke!

Issue #1155, opened 2023-07-31
Problems with conversion of BTI datasets


nugenta:
### Description of the problem

There are two problems with the conversion of BTI/4D data to BIDS format. The first is that if there is no headshape data, mne_bids.write_raw_bids will fail (in function copyfile_bti_mod). 

The second issue is both an MNE bids and an MNE issue - not sure if I should replicate this issue there? When reading in a 4D dataset, raw.__init_kwargs is correctly populated with the full path for the dataset, i.e.

raw._init_kwargs['pdf_fname'] = '/path_to_file/4D/c,rfDC'

but the field for the config file

raw._init_kwargs['config_fname'] = 'config'

Without the full path to the config file, copyfile_bti_mod will fail. 

### Steps to reproduce

```Python
#requires definition of the 'meg_fname' filepath and 'topdir' filepath

raw=mne.io.read_raw_bti(meg_fname, head_shape_fname=None, convert=False)
bids_path = BIDSPath(subject=subjid, session='1', task='rest',
                    run='01', root=topdir, suffix='meg')
write_raw_bids(raw, bids_path, overwrite=True)
```

### Expected results

I expect it to copy the files to the BIDS directory rather than erroring out

### Error because full path for config file is not specified:

Traceback (most recent call last):

  Cell In[7], line 48
    write_raw_bids(raw, bids_path, overwrite=True)

  File <decorator-gen-640>:12 in write_raw_bids

  File ~/mambaforge/envs/enigma_meg/lib/python3.11/site-packages/mne_bids/write.py:1979 in write_raw_bids
    copyfile_bti(raw_orig, raw_dir)

  File ~/mambaforge/envs/enigma_meg/lib/python3.11/site-packages/mne_bids/copyfiles.py:600 in copyfile_bti
    sh.copyfile(raw._init_kwargs['config_fname'],

  File ~/mambaforge/envs/enigma_meg/lib/python3.11/shutil.py:256 in copyfile
    with open(src, 'rb') as fsrc:

FileNotFoundError: [Errno 2] No such file or directory: 'config'

### Additional information

Using mne version 1.4.2 and mne-bids 0.12

sappelhoff:
Hi, thanks for the report!

> The second issue is both an MNE bids and an MNE issue - not sure if I should replicate this issue there? When reading in a 4D dataset, raw.__init_kwargs is correctly populated with the full path for the dataset, i.e.

please open an MNE-Python issue on the MNE-Python repository 👍 

[...]

PR #1160, opened 2023-08-02
Fix BTi handling in mne-bids

sappelhoff:
- xref: #1155
- xref: https://github.com/mne-tools/mne-python/pull/11851

I think BTI copying was broken all this time in MNE-Python, I found several things that don't make any sense, and seemingly we never even had a test for this.

I now introduced a test. this code will only work with mne >= 1.5

---

@nugenta we need your help to test this fix.

Can you please:

1. install the current development version of `mne`
2. install the fix from this branch? see instructions for installing from a branch here: https://github.com/mne-tools/mne-bids/pull/1158#issuecomment-1659939926

and then run your pipeline?

Are all your issues fixed?

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -638,7 +638,7 @@ def copyfile_bti(raw, dest):
             val = op.basename(keyfile)
 
         # If no headshape file present, cannot copy it
-        if key == "head_shape_fname" and keyfile is None:  # pragma: no cover
+        if key == "head_shape_fname" and keyfile is None:
             continue
 
         sh.copyfile(keyfile, op.join(dest, val))
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -619,10 +619,10 @@ def copyfile_bti(raw, dest):
     """
     # XXX: remove after support for mne<1.5 is dropped
     # BTi operations were unreliable in mne-bids prior to mne 1.5 / mne-bids 0.13
-    need_mne = "1.3"
-    if _compare_version(mne.__version__, "<", need_mne):
+    need_mne = "1.4.2"
+    if not _compare_version(mne.__version__, ">", need_mne):  # pragma: no cover
         raise ImportError(
-            f"mne >= {need_mne} is required for BTi operations, got {mne.__version__}"
+            f"mne > {need_mne} is required for BTi operations, got {mne.__version__}"
         )
 
     os.makedirs(dest, exist_ok=True)
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -16,18 +16,19 @@ due to internal pointers that are not being updated.
 # License: BSD-3-Clause
 import os
 import os.path as op
+from pathlib import Path
 import re
 import shutil as sh
-from pathlib import Path
 
-from scipy.io import loadmat, savemat
-
+import mne
 from mne.io import read_raw_brainvision, read_raw_edf, read_raw_bdf, anonymize_info
 from mne.utils import logger, verbose
+from mne.fixes import _compare_version
+import numpy as np
+from scipy.io import loadmat, savemat
 
 from mne_bids.path import BIDSPath, _parse_ext, _mkdir_p
 from mne_bids.utils import _get_mrk_meas_date, _check_anonymize, warn
-import numpy as np
 
 
 def _copytree(src, dst, **kwargs):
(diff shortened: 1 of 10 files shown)


mne_bids/copyfiles.py
```diff
@@ -616,13 +617,28 @@ def copyfile_bti(raw, dest):
...
-    hs_file = raw._init_kwargs.get("head_shape_fname")
-    if hs_file is not None:  # pragma: no cover
-        sh.copyfile(hs_file, op.join(dest, "hs_file"))
+    # XXX: remove after support for mne<1.5 is dropped
+    # BTi operations were unreliable in mne-bids prior to mne 1.5 / mne-bids 0.13
+    need_mne = "1.3"
+    if _compare_version(mne.__version__, "<", need_mne):
+        raise ImportError(
+            f"mne >= {need_mne} is required for BTi operations, got {mne.__version__}"
+        )
```
sappelhoff, line 626: @larsoner this `_compare_version` works in unexpected ways.

MNE's version is currently something like: `'1.4.0.dev224+g0fd4ded6b'`.

IMHO this should work (evaluate `True`): `_compare_version(mne.__version__, "<", 1.4):` ... because we're in the dev branch AFTER the 1.4 release. However, it doesn't evaluate to `True` (hence me setting it to 1.3 in the code).

Another question. When I make release in other softwares (including mne-bids), I:

- release a stable version (e.g., 0.12)
- then start developing the next version (e.g., 0.13)
- hence my dev version is called 0.13.dev0
- if I release 0.12.1 (and not 0.13) in the meantime, I want to switch back to a new dev version
- that would be called 0.13.dev1 ... and so on

do I get something wrong here? Where does mne's versioning scheme fit in here?

hoechenberger: IMHO the check is correct **but** the version number is not.

It should not be `1.4.0.devX` which AFAIK indicates a dev version prior to a 1.4.0 release:

> The developmental release segment consists of the string .dev, followed by a non-negative integer value. Developmental releases are ordered by their numerical component, immediately before the corresponding release (and before any pre-releases with the same release segment), and following any previous release (including any post-releases).
> 
> Developmental releases are also permitted for pre-releases and post-releases:
> 
> X.YaN.devM       # Developmental release of an alpha release
> X.YbN.devM       # Developmental release of a beta release
> X.YrcN.devM      # Developmental release of a release candidate
> X.Y.postN.devM   # Developmental release of a post-release
> 

https://peps.python.org/pep-0440/#developmental-releases

So the version number of our current dev version should be `1.4.0.post0.dev0`

sappelhoff: okay thanks @hoechenberger this is also my understanding

hoechenberger: > Or `1.5.0.dev0`, which I wouldn't know how to do with `setuptools-scm` without adding a new tag

Actually this should happen automatically:

https://github.com/pypa/setuptools_scm#default-versioning-scheme

> The next version is calculated by adding 1 to the last numeric component of the tag.


so we're screwing that up somehow.

hoechenberger: @sappelhoff I ran pip install again and now the version number seems correct…

sappelhoff: why? What changed? I am still getting `mne-1.4.0.dev237+ga6c07cb5c` after a fresh pull from upstream and an editable pip install

hoechenberger: May have to do with the fact that we're writing the version to `mne/_version.py` during install
And after that it doesn't get updated anymore

sappelhoff: what's the version you are getting?

hoechenberger: Now I'm getting 1.4 again … ummm … I have no idea what's going on

And in mne/_version.py it still says

```
__version__ = version = '1.5.0.dev88+ga6c07cb5c'
__version_tuple__ = version_tuple = (1, 5, 0, 'dev88', 'ga6c07cb5c')
```

sappelhoff: :thinking: 
```
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.4.0.dev237+ga6c07cb5c'
__version_tuple__ = version_tuple = (1, 4, 0, 'dev237', 'ga6c07cb5c')
```

larsoner: Yikes something is wrong:
```
$ pip install --no-build-isolation -ve .
...
Successfully installed mne-1.5.0.dev75+g0fd4ded6b
$ python -c "import mne; print(mne.__version__)"
1.4.2
$ python -c "import mne; print(mne)"
<module 'mne' from '/Users/larsoner/python/mne-python/mne/__init__.py'>
$ cat mne/_version.py 
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
__version__ = version = '1.5.0.dev75+g0fd4ded6b'
__version_tuple__ = version_tuple = (1, 5, 0, 'dev75', 'g0fd4ded6b')
```
I wonder if this has been broken ever since we release 1.4.1 / 1.4.2. I'll look a bit...

hoechenberger: I believe it's the call to

```
    from importlib.metadata import version

    __version__ = version("mne")
```

in MNE's `__init__.py` that's the problem

larsoner: For me the presence of a `mne.egg-info` made it mess up the search when in the MNE-Python directory:
```
mne-python larsoner $ rm -Rf mne.egg-info
mne-python larsoner$ python -c "import mne; print(mne, mne.__version__)"
<module 'mne' from '/Users/larsoner/python/mne-python/mne/__init__.py'> 1.5.0.dev75+g0fd4ded6b
```
I figured this out by looking at and seeing something like this
```
>>> from importlib.metadata import distribution, MetadataPathFinder
>>> d = distribution("mne")
>>> d._path
PosixPath('mne.egg-info')
>>> list(x._path for x in MetadataPathFinder.find_distributions() if "mne" == x.name)
[PosixPath('mne.egg-info'), PosixPath('/Users/larsoner/opt/miniconda3/envs/mne/lib/python3.10/site-packages/mne-1.5.0.dev75+g0fd4ded6b.dist-info')]
```

sappelhoff: I can't reproduce these results, I keep getting `1.4.0.dev237+ga6c07cb5c`

larsoner: @sappelhoff my only idea is maybe you haven't fetched the MNE tags from `upstream` in a while. The version is built off the tags in the repo.

If this fails, too, then try `git clone`ing just upstream and doing an editable install. If this also fails, maybe your `setuptools` or `setuptools_scm` is too old. Have you tried upgrading these and doing `pip install --no-build-isolation -ve .` or so? Then it will use your locally installed versions of these deps rather than isolating them.

larsoner: I don't have this file on my Linux machine (the above offender was working on my macOS machine) and the CIs haven't had this problem, so I'm hoping it's something transient / leftover on our machines from a while ago. I even tried `python setup.py bdist_wheel sdist` locally on my Linux machine and couldn't reproduce the problem. So hopefully it's something we can ignore and won't be a problem going forward, but if we see it happen again and can figure out the root cause we should definitely fix it!

larsoner: ... actually doing `python setup.py bdist_wheel sdist` *did* create the file on my Linux machine! So that's something to look into. I'll create a new issue in MNE about this.

nugenta:
Hi - sorry for only getting to this now. Yes, installing the latest devel versions of both mne and mne-bids have solved the BTi copying problems.  Thanks!

sappelhoff:
> Hi - sorry for only getting to this now. Yes, installing the latest devel versions of both mne and mne-bids have solved the BTi copying problems. Thanks!

no worries, thanks @nugenta ... but did you install mne-bids from this branch? ... or just the mne-bids development version?

anyhow, I am going to merge this now. Please update your mne-bids development version and tell us if it still works -- thanks!

PR #1161, opened 2023-08-02
test minimal vs full mne-bids install

sappelhoff:
explicitly testing a full and a minimal mne-bids install.

I think in practice this currently does not make a difference, but I think it's good to have the infrastructure here.


hoechenberger:
Hot take:

When I install software, I expect all features to work unless I explicitly opted for some stuff not to be installed.

Instead of a `[full]` classifier, there should be `[minimal]` for those who don't care about "full" functionality.

I know setuptools doesn't support this and it's a pity.

PR #1062, opened 2022-09-01
Fix BIDSPath.basename for cases where only an extension was provided

hoechenberger:
```python
#%%
from mne_bids import BIDSPath
bp = BIDSPath(extension='.vhdr')
bp
```

`main`:
Screen Shot 2022-09-01 at 20 55 03

this PR:
Screen Shot 2022-09-01 at 20 55 19

Merge checklist
---------------

Maintainer, please confirm the following before merging.
If applicable:

- [ ] All comments are resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)
- [ ] PR description includes phrase "closes <#issue-number>"

PR #1163, opened 2023-08-07
MRG: Fix BIDSPath.basename for cases where only an extension was provided

hoechenberger:
This supersedes the old PR #1062

Closes #1062


sappelhoff:
ok thank you for trying @hoechenberger

PR #1215, opened 2024-01-10
Do not zero-pad indices in BIDSPath entities anymore; do not warn about prepending missing dots for extensions

alexrockhill:
PR Description
--------------

I often have subject ids like `1` and `2` and I don't always store them as strings. I think it would be nice if `mne_bids` just handled this to allow for integer subject ids, runs, sessions, acquisitions etc.

Merge checklist
---------------

If applicable:

- [x] This is not your own PR
- [x] All CIs are happy
- [x] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -958,34 +958,21 @@ class BIDSPath:
                     f"Key must be one of " f"{ALLOWED_PATH_ENTITIES}, got {key}"
                 )
 
+            if isinstance(val, int):
+                kwargs[key] = f"{val:02}"
+
             if ENTITY_VALUE_TYPE[key] == "label":
-                if isinstance(val, int):
-                    val = str(val)
                 _validate_type(val, types=(None, str), item_name=key)
             else:
                 assert ENTITY_VALUE_TYPE[key] == "index"
                 _validate_type(val, types=(int, str, None), item_name=key)
                 if isinstance(val, str) and not val.isdigit():
                     raise ValueError(f"{key} is not an index (Got {val})")
-                elif isinstance(val, int):
-                    kwargs[key] = f"{val:02}"
 
         # ensure extension starts with a '.'
         extension = kwargs.get("extension")
         if extension is not None and not extension.startswith("."):
-            warn(
-                f'extension should start with a period ".", but got: '
-                f'"{extension}". Prepending "." to form: ".{extension}". '
-                f"This will raise an exception starting with MNE-BIDS 0.12.",
-                category=FutureWarning,
-            )
             kwargs["extension"] = f".{extension}"
-            # Uncomment in 0.12, and remove above code:
-            #
-            # raise ValueError(
-            #     f'Extension must start wie a period ".", but got: '
-            #     f'{extension}'
-            # )
         del extension
 
         # error check entities
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -965,7 +965,7 @@ class BIDSPath:
                 _validate_type(val, types=(int, str, None), item_name=key)
                 if isinstance(val, str) and not val.isdigit():
                     raise ValueError(f"{key} is not an index (Got {val})")
-                else:
+                elif isinstance(val, int):
                     kwargs[key] = f"{val}"
 
         # ensure extension starts with a '.'
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -966,7 +966,7 @@ class BIDSPath:
                 if isinstance(val, str) and not val.isdigit():
                     raise ValueError(f"{key} is not an index (Got {val})")
                 elif isinstance(val, int):
-                    kwargs[key] = f"{val:02}"
+                    kwargs[key] = f"{val}"
 
         # ensure extension starts with a '.'
         extension = kwargs.get("extension")
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -959,6 +959,8 @@ class BIDSPath:
                 )
 
             if ENTITY_VALUE_TYPE[key] == "label":
+                if isinstance(val, int):
+                    val = str(val)
                 _validate_type(val, types=(None, str), item_name=key)
             else:
                 assert ENTITY_VALUE_TYPE[key] == "index"
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -966,7 +966,7 @@ class BIDSPath:
                 if isinstance(val, str) and not val.isdigit():
                     raise ValueError(f"{key} is not an index (Got {val})")
                 elif isinstance(val, int):
-                    kwargs[key] = f"{val}"
+                    kwargs[key] = f"{val:02}"
 
         # ensure extension starts with a '.'
         extension = kwargs.get("extension")
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -958,9 +958,6 @@ class BIDSPath:
                     f"Key must be one of " f"{ALLOWED_PATH_ENTITIES}, got {key}"
                 )
 
-            if isinstance(val, int):
-                kwargs[key] = f"{val}"
-
             if ENTITY_VALUE_TYPE[key] == "label":
                 _validate_type(val, types=(None, str), item_name=key)
             else:
@@ -968,6 +965,8 @@ class BIDSPath:
                 _validate_type(val, types=(int, str, None), item_name=key)
                 if isinstance(val, str) and not val.isdigit():
                     raise ValueError(f"{key} is not an index (Got {val})")
+                else:
+                    kwargs[key] = f"{val}"
 
         # ensure extension starts with a '.'
         extension = kwargs.get("extension")
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -959,7 +959,7 @@ class BIDSPath:
                 )
 
             if isinstance(val, int):
-                kwargs[key] = f"{val:02}"
+                kwargs[key] = f"{val}"
 
             if ENTITY_VALUE_TYPE[key] == "label":
                 _validate_type(val, types=(None, str), item_name=key)
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -50,7 +50,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- nothing yet
+- Allow integer types to be passed to :class:`mne_bids.BIDSPath` by `Alex Rockhill`_ (:gh:`1215`)
 
 ⚕️ Code health
 ^^^^^^^^^^^^^^
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,6 +40,10 @@ Detailed list of changes
 
 - The experimental support for running MNE-BIDS examples from your browser using Binder has
   been removed, by `Stefan Appelhoff`_ (:gh:`1202`)
+- MNE-BIDS will no longer zero-pad ("zfill") entity indices passed to :class:`~mne_bids.BIDSPath`.
+  For example, If ``run=1`` is passed to MNE-BIDS, it will no longer be silently auto-converted to ``run-01``, by `Alex Rockhill`_ (:gh:`1215`)
+- MNE-BIDS will no longer warn about missing leading punctuation marks for extensions passed :class:`~mne_bids.BIDSPath`.
+  For example, MNE-BIDS will now silently auto-convert ``edf`` to ```.edf``, by `Alex Rockhill`_ (:gh:`1215`)
 
 🛠 Requirements
 ^^^^^^^^^^^^^^^
@@ -53,7 +57,6 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- Allow integer types to be passed to :class:`mne_bids.BIDSPath`, by `Alex Rockhill`_ (:gh:`1215`)
 - The datatype in the dataframe returned by :func:`mne_bids.stats.count_events` is now
   ``pandas.Int64Dtype`` instead of ``float64``, by `Eric Larson`_ (:gh:`1227`)
 
--- a/mne_bids/tests/test_path.py
+++ b/mne_bids/tests/test_path.py
@@ -352,10 +352,10 @@ def test_parse_ext():
 @pytest.mark.parametrize(
     "fname",
     [
-        "sub-01_ses-02_task-test_run-3_split-01_meg.fif",
-        "sub-01_ses-02_task-test_run-3_split-01",
-        "/bids_root/sub-01/ses-02/meg/sub-01_ses-02_task-test_run-3_split-01_meg.fif",  # noqa: E501
-        "sub-01/ses-02/meg/sub-01_ses-02_task-test_run-3_split-01_meg.fif",
+        "sub-01_ses-02_task-test_run-3_split-1_meg.fif",
+        "sub-01_ses-02_task-test_run-3_split-1",
+        "/bids_root/sub-01/ses-02/meg/sub-01_ses-02_task-test_run-3_split-1_meg.fif",  # noqa: E501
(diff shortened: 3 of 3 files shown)


alexrockhill:
Hmmm it looks like this was by design that integers were excluded, there's even a test (failing in this case) that `subject=1` fails. I find this a bit annoying because I don't understand the need for IDs like `"01"` on modern computing systems, it seems like a design choice that some but not all people make that can lead to Y2K phenomena if for instance you end up collecting more than 100 or 1000 subjects when you started with a "00" or "000" system. Is there a rationale for blocking users from doing this? To me it seems overly pedantic.

alexrockhill:
image

Given that it's the `"01"` way on the homepage of BIDS, I guess this is fairly used but I don't recall seeing an explanation of why you need to format them this way. With the key-value tags in BIDS, this seems superfluous and from an era when strings were hardcoded to a certain length which is a brittle design choice for long identifiers..

alexrockhill:
Also if you pass `run=1` it automatically gets mapped to `"01"` which seems very odd and unexpected.

hoechenberger:
I agree:

- we should accept integers and simply convert them to strings
- we should not prepend zeros (zfill)
- if a user wants to zfill, they can do this manually and pass the resulting string

alexrockhill:
Nice, okay, I thought I was alone in thinking this was a bit silly and that everyone else labeled their subjects this way for a minute. If the consensus is that this is that the labels should be less constrained, I'm happy to add that to the PR.

hoechenberger:
@alexrockhill We better double-check check the BIDS specs before we proceed here
Otherwise I'm +1

alexrockhill:
https://github.com/bids-standard/bids-specification/discussions/1679

Remi-Gau:
Note that in BIDS runs are indices (i.e must be a digit) and subjects, sessions (and most other entities) are labels and can be alphanumeric.

https://bids-specification.readthedocs.io/en/latest/appendices/entities.html#sub

https://bids-specification.readthedocs.io/en/latest/appendices/entities.html#run

This partly explains why you can assume:

run-01 --> 1

but not

sub-01 --> 1 because you could also have sub-1

Hope this helps a bit.

jasmainak:
I think @Remi-Gau is on the right track. It's why we disabled integers so that users are explicit and don't mix `sub-01` and `sub-1`

hoechenberger:
@alexrockhill I think if you pass strings, no alterations will take place. So this would be the easiest workaround I suppose?

alexrockhill:
> I think @Remi-Gau is on the right track. It's why we disabled integers so that users are explicit and don't mix `sub-01` and `sub-1`

I understand the principle but I don't understand the use case. Who labels their subjects `1` and `01`?

My preference would be to allow integers and cast them to strings since `subject=1` and `subject="1"` seem like reasonable subject identifiers to me and would mean the same subject. Since it's just annoying in code to cast everything e.g. `subject=str(subject)` and so it seems convenient to do this within `mne_bids` to make it as usable as possible. In general, I will say the annoying thing about BIDS is that it can sometimes feel like death by a million paper cuts where when you try and use it you get caught up on a bunch of little things like this and each one alone is rather silly to complain about but the end result is that people think it's more of a pain than it is. For example, in this context, figuring out if each tag is an index or a label (i.e. accepts int or string), having to remember if `extension=".edf"` is right or  `extension="edf"` is right or having to use `extension` and `suffix` even though if you pass `datatype` to `read_raw_bids` so the file type is able to be inferred (I know internally it's this way because `BIDSPath` isn't just used for `read_raw_bids` but also modified to find sidecars but that doesn't really matter from an end-user perspective). Anyway, the point being, this is one of several pain points that I think, especially for `write_raw_bids` and `read_raw_bids` is pretty mission-critical for adoption so that people don't run into an error that isn't one that is essential to ensuring that the dataset contains all the necessary information.

The `run` one is just a bug, that seems pretty simple. You should either not be allowed to pass integers or `run=1` should be cast to `run="1"` but definitely not `run="01"`.

alexrockhill:
I'd actually go further if it was entirely my design choice and I would match `subject=1` to `subject="01"` and only throw an error at run time if there are both `subject="01"` and `subject="1"`. To me, it seems way more practical and common of a use case to have incrementing subject indices and I think you'd actually want an error when you try and make `subject="1"` and `subject="01"` that seems like a situation that should not be allowed, it's too confusing.

alexrockhill:
Just wanted to add, I'm not trying to force this PR through, this is just my vision and addresses what I perceive to be pain points in the use case. Happy to discuss or close if people don't want to discuss and to just keep things how they are.

hoechenberger:
I agree our API is hard to use.

jasmainak:
I was just trying to explain what might have been a possible reason. It's important not to cause a regression bug. I'll let you figure out the best path forward with @hoechenberger !

agramfort:
Allowing different ways to be valid means adding more magic to the code. You need to multiply test to support 1 or str 01 and also tests of if both are subjects etc. You create more corner cases in the code and complexity is exponential with the number of options. Any bad default you point out was maybe a natural default for someone with a different prior and this person did not complain here. I am not trying to force the status quo here but be mindful of these arguments.

alexrockhill:
Totally, I asked why it was that way so the explanation is very helpful!

I think this would be backward compatible except if someone upgrades in the middle of converting a dataset which they aren't supposed to do.

It will silently return `subject="01"`  if you have a dataset with `subject="01"` and `subject="1"` which I think is reasonable, again I hope this isn't a use case. But otherwise, this PR just takes integers and casts them to the BIDS recommendation which is the best of both worlds, I think (i.e. I just say `subject=1` but that gets translated to `subject="01"` per BIDS recommendation). This obviates the need for the user to even think about zero padding which I would argue they probably don't want to and just want things to work and give readable results.

alexrockhill:
@agramfort agreed but there are only two cases so it seems manageable.

hoechenberger:
I'm not decided whether I want to change anything :) I just wanted to say that I think our API is not very easy to use :)

sappelhoff:
I agree that it's important to keep in mind what @agramfort said here: https://github.com/mne-tools/mne-bids/pull/1215#issuecomment-1887950194

If it were just me, I'd maybe improve the documentation about this, but I wouldn't start adding more "magic conversion" behind the scenes. The present behavior has been in place for a very long time and we rarely get issues about it (we did however, admittedly, get one about it not too long though: https://mne.discourse.group/t/remove-automatic-zero-padding-when-specifing-run-in-bidspath/7865/4?u=sappelhoff

hoechenberger:
Also we kept on **removing** "magic" over the past years, because it would cause all sorts of problems

alexrockhill:
My point was just that it makes sense to be consistent and have this magic everywhere (`subject` as well as `run`) if we want to enforce the recommendation or to have it nowhere.

Again, per the discourse use case comment/issue, people try and pass `int`s and by not handling it consistently, we're making them dive into these kinds of discussions zero padding etc. that are going to make people think BIDS is too much of a work-in-progress for _them_ to use currently.

So, if it's about removing magic, then maybe just a straight cast to string for integers with no modification.

I think people are going to pass integers though and throwing them an error doesn't seem helpful since the information is fundamentally there.

alexrockhill:
So now all the magic is removed from `run` and `split` as well which solves the discourse issue but will be slightly backward incompatible because if someone has `run=1` in their code, they'll have to change it to `run="01"` to access an already-converted dataset.

To do (in follow-up PRs):
- [ ] Remove matches to sidecars in `read_raw_bids` so that it doesn't have to be over-specified

mne_bids/path.py
```diff
@@ -958,32 +958,21 @@ def update(self, *, check=None, **kwargs):
...
-                f"This will raise an exception starting with MNE-BIDS 0.12.",
-                category=FutureWarning,
-            )
             kwargs["extension"] = f".{extension}"
-            # Uncomment in 0.12, and remove above code:
-            #
-            # raise ValueError(
-            #     f'Extension must start wie a period ".", but got: '
-            #     f'{extension}'
-            # )
```
agramfort, line 986: can you see in the PR tha introduced this what was the rationale for this warning? I am sure it was discussed.

agramfort (commented):
can this PR break any working code? I see you touching tests. Can you see if it breaks eg the mne-bids-pipeline by running the tests locally for on this branch?

mne_bids/path.py
hoechenberger: According to the BIDS specs, extensions start with a period, see e.g.
https://bids-specification.readthedocs.io/en/stable/appendices/meg-file-formats.html#neuromagelektamegin

I'm therefore -1 on this change for now. It's a tradeoff between usability and following the standard closely…

alexrockhill: Thanks for sharing that. This is where I get really confused on BIDS and the interaction with the user. I think that is an important technical distinction for comparing across packages perhaps but other than that I really fail to see how that is relevant to an end user. Leading period, no leading period, the convention is clear that all file names will have a period before the extension whether it's in the extension or not. So why are we torturing the user by making them consider this minutia when it doesn't have any functional difference? I'll look at the PR for more reasoning but when I was learning BIDS, I spent about 10x the time learning about this irrelevant stuff really and only 1x learning about crucial sidecar files and their required and recommended fields. And to be honest, it was a terrible learning experience, I still often feel lost in all the different style recommendations and requirements (see for instance this zero padding) and I've been using and developing BIDS for years! To be even more honest, I don't actually use `mne_bids` to convert my datasets anymore, I pulled out and simplified each sidecar function generator and I write each sidecar at the appropriate time (i.e. I write events after I find the events, I write channels after I find the bad channels and I write electrodes after I find the positions), I find that way more intuitive and then I stop and helps me think about what's needed in each step and each sidecar. That just works much better for me and my workflow, I'm not saying it's better for everyone but these types of things drive me away from using the `mne_bids` API because it's such a simple and practice fix to just fix one error in the ten you might run into if you've never converted a dataset before and I think a lot of rationale people stop at 9 or less!

hoechenberger: Good points! So the idea is to add a period if none was supplied? Yes, why not! I didn't fully think this through before posting my comment above 

Could you perhaps describe your custom workflow in more detail?

Like I said elsewhere, I, too, believe the API is difficult to master and very inflexible. If we could divide the workflow into more small, independent sub-steps, things might become easier to grasp

alexrockhill: No worries, I'm just trying to be transparent and giving my 2 cents, I think it's very reasonable to default to the specification and I think you're doing a good job with the code I've seen you change, I just made this PR because I was tired of running into little errors like this.

For intracranial recordings, I find the events from a photodiode channel then save those to events, I find the contacts from a post op CT and save those to electrodes and I find the bad channels and save those to channels, not much more to it than that!

hoechenberger: I like the idea of allowing for a more "incremental" writing if the BIDS data. Like, maybe write the raw file right away after recording; update the events later after manual inspection; etc.

@sappelhoff WDYT? I'd be happy to brainstorm about this in a video call or the next time I come to Berlin :)

alexrockhill:
I'm getting a bunch of `pydantic` import errors when I went to test `mne_bids_pipeline`...

alexrockhill:
Could someone with a working version of `mne-bids-pipeline` test this and confirm that it doesn't break anything please?

larsoner:
An easy way if you don't want to set it up yourself would be to open a PR to mne-bids-pipeline where you change `pyproject.toml` to tell it to install your branch of mne-bids:

https://github.com/mne-tools/mne-bids-pipeline/blob/main/pyproject.toml#L49

Something like `mne-bids[full] @ git+https://github.com/alexrockhill/mne-bids@int` rather than just `mne-bids[full]` should do it.

alexrockhill:
Looks like it doesn't break anything in `mne-bids-pipeline` so looks good to go possibly

alexrockhill:
LGTM?

sappelhoff (commented):
as I said in https://github.com/mne-tools/mne-bids/pull/1215#issuecomment-1888707118, I am not convinced that we should make this change. So @hoechenberger or @agramfort should perhaps make a verdict on this, as they seemed more open for it.

hoechenberger:
Unfortunately I currently don't have the tine to really think through the implications of this change, so I'd like to withhold my vote for now

alexrockhill:
Oh I was under the impression we talked it through and came to an intermediate compromise everyone was happy with. I.e. remove magic zero padding for all fields (it was on run before) and allow int for all fields but just cast it to string. That includes no magic and changes current behavior in a small way: i.e. run=1 just goes to run='1' not run='01'.

alexrockhill:
Also worth noting, I'm really not opinionated as to whether `subject/run=1` goes to `subject/run="01"` or `subject/run='1'`, I think both are fine, I just made this PR because I find `mne_bids` quite annoying that it doesn't accept `subject=1` (and `extension='edf'` not being allowed is also very annoying).

sappelhoff:
I'll take some time and think through this within the next days, sorry for causing some frustration here.

alexrockhill:
No worries, take your time, not frustrated by you, just a bit annoyed by how `mne_bids` works but that's what PRs are for!

sappelhoff (approved):
I have gone over the proposed changes and adjusted the changelog and code as such:

- we will no longer zero-pad indices ... so run=1 will be run=1 instead of run=01
- we will now silently auto-prepend missing dots from extensions

I decided against accepting "int" for all possible inputs to BIDSPath. Having looked at it, we have a policy of accepting ints for BIDS `entity-<index>` entities, and strs for BIDS `entity-<label>` entities -- and I think that's clear and reasonable, and I would like to keep it as such.

The present changes solve some issues (like documented in: https://mne.discourse.group/t/remove-automatic-zero-padding-when-specifing-run-in-bidspath/7865?u=sappelhoff) without causing too much of a code base disruption.

This is good to go from my side as soon as tests pass. Please feel free to merge @hoechenberger

alexrockhill:
Thanks for getting back @sappelhoff , the changes look good and fix most of the pain points I was running into.

I understand the index/label distinction and the motivation between trying to have nice categories. However, in practice, I have observed that even for labels, people tend to use integers more than not (e.g. `sub-1`, `ses-1`). I feel like if people actually did use non-index labels, this wouldn't even be up for discussion because it would be very clear that of course you call your subject `A`, `B`, `C` etc. but that just isn't how the community has spontaneously organized, people picked indices.

I'll defer to your judgment though, thanks for discussing. Maybe it's an issue better taken upstream in the BIDS specification that that there even be this distinction in the first place since I feel it's a bit more aspirational than actually adopted, if I really care that much.

sappelhoff:
> However, in practice, I have observed that even for labels, people tend to use integers more than not

that's true, and that's also okay, because labels may be "alphanumeric" ... whereas indices are "non-negative integers".

(https://bids-specification.readthedocs.io/en/stable/common-principles.html#definitions)

You are right that people use `sub-<label>` as `sub-01` often. But I have rarely seen `task-<label>` to be `task-01` --> I think we don't really have a problem here. We **could** make a big PR and add the option to have int for each "label-entity", but I classify this as a high effort, low impact task.

Gonna merge now! :-) 

Thanks for the discussion, I think the changes here improve the user experience!

hoechenberger:
Thanks everyone!!

alexrockhill:
Sure, I think that's a good place to leave the discussion for later but I'll just add a last comment that because task=1 makes no sense and is never used and sub=1 is the most common use case (note 1 not '01' from what I've seen from the community at large, BIDS core devs tend to use '01' which I think is a bit of a disconnect with the broader community) so it doesn't make sense to me that those are the same category although I agree alphanumeric is a fitting umbrella category, I wouldn't call sub a label in the way that task is since sub tends to be an index in practice. Which is to say it's a totally understandable and workable difference in opinions and maybe I'll just save this thread for when this viewpoint wins out in 5 years :)

PR #1240, opened 2024-03-20
account for optional Z in time strs

sappelhoff:
PR Description
--------------

fix a problem that was reported here: https://mne.discourse.group/t/how-to-import-fnirs-data-from-a-local-dataset/8481/3?u=sappelhoff

cc @hoechenberger 

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -341,7 +341,7 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
             else:
                 acq_time += ".0"
 
-        dt_string = "%Y-%m-%dT%H:%M:%S.%fZ"
+        dt_string = "%Y-%m-%dT%H:%M:%S.%f"
         if acq_time_is_utc:
             dt_string += "Z"
 
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -327,10 +327,10 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
     # extract the acquisition time from scans file
     acq_time = acq_times[row_ind]
     if acq_time != "n/a":
-        # BIDS allows the time to be stored in UTC with a time-zone offset, which is
+        # BIDS allows the time to be stored in UTC with a zero time-zone offset, as
         # indicated by a trailing "Z" in the datetime string. If the "Z" is missing, the
         # time is represented as "local" time. We have no way to know what the local
-        # time zone is at the acquisition site; so we simply assume the same time zone
+        # time zone is at the *acquisition* site; so we simply assume the same time zone
         # as the user's current system (this is what spec demands anyway).
         acq_time_is_utc = acq_time.endswith("Z")
 
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -341,11 +341,11 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
             else:
                 acq_time += ".0"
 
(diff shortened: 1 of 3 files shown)


mne_bids/read.py
```diff
@@ -331,6 +331,8 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
         if "." not in acq_time:
             # acquisition time ends with '.%fZ' microseconds string
             acq_time += ".0Z"
+        if "Z" not in acq_time:
```
hoechenberger, line 334: I don't think we should do this. If the Z is not there, adding it will alter the time

and we don't want to cause disruptions to the time-space continuum, do we? 😳

hoechenberger (changes requested):

mne_bids/read.py
hoechenberger: Z indicates a time zone, it has nothing to do with the nanosecond (?) bit at the end of the time string per se

sappelhoff: > this also means that the line above is actually buggy too

you are absolutely right 😳  I simply followed the existing logic without questioning it.

hoechenberger: yes no worries, these datetime strings are anything but intuitive! I also had to google a bit first…

hoechenberger: Question is, *must* we have datetime in UTC in the end? What does MNE expect?

sappelhoff: I won't have time to think about this in the next days/weeks, but I have it on my to do list.

hoechenberger:
@sappelhoff I tried to give it a shot

Code isn't super pretty but I think the logical flow is very clear

Didn't run or write any tests yet

hoechenberger:
I added some tests

hoechenberger (approved):

mne_bids/read.py
```diff
@@ -327,12 +327,32 @@ def _handle_scans_reading(scans_fname, raw, bids_path):
     # extract the acquisition time from scans file
     acq_time = acq_times[row_ind]
     if acq_time != "n/a":
-        # microseconds in the acquisition time is optional
+        # BIDS allows the time to be stored in UTC with a zero time-zone offset, as
+        # indicated by a trailing "Z" in the datetime string. If the "Z" is missing, the
+        # time is represented as "local" time. We have no way to know what the local
+        # time zone is at the *acquisition* site; so we simply assume the same time zone
+        # as the user's current system (this is what spec demands anyway).
```
hoechenberger, line 334:
```suggestion
        # as the user's current system (this is what the spec demands anyway).
```

hoechenberger (approved):
I think this should be good to go. Another example of a poorly designed part of BIDS, unfortunately.

sappelhoff (commented):
LGTM! Thanks a lot! Would you mind adding a changelog entry?

sappelhoff:
Very kind of you to include me in the commit message 🤣 

Thanks, I think this is much saner behavior now!!

hoechenberger:
> Very kind of you to include me in the ~commit~ changelog message 🤣

yeah now worries, you started the PR after all :)

> Thanks, I think this is much saner behavior now!!

let's hope I / we didn't overlook something here … but I think in any case, we're behaving more correctly now than before!

PR #1243, opened 2024-04-02
Fix string format `"blabla" "bla"`

mscheltienne:
PR Description
--------------

I propose to change the `stacklevel` of a couple of warnings I encountered, because

```
<ipython-input-34-b78341704568>:1: RuntimeWarning: Converting data files to BrainVision format
```

is really not that informative and I would appreciate to quickly jump to the function issuing the warning.

While I was on it, I fixed probably all the occurrences of `"blabla" "bla"` strings.
I don't believe this minor maintenance PR needs a changelog entry.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/commands/mne_bids_raw_to_bids.py
+++ b/mne_bids/commands/mne_bids_raw_to_bids.py
@@ -30,7 +30,7 @@ def run():
     parser.add_option(
         "--subject_id",
         dest="subject_id",
-        help=("subject name in BIDS compatible format " "(01, 02, etc.)"),
+        help=("subject name in BIDS compatible format (01, 02, etc.)"),
     )
     parser.add_option(
         "--task", dest="task", help="name of the task the data is based on"
--- a/mne_bids/commands/mne_bids_raw_to_bids.py
+++ b/mne_bids/commands/mne_bids_raw_to_bids.py
@@ -30,7 +30,7 @@ def run():
     parser.add_option(
         "--subject_id",
         dest="subject_id",
-        help=("subject name in BIDS compatible format (01, 02, etc.)"),
+        help="subject name in BIDS compatible format (01, 02, etc.)",
     )
     parser.add_option(
         "--task", dest="task", help="name of the task the data is based on"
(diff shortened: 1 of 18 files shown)


mscheltienne:
OK, I did not notice this was using `mne.utils.warn` and not directly `warnings.warn`, so indeed the caller line is returned which can be informative.

In an ideal world, we would get both the caller line *and* the line where the warning was emitted (to quickly jump to the warning line and figure out why it was emitted in the first place). I use this second option very often.

Anyway, for now I removed that erroneous `stacklevel=2` part.

larsoner:
> In an ideal world, we would get both the caller line and the line where the warning was emitted (stacklevel=1, to quickly jump to the warning line and figure out why it was emitted in the first place). I use this second option very often.

In theory it should be a few lines to add an `add_caller=False` to `mne.utils.warn` that uses the adjacent stack frame calling function somehow like:
```
<ipython-input-blah>:1: RuntimeWarning: In call to mne_bids.something: Converting data files to BrainVision format
```
then mne-bids could make use of it.

Maybe we'd even want `add_caller=True` by default...?

mne_bids/commands/mne_bids_raw_to_bids.py
```diff
@@ -30,7 +30,7 @@ def run():
     parser.add_option(
         "--subject_id",
         dest="subject_id",
-        help=("subject name in BIDS compatible format " "(01, 02, etc.)"),
+        help=("subject name in BIDS compatible format (01, 02, etc.)"),
```
hoechenberger, line 33: I don't think the parentheses are needed here

mscheltienne:
@larsoner That's a nice idea, and would give the `caller` line, but only as text right? VSCode terminal does some nice magic to enable a `Ctrl + click` on the emitted warning to jump to the line.

Screenshot from 2024-04-02 15-41-23

With that proposition, the caller line would not benefit from the same 'link' to jump to the definition; and thus without this I don't see the added benefit compared to quickly searching for the error string within the codebase.

Anyway, I will give it a try and see how it renders ;)

mne_bids/commands/mne_bids_raw_to_bids.py
mscheltienne: I'm surprised `ruff` does not fix those automatically :/

sappelhoff (approved):
Thanks @mscheltienne!

PR #1251, opened 2024-05-19
fix inspect on macos

### LINKED ISSUE #1250: issue with CI and inspect module
@drammock @hoechenberger I think this is the failure that threw me off recently: https://github.com/mne-tools/mne-bids/actions/runs/9106401009/job/25033546552#step:20:1185

>  FAILED mne_bids/tests/test_inspect.py::test_inspect_single_file[True] - matplotlib._api.deprecation.MatplotlibDeprecationWarning: Auto-close()ing of figures upon backend switching is deprecated since 3.8 and will be removed in 3.10.  To suppress this warning, explicitly call plt.close('all') first.
FAILED mne_bids/tests/test_inspect.py::test_inspect_single_file[False] - UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown

_Originally posted by @sappelhoff in https://github.com/mne-tools/mne-bids/issues/1245#issuecomment-2114663602_

sappelhoff:
PR Description
--------------

Trying to fix #1250 ... currently no success locally, and this only seems to be a problem on macos. cc @hoechenberger 

- closes #1250

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy


hoechenberger:
@sappelhoff I can perhaps take a look later today

I'm still not convinced the inspect module is too useful these days anymore ... especially since it doesn't support the Qt browser :(

sappelhoff:
> I'm still not convinced the inspect module is too useful these days anymore ... especially since it doesn't support the Qt browser :(

I also think that the lack of support for the newish (and amazing) Qt browser is a drawback ... and I am also not sure whether we have a ton of users who like this mne-bids feature. And I also understand how that's a bit upsetting, as you invested quite a bit of thought and work into it.

I don't know whether this is where you were going with it, but we could think about cutting that part out of mne-bids, instead of continuing to maintain it. After all, it does say that the functionality is experimental:

> This functionality is still experimental and will be extended in the future. Its API will likely change. Planned features include automated bad channel detection and visualization of MRI images.

source: https://mne.tools/mne-bids/dev/generated/mne_bids.inspect_dataset.html

I personally would always use this workflow:

1. automatically mark bad channels and segments (e.g., via pyprep and other pipelines)
2. open the data in the QT browser, manually selecting or deselecting data features (e.g., if an automatically marked segment looks like a false positive, or if the automatic pipelines missed something)
3. save bad channels and data segments

For this workflow, the `mark_bad_channels` function would arguably be enough (bad channels are marked in `channels.tsv`). The "bad segments" could potentially be saved in the `events.tsv` files, but personally, I have been saving them in txt files under a `derivatives` folder (sub-optimal, I know).

--> i.e., for this workflow, the interactive inspection from within mne-bids is not needed, but perhaps rather a `mark_bad_segments` that edits an `events.tsv` file with what can be parsed from mne annotations.

Speaking of `mark_bad_channels`, I think there are some errors in the tutorial, do you agree? Look at this part:

https://mne.tools/mne-bids/dev/auto_examples/mark_bad_channels.html#non-interactive-programmatic-bad-channel-selection

and in specific these sections:

1. "If you instead would like to replace the collection of bad channels entirely, pass the argument overwrite=True:"

--> there is no `overwrite` argument for `mark_bad_channels`

2. "Lastly, if you’re looking for a way to mark all channels as good, simply pass an empty list as ch_names, combined with overwrite=True:"

--> the printed results show that rather than all channels being marked as good, they are marked as bad

sappelhoff:
@larsoner unfortunately your suggestion does not seem to do the trick, but thanks for having a look!

hoechenberger:
@sappelhoff I'm not emotionally attached to this feature ;) I'm totally okay with removing it. I'm only worried some folks might be using it – IIRC, @adam2392 once mentioned it's part of his workflow. Not sure about @SophieHerbst

PR #1283, opened 2024-07-28
Add new lint rules, remove unneeded noqa calls, refine noqa calls

sappelhoff:


doc/conf.py
```diff
@@ -7,7 +7,6 @@
 import sys
 from datetime import date
 
-import sphinx_gallery  # noqa: F401
```
sappelhoff, line 10: let's see in CI whether we really need this ...

PR #1293, opened 2024-08-16
fix doc and logging mark_channels

### LINKED ISSUE #1292: docstring for mark_channels is misleading
it says that the "default" for `status` is "bad", but in fact `status` is a required kwarg with no default.

sappelhoff:
PR Description
--------------

- closes #1292 

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -39,7 +39,7 @@ Detailed list of changes
 
 - :func:`mne_bids.read_raw_bids` no longer warns about unit changes in channels upon reading, as that information is taken from ``channels.tsv`` and judged authorative, by `Stefan Appelhoff`_ (:gh:`1282`)
 - MEG OPM channels are now experimentally included, by `Amaia Benitez`_ (:gh:`1222`)
-- :func:`mne_bids.mark_channels` will no longer created a ``status_description`` column filled with ``n/a`` in the ``channels.tsv`` file, by `Stefan Appelhoff`_ (:gh:`1293`)
+- :func:`mne_bids.mark_channels` will no longer create a ``status_description`` column filled with ``n/a`` in the ``channels.tsv`` file, by `Stefan Appelhoff`_ (:gh:`1293`)
 
 🛠 Requirements
 ^^^^^^^^^^^^^^^
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -39,6 +39,7 @@ Detailed list of changes
 
 - :func:`mne_bids.read_raw_bids` no longer warns about unit changes in channels upon reading, as that information is taken from ``channels.tsv`` and judged authorative, by `Stefan Appelhoff`_ (:gh:`1282`)
 - MEG OPM channels are now experimentally included, by `Amaia Benitez`_ (:gh:`1222`)
+- :func:`mne_bids.mark_channels` will no longer created a ``status_description`` column filled with ``n/a`` in the ``channels.tsv`` file, by `Stefan Appelhoff`_ (:gh:`1293`)
 
 🛠 Requirements
 ^^^^^^^^^^^^^^^
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -2551,7 +2551,7 @@ def mark_channels(bids_path, *, ch_names, status, descriptions=None, verbose=Non
             f"({len(ch_names)})."
         )
 
-    if set(status) != {"good", "bad"}:
+    if not set(status).issubset({"good", "bad"}):
         raise ValueError(
             'Setting the status of a channel must only be "good", or "bad".'
         )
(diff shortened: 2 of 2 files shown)


mne_bids/write.py
```diff
@@ -2565,7 +2564,7 @@ def mark_channels(bids_path, *, ch_names, status, descriptions=None, verbose=Non
         idx = tsv_data["name"].index(ch_name)
         logger.info(
             f"Processing channel {ch_name}:\n"
-            f"    status: bad\n"
+            f"    status: {status_}\n"
```
sappelhoff, line 2567: no idea why this was just `bad` before 🤔  especially because the column is initialized with all `good` if it wasn't present before.

```
    # Read sidecar and create required columns if they do not exist.
    if "status" not in tsv_data:
        logger.info('No "status" column found in input file. Creating.')
        tsv_data["status"] = ["good"] * len(tsv_data["name"])
```

I think this is also why in the docstring it said something about "default is 'bad'". If we start writing the status for a few channels only, the status for all other channels gets automatically set. However it gets set to "good", not "bad", ... which I find reasonable, but which was misleading in the function docstr.

sappelhoff: ok, I followed up on this in https://github.com/mne-tools/mne-bids/pull/1293/commits/5c042796f634e96c220682328de7968082c7b874

sappelhoff:
Thanks for making me aware of this @drammock, I think the present PR should fix a thing or two.

doc/whats_new.rst
```diff
@@ -39,6 +39,7 @@ Detailed list of changes
 
 - :func:`mne_bids.read_raw_bids` no longer warns about unit changes in channels upon reading, as that information is taken from ``channels.tsv`` and judged authorative, by `Stefan Appelhoff`_ (:gh:`1282`)
 - MEG OPM channels are now experimentally included, by `Amaia Benitez`_ (:gh:`1222`)
+- :func:`mne_bids.mark_channels` will no longer created a ``status_description`` column filled with ``n/a`` in the ``channels.tsv`` file, by `Stefan Appelhoff`_ (:gh:`1293`)
```
drammock, line 42:
```suggestion
- :func:`mne_bids.mark_channels` will no longer create a ``status_description`` column filled with ``n/a`` in the ``channels.tsv`` file, by `Stefan Appelhoff`_ (:gh:`1293`)
```

mne_bids/write.py
```diff
@@ -2543,18 +2551,24 @@ def mark_channels(bids_path, *, ch_names, status, descriptions=None, verbose=Non
             f"({len(ch_names)})."
         )
 
-    if not all(status in ["good", "bad"] for status in status):
+    if not all(status_ in ["good", "bad"] for status_ in status):
```
drammock, line 2554: [nitpick] as part of my crusade to increase usage of `set()` where appropriate, I would write this as:

```suggestion
    if set(status) != {"good", "bad"}:
```

feel free to disregard though! performance difference is negligible (41 vs 50 μs when `status` has 300 elements).

mne_bids/write.py
```diff
@@ -2543,18 +2551,24 @@ def mark_channels(bids_path, *, ch_names, status, descriptions=None, verbose=Non
...
         raise ValueError(
             'Setting the status of a channel must only be "good", or "bad".'
         )
 
     # Read sidecar and create required columns if they do not exist.
     if "status" not in tsv_data:
-        logger.info('No "status" column found in input file. Creating.')
+        logger.info(
+            'No "status" column found in input file.'
+            'Creating it with default value "good".'
```
drammock, line 2563: (here and below) it seems odd to call it "input file" since the user never passed the `*_channels.tsv` filename to the function (only the BIDSPath)

```suggestion
            'No "status" column found in channels file.'
            'Creating it with default value "good".'
```

drammock (commented):
I'm about to open a separate issue about this, but FWIW the behavior of "empty list means mark all channels" was *really* surprising and caused me to have to alter and re-run my dataset preparation script (no big deal, this is why we dry-run those scripts with only 2 subjs/2 sessions).

mne_bids/write.py
sappelhoff: I think this is great, thanks.

sappelhoff:
Thanks for the review, yes let's discuss your concerns in a separate issue.

mne_bids/write.py
```diff
@@ -2543,18 +2551,24 @@ def mark_channels(bids_path, *, ch_names, status, descriptions=None, verbose=Non
             f"({len(ch_names)})."
         )
 
-    if not all(status in ["good", "bad"] for status in status):
+    if set(status) != {"good", "bad"}:
```
sappelhoff, line 2554: I think this is what might be needed:

```suggestion
    if not set(status).issubset({"good", "bad"}):
```

sappelhoff:
test failures are unrelated, merging.

PR #1297, opened 2024-08-20
MAINT: Bump to 3.10 min req

larsoner:
Suggested move to require 3.10+. Clean up something that could have been cleaned up once 3.9+ was required. Clean up incorrect installation versions for stable and prev-stable (which used to be the same!).

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -113,8 +113,8 @@ jobs:
   test:
     # For GitHub "required" CI checks, add in branch protection:
     #
-    # 6 checks:
-    # for each OS (ubuntu, macos, macos-13, windows):
+    # 8 checks:
+    # for each machine type (ubuntu, macos, macos-13, windows):
     # (NOTE: macos-13=x86_64, macos>13=arm64)
     # 3.10 / mne-stable / full / validator-stable
     # 3.12 / mne-stable / full / validator-stable
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -114,11 +114,12 @@ jobs:
     # For GitHub "required" CI checks, add in branch protection:
     #
     # 6 checks:
-    # for each OS (ubuntu, macos, windows):
+    # for each OS (ubuntu, macos, macos-13, windows):
+    # (NOTE: macos-13=x86_64, macos>13=arm64)
     # 3.10 / mne-stable / full / validator-stable
     # 3.12 / mne-stable / full / validator-stable
     #
-    # 5 additional checks with alternative MNE-Python and BIDS validator versions:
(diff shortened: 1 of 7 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -115,38 +115,26 @@ jobs:
...
+    # ubuntu / 3.10 / mne-prev / full / validator-stable
     # ubuntu / 3.12 / mne-stable / minimal / validator-stable
     timeout-minutes: 60
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.9", "3.12"]  # Oldest and newest supported versions
+        os: [ubuntu-latest, macos-latest, macos-13, windows-latest]
```
larsoner, line 130: Based on the comments below I think this was also incorrectly specified once `macos-latest` switched from `-13` (x86_64) to `-14` (arm64). In any case might as well test both now I think

sappelhoff (approved):
LGTM! Thanks Eric.

.github/workflows/unit_tests.yml
```diff
@@ -115,38 +115,26 @@ jobs:
     #
     # 6 checks:
     # for each OS (ubuntu, macos, windows):
```
sappelhoff, line 117:
```suggestion
    # for each OS (ubuntu, macos, macos-13, windows):
    # (NOTE: macos-13=x86_64, macos>13=arm64)
```

.github/workflows/unit_tests.yml
```diff
@@ -115,38 +115,26 @@ jobs:
...
-    # 3.9 / mne-stable / full / validator-stable
-    # 3.12 / mne-stable / full / validator-stable
-    #
-    # 1 additional check for Apple Silicon (doesn't support Python 3.9):
+    # 3.10 / mne-stable / full / validator-stable
     # 3.12 / mne-stable / full / validator-stable
     #
     # 5 additional checks with alternative MNE-Python and BIDS validator versions:
```
sappelhoff, line 121:
```suggestion
    # 3 additional checks with alternative MNE-Python and BIDS validator versions:
```

sappelhoff (commented):
I'll adjust the branch protection rules for required checks.

.github/workflows/unit_tests.yml
sappelhoff: actually, the `    # 6 checks:` above needs to be changed to "8 checks" 🤔 

and it'd be nice to have a changelog entry saying that we now require Python 3.10 minimum. Could you add that please @larsoner ? If not, let me know and I'll do it at some later point. Thanks again.

larsoner:
Okay should be good @sappelhoff !

PR #1307, opened 2024-09-19
deprecate empty list as indicator for marking all channels

### LINKED ISSUE #1294: mark_channels behavior unexpected/unintuitive
Yesterday I did something like `mark_channels(bids_path=my_bp, channels=[], status="bad")`. I was expecting it to be a no-op (because `channels` is an empty list) but in fact it marked *all* channels as bad. Now, the docstring is clear on this point, so I can't really complain about that. But I would argue that it is *really unintuitive* that passing a list of 1 channel name marks 1 channel, but a list of 0 channel names marks *all* channels (instead of zero channels).

My use case: we have `prebad.txt` files listing which channels were noted as noisy during the recording. Sometimes those files are empty.  It's useful to have that file exist and be empty, rather than just not existing at all, because then you know that someone looked for bad channels and found none (if "no bad channels" was indicated by absence of the file, you wouldn't know whether someone checked and found no bad channels, or just forgot to check at all).

In that context, my script finds the `prebad` file, reads it, and if there are no channel names in that file, the `prebads` variable (which I pass to `mark_channels()`) is an empty list --- which does not do what I want it to!

My suggestion is that if you want to have a signal value that will mark all channels as good/bad, that signal value should be something like `"all"` rather than `[]`. If you agree, feel free to assign this issue to me and I can implement the change.

drammock:
PR Description
--------------

paves the way for future change in behavior of `mark_channels(..., ch_names=[])` from marking all channels to marking no channels.

Closes #1294 

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -2490,7 +2490,7 @@ def test_mark_channels(
     # Mark `existing_ch_names` as bad in raw and sidecar TSV before we
     # begin our actual tests, which should then add additional channels
     # to the list of bads, retaining the ones we're specifying here.
-    mark_channels(ch_names=[], bids_path=bids_path, status="good", verbose=False)
+    mark_channels(ch_names="all", bids_path=bids_path, status="good", verbose=False)
     _bids_validate(bids_root)
     raw = read_raw_bids(bids_path=bids_path, verbose=False)
     # Order is not preserved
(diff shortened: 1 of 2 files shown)


agramfort (approved):

sappelhoff (approved):
clean and great, thanks @drammock

sappelhoff:
Most of the test failures are flukes, see:

- https://github.com/bids-standard/bids-validator/issues/2124

However, these ones are real:

- https://github.com/mne-tools/mne-bids/actions/runs/10948765960/job/30400543539?pr=1307#step:20:612

could you address them before we merge this, please?

hoechenberger (approved):

drammock:
ok @sappelhoff, looks like all CIs are happy now

hoechenberger:
Great stuff, thanks @drammock

sappelhoff:
thanks!

PR #1314, opened 2024-10-04
Add intersphinx_registry and toml-sort

### LINKED ISSUE #1300: Use intersphinx_registry
in our `conf.py`, we should use https://github.com/Quansight-Labs/intersphinx_registry and we should add mne-bids to that registry.

mscheltienne:
Add `toml-sort` and `intersphinx_registry` to the tooling chain.
Closes #1300

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,9 +61,9 @@ doc = [
 ]
 # Dependencies for using all mne_bids features
 full = [
-  "defusedxml", # For reading EGI MFF data and BrainVision montages
+  "defusedxml",  # For reading EGI MFF data and BrainVision montages
   "edfio >= 0.2.1",
-  "EDFlib-Python >= 1.0.6", # XXX: drop once mne <1.7 is no longer supported
+  "EDFlib-Python >= 1.0.6",  # XXX: drop once mne <1.7 is no longer supported
   "eeglabio >= 0.0.2",
   "matplotlib >= 3.5.0",
   "nibabel >= 3.2.1",
@@ -102,7 +102,7 @@ exclude = [
 ]
 
 [tool.hatch.metadata]
-allow-direct-references = true # allow specifying URLs in our dependencies
+allow-direct-references = true  # allow specifying URLs in our dependencies
 
 [tool.hatch.version]
 raw-options = {version_scheme = "release-branch-semver"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,77 +6,71 @@ build-backend = "hatchling.build"
 name = "mne-bids"
 description = "MNE-BIDS: Organizing MEG, EEG, and iEEG data according to the BIDS specification and facilitating their analysis with MNE-Python"
 dynamic = ["version"]
-authors = [{ name = "The MNE-BIDS developers" }]
+authors = [{name = "The MNE-BIDS developers"}]
 maintainers = [
-    { name = "Stefan Appelhoff", email = "stefan.appelhoff@mailbox.org" },
+  {name = "Stefan Appelhoff", email = "stefan.appelhoff@mailbox.org"}
 ]
-license = { text = "BSD-3-Clause" }
-readme = { file = "README.md", content-type = "text/markdown" }
+license = {text = "BSD-3-Clause"}
+readme = {file = "README.md", content-type = "text/markdown"}
 requires-python = ">=3.10"
 keywords = [
-    "meg",
-    "eeg",
-    "ieeg",
-    "bids",
-    "brain imaging data structure",
-    "neuroscience",
-    "neuroimaging",
+  "meg",
+  "eeg",
+  "ieeg",
+  "bids",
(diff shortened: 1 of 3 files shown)


larsoner (approved):

drammock (commented):
I suggest adding the following to `pyproject.toml`:

```toml
[tool.tomlsort]
sort_table_keys = true
sort_inline_tables = true
sort_inline_arrays = true
spaces_before_inline_comment = 2
spaces_indent_inline_array = 4
trailing_comma_inline_array = true
```
the `sort_*` settings will make order fully deterministic, and the other settings will make the result more similar to how `ruff` formats python code, which IMO makes manual edits slightly easier (e.g., hitting TAB in the IDE will yield the right number of spaces by default)

sappelhoff (approved):

mscheltienne:
I was certain I putted:

```
[tool.tomlsort]
all = true
ignore_case = true
trailing_comma_inline_array = true
```

But now I can't find it..
This config is also deterministic and IIRC `all` replaces your 3 `sort` lines. For the spacing, I would argue it's a configuration of the IDE. If I hit TAB in a `pyproject.toml`, I get a spacing of 2, not 4.

pyproject.toml
```diff
@@ -87,24 +82,64 @@ dev = ["mne_bids[test,doc,full]", "pre-commit"]
...
+  # numba with NumPy dev
+  "ignore:`np.MachAr` is deprecated.*:DeprecationWarning",
+  # old MNE _fake_click
+  "ignore:The .*_event function was deprecated in Matplotlib.*:",
+  "ignore:datetime\\.datetime\\.utcfromtimestamp.* is deprecated and scheduled for removal in a future version.*:DeprecationWarning",
+  # matplotlib
+  "ignore:Figure.*is non-interactive.*cannot be shown:UserWarning",
+  # NumPy 2.1 bug (probably)
+  "ignore:__array__ implementation doesn.*:DeprecationWarning"
+]
```
mscheltienne, line 142:
```suggestion
  "ignore:__array__ implementation doesn.*:DeprecationWarning"
]

[tool.tomlsort]
all = true
ignore_case = true
trailing_comma_inline_array = true

```

drammock:
> This config is also deterministic and IIRC `all` replaces your 3 `sort` lines. For the spacing, I would argue it's a configuration of the IDE. If I hit TAB in VSCode in a `pyproject.toml`, I get a spacing of 2, not 4.

yep, confirmed that `all` is equiv to the 3 sort lines I suggested. (I was skimming docs quickly and assumed it meant "format all discovered files" or something).

Your point about spacing is true --- users *can* configure their IDEs to do tabs differently for different file formats.  But if we use the setting I suggested, they won't *need* to do that... though I guess it's possible that my suggested setting will lead folks to need to undo their 2-space tab settings?

For other settings though (e.g., 2 spaces before inline comments) it's less about IDEs and more about muscle memory: chances are if I'm typing an inline comment I'll use 2 spaces, and when reading code I'm used to things being looking a certain way (I'm guessing that's why you added trailing comma --- after months/years of using black/ruff, it's starting to look "wrong" if it's not there?)

If you're not convinced that's OK. I don't insist on anything other than the sorting, which you've already done.

mscheltienne:
> For other settings though (e.g., 2 spaces before inline comments) it's less about IDEs and more about muscle memory: chances are if I'm typing an inline comment I'll use 2 spaces, and when reading code I'm used to things being looking a certain way (I'm guessing that's why you added trailing comma --- after months/years of using black/ruff, it's starting to look "wrong" if it's not there?)

I completely agree and wil ladd the 2 spaces before inline comments at least ;)

> Your point about spacing is true --- users can configure their IDEs to do tabs differently for different file formats. But if we use the setting I suggested, they won't need to do that... though I guess it's possible that my suggested setting will lead folks to need to undo their 2-space tab settings?

My issue here, is that TBH, I don't think I actually configured anything here and yet that's the default I got (it might be from the `Even Better TOML` externsion that I have installed). So in this case, we have 2 different 'default' behavior, which is not great. I'm enclined to keep the default from `toml-sort` for this one.

pyproject.toml
```diff
@@ -1,155 +1,155 @@
...
-    "ignore:The .*_event function was deprecated in Matplotlib.*:",
-    "ignore:datetime\\.datetime\\.utcfromtimestamp.* is deprecated and scheduled for removal in a future version.*:DeprecationWarning",
-    # matplotlib
-    "ignore:Figure.*is non-interactive.*cannot be shown:UserWarning",
-    # NumPy 2.1 bug (probably)
-    "ignore:__array__ implementation doesn.*:DeprecationWarning",
-]
+[tool.tomlsort]
+all = true
+ignore_case = true
```
mscheltienne, line 154:
```suggestion
ignore_case = true
spaces_before_inline_comment = 2
```

drammock (approved):

PR #1318, opened 2024-10-10
Fix handling of README files with extensions

thht:
PR Description
--------------

According to [the BIDS specs](https://bids-specification.readthedocs.io/en/stable/glossary.html#readme-files), a README file can have no extension or one of the following: `.txt`, `.md`, `.rst.`

The current behavior of MNE-BIDS is to always assume an extensionless README to be present. A README file with one of the valid extensions is simply ignored, a new README file is written. This results in two README files being present in the BIDS root folder which violates the specs.

This PR fixes this by looking whether one of the valid extensions are present and uses the correct file.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1959,7 +1959,7 @@ def write_raw_bids(
     allowed_readme_extensions = ("", ".md", ".txt", ".rst")
     for cur_extension in allowed_readme_extensions:
         if Path(bids_path.root, f"README{cur_extension}").exists():
-            readme_fname = str(Path(bids_path.root, f"README{cur_extension}"))
+            readme_fname = Path(bids_path.root, f"README{cur_extension}")
             break
     participants_tsv_fname = op.join(bids_path.root, "participants.tsv")
     participants_json_fname = participants_tsv_fname.replace(".tsv", ".json")
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1955,12 +1955,18 @@ def write_raw_bids(
     )
 
     # For the remaining files, we can use BIDSPath to alter.
-    readme_fname = op.join(bids_path.root, "README")
-    allowed_readme_extensions = ("", ".md", ".txt", ".rst")
-    for cur_extension in allowed_readme_extensions:
-        if Path(bids_path.root, f"README{cur_extension}").exists():
-            readme_fname = Path(bids_path.root, f"README{cur_extension}")
-            break
+    readme_suffixes = ("", ".md", ".rst", ".txt")
+    found_readmes = sorted(
+        filter(lambda x: x.suffix in readme_suffixes, bids_path.root.glob("README*"))
+    )
+    if len(found_readmes) > 1:
+        raise RuntimeError(
+            "Multiple README files found in the BIDS root folder. "
+            "This violates the BIDS specifications. "
+            "Please ensure there is only one README file."
+        )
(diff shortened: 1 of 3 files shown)


sappelhoff:
Thanks! Please ping me once this is ready for review/merge.

thht:
pinging @sappelhoff. all checks green. looking forward to your thoughts and input!

mne_bids/write.py
```diff
@@ -1956,6 +1956,11 @@ def write_raw_bids(
 
     # For the remaining files, we can use BIDSPath to alter.
     readme_fname = op.join(bids_path.root, "README")
+    allowed_readme_extensions = ("", ".md", ".txt", ".rst")
+    for cur_extension in allowed_readme_extensions:
+        if Path(bids_path.root, f"README{cur_extension}").exists():
+            readme_fname = str(Path(bids_path.root, f"README{cur_extension}"))
+            break
```
drammock, line 1963:
```suggestion
    readme_suffixes = ("", ".md", ".rst", ".txt")
    found_readmes = sorted(
        filter(lambda x: x.suffix in readme_suffixes, bids_path.root.glob("README*"))
    )
    readme_fname = str((found_readmes or [bids_path.root / "README"])[0])
```

This suggestion seems a bit cleaner to me, but may be just stylistic preference; feel free to disregard @sappelhoff if you prefer @thht's approach. Not sure if the enclosing cast to `str` is necessary but that's what `op.join` would have returned. Behavior is slightly different if (for some reason) multiple READMEs are present; PR as written prefers `README` > `README.md` > `README.txt` > `README.rst`, whereas my suggestion prefers `.rst` over `.txt` if both are present. I doubt this matters; IIUC multiple READMEs present is considered a violation of the standard and should be very rare --- nonetheless do we want to detect this and error if it happens?

NB: Py3.12 adds a `case_sensitive=False` option to `Path.glob` that might be worth adding once 3.12 is minimum version.

thht: The cast to `str` is actually not necessary. I removed it.

Regarding possible multiple READMEs: I'm fine with both solutions. And I'd also leave it up to @sappelhoff if we should warn or throw if multiple READMEs are found....

hoechenberger (approved):

sappelhoff (commented):
LGTM in general, thanks for the contribution.

However I'd also like the behavior to be that when multiple READMEs are discovered, we throw an error, instead of what this PR currently implements (selecting one README, and potentially ignoring the presence of the others)

thht:
> LGTM in general, thanks for the contribution.
> 
> However I'd also like the behavior to be that when multiple READMEs are discovered, we throw an error, instead of what this PR currently implements (selecting one README, and potentially ignoring the presence of the others)

Working on it right now ;-)

mne_bids/write.py
thht: So, I adapted the approach by @drammock and added the check and exception raising when multiple readmes are found. I also added a test for this.

thht:
As you can see in [my other reply](https://github.com/mne-tools/mne-bids/pull/1318#discussion_r1796813118), I added raising a `RuntimeError` when multiple READMEs are encounted including a test for this.

sappelhoff (approved):

sappelhoff:
Thanks @thht!

PR #1322, opened 2024-10-16
remove bids-validator pin

sappelhoff:
the reason for the pin has been fixed:

- https://github.com/bids-standard/bids-validator/issues/2124


sappelhoff:
I have no idea why this fails with:

> python: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /opt/hostedtoolcache/Python/3.12.7/x64/lib/libpython3.12.so.1.0)

There are other jobs on the same sort of specification that run without problems.

PR #1339, opened 2024-11-21
os.path to pathlib: part 1 / n

### LINKED ISSUE #1268: modernize path handling
we still use `op.join` in many cases that could be simplified with the `pathlib` syntax of concatenation via `/`.

sappelhoff:
- addresses a part for #1268

### DIFF
--- a/examples/convert_empty_room.py
+++ b/examples/convert_empty_room.py
@@ -42,7 +42,7 @@ from mne_bids import BIDSPath, print_dir_tree, read_raw_bids, write_raw_bids
 data_path = sample.data_path()
 raw_fname = data_path / "MEG" / "sample" / "sample_audvis_raw.fif"
 
-bids_root = data_path / ".." / "MNE-sample-data-bids"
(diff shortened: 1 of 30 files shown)


examples/convert_empty_room.py
```diff
@@ -41,9 +40,9 @@
 # And define the paths and event_id dictionary.
 
 data_path = sample.data_path()
-raw_fname = op.join(data_path, "MEG", "sample", "sample_audvis_raw.fif")
+raw_fname = data_path / "MEG" / "sample" / "sample_audvis_raw.fif"
 
-bids_root = op.join(data_path, "..", "MNE-sample-data-bids")
+bids_root = data_path / ".." / "MNE-sample-data-bids"
```
hoechenberger, line 45: maybe

```suggestion
bids_root = data_path.parent / "MNE-sample-data-bids"
```

hoechenberger (approved):

sappelhoff:
FYI, I will keep pushing to this for a bit longer, there is still a lot to adjust.

PR #1349, opened 2024-12-05
optionally return event_id dict from read_raw_bids, and use value column from events file if present

drammock:
PR Description
--------------

closes #223 

I opted for @hoechenberger's suggestion in https://github.com/mne-tools/mne-bids/issues/223#issuecomment-964165378, namely, to allow `read_raw_bids` to optionally return an `event_id` dict; when a `values` column is present in `events.tsv` it will use those values for the integer IDs. That dict can then be passed to `mne.events_from_annotations` to recover the original events array.

Merge checklist
---------------

Maintainer, please confirm the following before merging.
If applicable:

- [ ] All comments are resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)
- [x] PR description includes phrase "closes <#issue-number>"

PR #1353, opened 2024-12-16
fix eager dtype conversion of value column

### LINKED ISSUE #1358: Documentation Build Errors.
As seen in #1357 and #1355

the `docs-build` job is giving this error:

```
/home/circleci/project/mne_bids/path.py:docstring of mne_bids.BIDSPath.root:1: WARNING: py:class reference target not found: pathlib._local.Path [ref.class]
```

From local tests, my current hypothesis is that this has something to do with Python 3.13, which the CircleCI runner uses:

Because Locally, on python **3.10**, I don't get the error above, and the docstring of `BIDSPath.root` correctly hyperlinks to the python docs:

Screenshot 2024-12-20 at 11 05 49 AM

on Python **3.13** it looks like this ("Path" does not hyperlink):

Screenshot 2024-12-20 at 11 10 54 AM

I don't think I have the skill to track down the source of that issue. But we can force the CircleCI runner that deploys `docs-build` to user a lower version of Python as a work around?

drammock:
Follow-up to #1349, see https://github.com/mne-tools/mne-bids/pull/1349/files#r1885104885

- fixes #1358 

PR Description
--------------

Basically: use defensive programming when trying to dtype-convert the `value` column from `events.tsv`.  I was a bit worried that a column called `value` could easily exist in real datasets and not conform to the way that column gets written by MNE-BIDS, and it didn't take long for @larsoner to find one :)

The bug was already caught by a downstream test in MNE-BIDS-Pipeline, so I didn't bother adding a new test here, but LMK if you want me to add a simple one.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -531,7 +531,8 @@ def _handle_events_reading(events_fname, raw):
     logger.info(f"Reading events from {events_fname}.")
     events_dict = _from_tsv(events_fname)
 
-    # drop events where onset is n/a
+    # drop events where onset is n/a; we can't annotate them and thus don't need entries
+    # for them in event_id either
     events_dict = _drop(events_dict, "n/a", "onset")
 
     # Get event descriptions. Use `trial_type` column if available.
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -548,9 +548,11 @@ def _handle_events_reading(events_fname, raw):
     # If we lack proper event descriptions, perhaps we have at least an event value?
     elif "value" in events_dict:
         trial_type_col_name = "value"
-    # Worst case: all events will become `n/a` and all values will be `1`
+    # Worst case: all events become `n/a` and all values become `1`
     else:
         trial_type_col_name = None
+        descrs = np.full(len(events_dict["onset"]), "n/a")
+        event_id = {descrs[0]: 1}
 
     if trial_type_col_name is not None:
         # Drop events unrelated to a trial type
@@ -596,10 +598,6 @@ def _handle_events_reading(events_fname, raw):
             event_id = dict(zip(trial_types, np.arange(len(trial_types))))
         descrs = np.asarray(trial_types, dtype=str)
 
-    # Worst case: all events become `n/a` and all values become `1`
-    else:
-        descrs = np.full(len(events_dict["onset"]), "n/a")
-        event_id = {descrs[0]: 1}
     # Deal with "n/a" strings before converting to float
     ons = np.asarray(events_dict["onset"], dtype=float)
     durs = np.array(
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -586,19 +586,19 @@ def _handle_events_reading(events_fname, raw):
             culled_vals = culled["value"]
             try:
                 culled_vals = np.asarray(culled_vals, dtype=float)
-            except ValueError:
+            except ValueError:  # contained strings or complex numbers
                 pass
             else:
                 try:
                     culled_vals = culled_vals.astype(int)
-                except ValueError:
+                except ValueError:  # numeric, but has some non-integer values
                     pass
             event_id = dict(zip(culled[trial_type_col_name], culled_vals))
         else:
             event_id = dict(zip(trial_types, np.arange(len(trial_types))))
         descrs = np.asarray(trial_types, dtype=str)
 
-    # Deal with "n/a" strings before converting to float
+    # convert onsets & durations to floats ("n/a" onsets were already dropped)
     ons = np.asarray(events_dict["onset"], dtype=float)
     durs = np.array(
         [0 if du == "n/a" else du for du in events_dict["duration"]], dtype=float
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -574,12 +574,23 @@ def _handle_events_reading(events_fname, raw):
                         new_name = f"{trial_type}/{value}"
                         logger.info(f"    Renaming event: {trial_type} -> {new_name}")
                         trial_types[ii] = new_name
-            # drop rows where `value` is `n/a` & convert remaining `value` to int (only
-            # when making our `event_id` dict; `value = n/a` doesn't prevent annotation)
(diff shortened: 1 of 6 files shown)


drammock:
circleCI failure is a bit mysterious (failure to cross-ref `pathlib.Path` in an attribute docstring).  I thought it was https://github.com/sphinx-doc/sphinx/pull/12317 but CircleCI here is running Sphinx 8.1.3 so that fix should definitely be in place. I'll try to dig deeper tomorrow.  Also haven't looked into the Py3.10 build failure yet, beyond noting that it's a metadata-missing on `twine --check` error.  anyone else should feel free to investigate if you have a debugging craving and a few minutes.

agramfort (approved):

drammock:
OK, the cross-ref failure looks like https://github.com/sphinx-doc/sphinx/issues/13178; `autodoc` is the cause but so far I can't really understand the root of the problem. Probably we just need to add a warning ignore :(

sadly I can't replicate the `twine check --strict dist/*` failure locally using a py3.10 environment :(

drammock:
ok was able to reproduce the twine failure, and found a workaround here: https://github.com/astral-sh/rye/issues/1446#issuecomment-2545154562

TL;DR some kind of mismatch between twine and hatchling, related to metadata versions. 
Installing `twine>=6` didn't fix it though, whereas pinning `hatchling==1.26.3` did (latest is `1.27.0`).  TBH IDK what the problem is, both `name = "mne-bids"` and `dynamic = ["version"]` are in `pyproject.toml` 🤷🏻

hoechenberger:
Thanks for looking into this, @drammock

scott-huberty:
Hi @drammock 

To understand the original issue I actually had to create a test myself that would be caught by the  try-catches added in this PR, so maybe it is indeed worth adding a test? 

Something like this to `mne_bids/tests/test_read/test_handle_events_reading`:


```python
# Test with only a `value` column, but where the values in this `value` column don't
# meet our assumption that they can be coerced to integers/floats.
events = {
        "onset": [10, 15],
        "duration": [1, 1],
        "value": ["Condition 1", "Condition 2"],
}
events_fname = tmp_path / "bids6" / "sub-01_task-test_events.tsv"
events_fname.parent.mkdir()
_to_tsv(events, events_fname)

raw, event_id = _handle_events_reading(events_fname, raw)
# In this edge case, the keys and values of `event_id` are the same, i.e. they are
# both the original values from the `value` column. These events also exist as
# annotations in the raw data.
descs = raw.annotations.description.tolist() # ['Condition 1', 'Condition 2']
event_keys = list(event_id.keys())
event_vals = list(event_id.values())
assert descs == event_keys == event_vals == ["Condition 1", "Condition 2"]
```

This test creates an events TSV file like:

```
onset	duration	value
10	1	Condition 1
15	1	Condition 2
```


Where `_, event_id = _handle_events_reading(events_fname, raw)` returns

```
{'Condition 1': 'Condition 1', 'Condition 2': 'Condition 2'}
```


And AFAICT, from the User's perspective:

1.  This `event_id` would only get returned to the user if they do `read_raw_bids(..., return_event_dict=True)`
2. The `condition 1` and `condition 2` events are assigned to annotations in the raw object.

scott-huberty (approved):

scott-huberty:
On second thought... 

`{'Condition 1': 'Condition 1', 'Condition 2': 'Condition 2'}` is strange. 

In this case, maybe we should either

- return `None` (and warn the user)
- or use `np.arange` to create a "proper" event dict.

mne_bids/read.py
```diff
@@ -574,12 +575,23 @@ def _handle_events_reading(events_fname, raw):
...
-                zip(culled[trial_type_col_name], np.asarray(culled["value"], dtype=int))
-            )
+            # Often (but not always!) the `value` column was written by MNE-BIDS and
+            # represents integer event IDs (as would be found in MNE-Python events
+            # arrays / event_id dicts). But in case not, let's be defensive:
+            culled_vals = culled["value"]
+            try:
+                culled_vals = np.asarray(culled_vals, dtype=float)
+            except ValueError:
+                pass
```
scott-huberty, line 588:
```suggestion
               #  If the values are not numeric, just use the row index as the event ID
                culled_vals = np.arange(len(culled_vals))
```

sappelhoff:
@scott-huberty could you please push your test to this PR? For now without changing the behavior as in your suggestions. I would rather merge this one sooner than later as it contains two important fixes / workarounds.

We can then discuss your suggested changes in a new issue if you like.

scott-huberty:
> @scott-huberty could you please push your test to this PR? For now without changing the behavior as in your suggestions. I would rather merge this one sooner than later as it contains two important fixes / workarounds.
> 
> We can then discuss your suggested changes in a new issue if you like.

sorry @sappelhoff Since I am not an MNE maintainer I don't think that I have the push rights needed to do that..(?)

drammock:
> I don't think that I have the push rights needed to do that

You can open a PR into my fork+branch.

drammock:
> > I don't think that I have the push rights needed to do that
> 
> You can open a PR into my fork+branch.

never mind @scott-huberty, I pushed the (simplified) test.  I gave some thought to your suggestion about setting the event_id to use row numbers instead of the (string) values in the values column. On one hand I think it makes sense (so that it would be possible to actually use that event_id dict in mne-python functions) but I also decided it's not crucial to get it in to this PR, because in the case where the events.tsv value column contains strings, the user won't be *expecting* to be able to recover an original mapping of eventID integers (because their value column wasn't written by MNE-BIDS in the first place).

The other reason I didn't try to add it in here, is that I think it's not ideal to just assign eventID integers from row numbers --- I think you'd want to take (sorted?) unique values and then assign sequential integers starting from 1 probably.  Otherwise the event_id mapping changes depending on things like randomized trial presentation order, meaning different subjects could get different mappings.  Or, to be really safe, somehow create integers from the hashes of the keys; that would be safest (though perhaps too annoying to be worth it --- especially given that I don't expect users who are in this particular corner case to even *want* the event_id back).

sappelhoff:
Thanks!

PR #1357, opened 2024-12-19
FIX: Support UTF-8 encoding for JSON files

### LINKED ISSUE #1356: UTF-8 Encoding is not respected by make_dataset_description

Given the wide use of `encoding=utf-8-sig` in mne-bids, I assume that the example below is a bug, and `mne-bids` does want to support characters like ł  ([latin small letter with stroke](https://www.compart.com/en/unicode/U+0142)) that appear in some people's names

### Steps to reproduce

```Python
import tempfile
from pathlib import Path

import mne_bids

text = "Michał is a name that includes the character ł."
with tempfile.TemporaryDirectory() as temp_dir:
    mne_bids.make_dataset_description(
        path=temp_dir,
        name=text,
        authors=["Michał"],
    )
    with Path(f"{temp_dir}/dataset_description.json").open(encoding="utf-8") as fid:
        dataset_description = fid.read()
    # this works
    with Path("./test.txt").open("w", encoding="utf-8") as temp_file:
        temp_file.write(text)
    with Path("./test.txt").open("r", encoding="utf-8") as temp_file:
        expected_text = temp_file.read()

print("WHAT MNE BIDS WROTE TO DATASET_DESCRIPTION.JSON:")
print(dataset_description)
print(" ")
print(f"WHAT I WOULD EXPECT: {expected_text}")
```

### Expected results

`WHAT I WOULD EXPECT: Michał is a name that includes the character ł.`

### Actual results

```
{
    "Name": "Micha\u0142 is a name that includes the character \u0142.",
    "BIDSVersion": "1.7.0",
    "DatasetType": "raw",
    "Authors": [
        "Micha\u0142"
    ]
}
```
### Additional information

I took a quick look at the code but haven't figured out yet what is going wrong ☹️

scott-huberty:
fixes #1356 

`_write_json` is used _a lot_ in the MNE-BIDS codebase... I want to get your feedback on whether adding a new parameter `ensure_ascii=False` to `_write_tsv` is the right way to go?

Because I'm not sure if there are cases in MNE-BIDS where we do want to enforce ASCII encoding during JSON writing.. If so, we could change the default to `ensure_ascii=True`, and pass `False` explicitly for the cases where we want to support UTF-8?

cc @sappelhoff @hoechenberger  @larsoner

### TODO

- [x] add test

### DIFF
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -387,10 +387,9 @@ def test_make_dataset_description(tmp_path, monkeypatch):
     with open(op.join(tmp_path, "dataset_description.json"), encoding="utf-8") as fid:
         dataset_description_json = json.load(fid)
         assert dataset_description_json["Authors"] == ["MNE B.", "MNE P.", "MNE Ł."]
-
-    # If the text on disk is unicode, json.load will convert it. So let's test that the
-    # text was encoded correctly on disk.
-    with open(op.join(tmp_path, "dataset_description.json"), encoding="utf-8") as fid:
+        # If the text on disk is unicode, json.load will convert it. So let's test that
+        # the text was encoded correctly on disk.
+        fid.seek(0)
         # don't use json.load here, as it will convert unicode to str
         dataset_description_string = fid.read()
         # Check that U+0141 was correctly encoded as Ł on disk
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -376,7 +376,7 @@ def test_make_dataset_description(tmp_path, monkeypatch):
     make_dataset_description(
         path=tmp_path,
         name="tst2",
-        authors="MNE B., MNE P.",
+        authors="MNE B., MNE P., MNE Ł.",
         funding="GSOC2019, GSOC2021",
         references_and_links="https://doi.org/10.21105/joss.01896",
         dataset_type="derivative",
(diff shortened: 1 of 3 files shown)


hoechenberger:
I don't think we want to convert to ASCII for any JSON we're writing to disk.

hoechenberger (approved):

mne_bids/tests/test_write.py
```diff
@@ -386,7 +386,15 @@ def test_make_dataset_description(tmp_path, monkeypatch):
 
     with open(op.join(tmp_path, "dataset_description.json"), encoding="utf-8") as fid:
         dataset_description_json = json.load(fid)
-        assert dataset_description_json["Authors"] == ["MNE B.", "MNE P."]
+        assert dataset_description_json["Authors"] == ["MNE B.", "MNE P.", "MNE Ł."]
+
+    # If the text on disk is unicode, json.load will convert it. So let's test that the
+    # text was encoded correctly on disk.
+    with open(op.join(tmp_path, "dataset_description.json"), encoding="utf-8") as fid:
```
drammock, line 393: Instead of closing and re-opening the file, you can rewind the "playhead" to the start of the open file, then use `fid.read()` as usual

```suggestion
        # If the text on disk is unicode, json.load will convert it. So let's test that
        # the text was encoded correctly on disk.
        fid.seek(0)
```

drammock (approved):
LGTM, nice adaptation of the existing test. One suggested improvement, see below

sappelhoff:
Thanks @scott-huberty

PR #1362, opened 2025-01-02
allow get_entity_vals to ignore sidecars

### LINKED ISSUE #1351: get_entity_vals should ignore sidecars
### Describe the problem

My use case for `get_entity_vals` is this: for a given subject + task, find which sessions exist where the subject performed that task. This is proving to be quite hard (impossible?) to do with `get_entity_vals`, because `get_entity_vals` doesn't seem to discriminate when it comes to data files vs sidecars. MWE:

```py
from pathlib import Path
from tempfile import TemporaryDirectory

import mne_bids as mb

# create temporary fake BIDS dataset
with TemporaryDirectory() as root:
    root = Path(root)
    subdir = root / "sub-1" / "ses-a" / "meg"
    subdir.mkdir(parents=True)
    stem = "sub-1_ses-a_{}"
    for fname in ("task-IDS_meg.fif", "coordsystem.json"):
        (subdir / stem.format(fname)).touch()
    (subdir.parent / stem.format("scans.tsv")).touch()

    mb.print_dir_tree(root)

    # which of sub-1's session(s) have data for tasks other than IDS?
    result = mb.get_entity_vals(
        root / "sub-1",
        entity_key="session",
        ignore_tasks=("IDS",),
    )
print(result)
```

Result is this:
```
|tmp39vnjgza/
|--- sub-1/
|------ ses-a/
|--------- sub-1_ses-a_scans.tsv   # <-- false positive?
|--------- meg/
|------------ sub-1_ses-a_coordsystem.json   # <-- false positive?
|------------ sub-1_ses-a_task-IDS_meg.fif   # this is ignored by `ignore_tasks`
['a']
```

Expected (or at least desired) result here is that `get_entity_vals` returns `[]` because there are no *data files* once `ignore_tasks` has taken effect. I don't care if there are `coordsystem` or `scans` files present, because they relate to task(s) that I'm ignoring.

### Describe your solution

`get_entity_vals` should (have a param that lets you easily) ignore non-data sidecars.  I'm not super clear on what the ramifications of that would be, exactly which sidecars should be ignored, or how widely used `get_entity_vals` is internally or in user code. But at least the use cases I've thought about, I can't see why non-data files would be of interest --- in theory they'll only be present if data files are *also* present, and the spirit of the params like `ignore_tasks`, `ignore_sessions` etc seems to suggest that (one of) the intended uses of this function is to check whether a given task is present for a given subject, or check which subjects have task IDS in session a, or similar such queries.

Maybe there's a better way to do what I want than `get_entity_vals`, that I'm not thinking of?

### Additional context

_No response_

drammock:
closes #1351 

PR Description
--------------

adds a new parameter `ignore_suffixes` to `get_entity_vals` so that sidecars (scans, coordsystem) may easily be ignored.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [x] All CIs are happy
- [x] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -32,6 +32,7 @@ Detailed list of changes
 ^^^^^^^^^^^^^^^
 
 - :func:`mne_bids.write_raw_bids()` can now handle mne `Raw` objects with `eyegaze` and `pupil` channels, by `Christian O'Reilly`_ (:gh:`1344`)
+- :func:`mne_bids.get_entity_vals()` has a new parameter ``ignore_suffixes`` to easily ignore sidecar files, by `Daniel McCloy`_ (:gh:`1362`)
 
 
 🧐 API and behavior changes
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import glob
+import inspect
 import json
 import os
 import re
@@ -1996,6 +1997,7 @@ def get_entity_vals(
     .. [1] https://bids-specification.rtfd.io/en/latest/common-principles.html#entities
 
     """
+    params = inspect.signature(get_entity_vals).parameters  # for debug messages
     root = _check_fname(
         fname=root,
         overwrite="read",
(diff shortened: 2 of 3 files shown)


drammock:
All green, ready for review/merge

sappelhoff (approved):
Thanks a lot!

PR #1355, opened 2024-12-18
WIP Rework mne bids path match

waldie11:
PR Description
--------------

When `_return_root_paths` now uses `ignore_nosub`, it can benefit from from only exploring directories within `root/sub-*` instead of exploring the whole tree and subsequently applying the regex filter on visited files.

 `get_entity_vals` is extended by an argument `include_match` to introduce a prefilter on regex path match. This is a whitelist approach, and extends both functionality and performance compared to a equivalent blacklist approach of existing option `ignore_dirs`.

Merge checklist
---------------

Maintainer, please confirm the following before merging.
If applicable:

- [ ] All comments are resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)
- [ ] PR description includes phrase "closes <#issue-number>"

PR #1376, opened 2025-02-24
Revert "WIP Rework mne bids path match"

drammock:
Reverts mne-tools/mne-bids#1355

This changed behavior in a way that broke downstream tests, see https://github.com/mne-tools/mne-bids-pipeline/issues/1059

cc @waldie11 @sappelhoff

### DIFF
--- a/doc/authors.rst
+++ b/doc/authors.rst
@@ -5,12 +5,10 @@
 .. _Amaia Benitez: https://github.com/AmaiaBA
 .. _Anand Saini: https://github.com/anandsaini024
 .. _Ariel Rokem: https://github.com/arokem
-.. _Arne Gottwald: https://github.com/waldie11
 .. _Austin Hurst: https://github.com/a-hurst
 .. _Berk Gerçek: https://github.com/berkgercek
 .. _Bruno Hebling Vieira: https://github.com/bhvieira
 .. _Chris Holdgraf: https://bids.berkeley.edu/people/chris-holdgraf
-.. _Christian O'Reilly: https://github.com/christian-oreilly
 .. _Clemens Brunner: https://github.com/cbrnr
 .. _Daniel McCloy: http://dan.mccloy.info
 .. _Denis Engemann: https://github.com/dengemann
@@ -54,3 +52,4 @@
 .. _Tom Donoghue: https://github.com/TomDonoghue
 .. _William Turner: https://bootstrapbill.github.io/
 .. _Yorguin Mantilla: https://github.com/yjmantilla
+.. _Christian O'Reilly: https://github.com/christian-oreilly
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -19,7 +19,6 @@ The following authors contributed for the first time. Thank you so much! 🤩
 
 * `Christian O'Reilly`_
 * `Berk Gerçek`_
-* `Arne Gottwald`_
 
 The following authors had contributed before. Thank you for sticking around! 🤘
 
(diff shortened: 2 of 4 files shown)


sappelhoff:
Interesting, thanks for the report @drammock. Is it possible that this is an issue in the mne-bids-pipeline rather than mne-bids?

@waldie11 can you have a look as well please?

Finally (also @hoechenberger), should we perhaps run a part of the mne-bids-pipeline tests as part of the mne-bids testsuite?

larsoner:
> Finally (also @hoechenberger), should we perhaps run a part of the mne-bids-pipeline tests as part of the mne-bids testsuite?

The mne-bids-pipeline tests are generally kind of long and require a lot of data transfer, so we run most of them on CircleCI (though there are a couple of datasets that run on GHA). My sense is that so far the incidence of breakage has been fairly low so we could hold off and wait to see if we see more frequent breakages over there due to MNE-BIDS changes. But it looks like it would only be ~30 lines to add an equivalent test here -- wouldn't need to test the "caching" / rerun and could, so would probably just be a simplified (e.g., one OS, one dataset) version of:

https://github.com/mne-tools/mne-bids-pipeline/blob/58f4b76de169c02a824c8462c60406ce1d2f46b7/.github/workflows/tests.yml#L31-L64

I could make a quick PR if it would help and we could see if the runtime is acceptable etc. It wouldn't catch all breakages but probably would catch some.

sappelhoff:
> I could make a quick PR if it would help and we could see if the runtime is acceptable etc. It wouldn't catch all breakages but probably would catch some.

that would be much appreciated!

> Maybe a cause of https://github.com/mne-tools/mne-nirs/pull/601 failures as well?

😳 seems like there is a lot of stuff that our mne-bids tests do not catch.

larsoner:
`mne-nirs` bug does appear to be fixed by this revert PR. If I add this line to debug:
```diff
diff --git a/mne_bids/stats.py b/mne_bids/stats.py
index aeed07b3..df1bcad3 100644
--- a/mne_bids/stats.py
+++ b/mne_bids/stats.py
@@ -71,6 +71,7 @@ def count_events(root_or_path, datatype="auto"):
 
     bids_path.update(datatype=datatype)
 
+    print(f"{repr(bids_path).replace('\n',' ')} ({len(bids_path.match())} files)")
     tasks = sorted(set([bp.task for bp in bids_path.match()]))
 
     all_counts = []
```
on this branch I get:
```
BIDSPath( root: /Users/larsoner/mne_data/fNIRS-motor-group datatype: nirs basename: events.tsv) (5 files)
```
Whereas `main` I see there is no `.match` for the tasks:
```
BIDSPath( root: /Users/larsoner/mne_data/fNIRS-motor-group datatype: nirs basename: events.tsv) (0 files)
```
And there are files like:
```console
$ ls /Users/larsoner/mne_data/fNIRS-motor-group/*/*/*_events.tsv 
/Users/larsoner/mne_data/fNIRS-motor-group/sub-01/nirs/sub-01_task-tapping_events.tsv
/Users/larsoner/mne_data/fNIRS-motor-group/sub-02/nirs/sub-02_task-tapping_events.tsv
/Users/larsoner/mne_data/fNIRS-motor-group/sub-03/nirs/sub-03_task-tapping_events.tsv
/Users/larsoner/mne_data/fNIRS-motor-group/sub-04/nirs/sub-04_task-tapping_events.tsv
/Users/larsoner/mne_data/fNIRS-motor-group/sub-05/nirs/sub-05_task-tapping_events.tsv
```
maybe this will help a test to be added?

sappelhoff:
Thanks for looking into this @larsoner and @drammock.

I unfortunately do not have the time to properly help debugging this, so if it is urgent for you, or you deem it wise, please go ahead with reverting the change.

@waldie11 can maybe then pick it up again in the near future to resolve the failures that our tests and my review did not pick up on.

larsoner:
I wrote a little test that fails on `main` but passes on this PR based on what I wrote above :point_up: I'll wait until https://github.com/mne-tools/mne-python/pull/13129 lands, which should fix CIs, then restart CIs here. Then I'll push that test. Then I think we should merge.

larsoner:
Okay I think I actually found the bug, I'll open a PR. I think there was just a missing `*` :facepalm:

PR #1379, opened 2025-02-25
BUG: Fix bug with path search

larsoner:
Closes #1376

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -12,7 +12,6 @@ import shutil as sh
 from copy import deepcopy
 from datetime import datetime
 from io import StringIO
-from itertools import chain as iter_chain
 from os import path as op
 from pathlib import Path
 from textwrap import indent
@@ -2017,8 +2016,9 @@ def get_entity_vals(
 
         .. versionadded:: 0.17
     include_match : str | array-like of str | None
-        Apply a starting match pragma following Unix style pattern syntax from
-        package glob to prefilter search criterion.
+        Glob-style match string(s) of directories to include in the search (i.e., each
+        must end with ``"/"``). ``None`` (the default) is equivalent to ``"**/"``
+        (within any subdirectory of the BIDS root).
 
         .. versionadded:: 0.17
     with_key : bool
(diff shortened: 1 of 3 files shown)


mne_bids/path.py
```diff
@@ -2542,24 +2544,19 @@ def _return_root_paths(root, datatype=None, ignore_json=True, ignore_nosub=False
 
     if datatype is None and not ignore_nosub:
         search_str = "*.*"
-        paths = root.rglob(search_str)
     else:
         if datatype is not None:
             datatype = _ensure_tuple(datatype)
-            search_str = f"**/{'|'.join(datatype)}/.*"
+            search_str = f"**/{'|'.join(datatype)}/*.*"
```
larsoner, line 2550: This was the real bug. All else is just cleanups + test

larsoner:
Yikes the benchmark test takes 60 seconds (!). It also no longer passes for me locally. I'll see if the speedup comes from
```
        paths = [
            Path(root, fn)
            for fn in glob.iglob(search_str, root_dir=root, recursive=True)
        ]
```
in which case it should be used in both code paths. But that would be weird...

drammock:
I'd also like to fixup a couple things about #1355 while this is open. I see you already restored the `versionchanged` (thanks!). My TODO list:

- some of the changelog entries reference private functions, should reword / remove
- the docstring for `include_match` is IMO very hard to understand, and the param name itself doesn't really help.

larsoner:
Okay should be good to go @drammock

mne_bids/path.py
```diff
@@ -2014,9 +2013,12 @@ def get_entity_vals(
...
         Suffixes to ignore. If ``None``, include all suffixes. This can be helpful for
         ignoring non-data sidecars such as `*_scans.tsv` or `*_coordsystem.json`.
+
+        .. versionadded:: 0.17
     include_match : str | array-like of str | None
-        Apply a starting match pragma following Unix style pattern syntax from
-        package glob to prefilter search criterion.
+        Glob-style match string(s) of directories to include in the search (i.e., each
+        must end with ``"/"``). ``None`` (the default) is equivalent to ``"**/"``
+        (within any subdirectory of the BIDS root).
```
drammock, line 2021:
```suggestion
        Glob-style pattern(s) of *directories* to include in the search (i.e., each
        must end with ``"/"``). ``None`` (the default) is equivalent to ``"**/"``
        (search within any subdirectory of the BIDS root).
```

drammock (approved):
LGTM, just one tiny docstring tweak. I'll apply it and merge. Thanks @larsoner

PR #1389, opened 2025-04-03
Refactor _handle_events_reading to allow extracting annotation information stand-alone

matthiasdold:
PR Description
--------------

**Aim**: 
The `_handle_events_reading` in [read.py](https://github.com/matthiasdold/mne-bids/blob/5d4320f0239c5efac25261d123834b1cb0357b1e/mne_bids/read.py#L526) reads onsets, duration and descriptions/event_ids from the `events.tsv`, then filters the columns for n\a values etc. and finally sets annotations to the raw object based on these filtered version of the `events.tsv` data.

```python
    annot_from_raw = raw.annotations.copy()
    annot_from_events = mne.Annotations(onset=ons, duration=durs, description=descrs)
    raw.set_annotations(annot_from_events)
```

On default, this will encode information from three columns within the `events.tsv`.

In order to [extract additional information from BIDS dataset within MOABB](https://github.com/NeuroTechX/moabb/pull/744) it would be necessary to ensure that all annotations that end up in the raw and whatever is extracted from the `events.tsv` are aligned. To allow for this, it would be necessary to have any filtering that happens in [L526-602](https://github.com/matthiasdold/mne-bids/blob/5d4320f0239c5efac25261d123834b1cb0357b1e/mne_bids/read.py#L526C1-L602C1) accessible standalone (without the need to provide a `raw`).

**Proposition**:
Simply have the first part of the [`_handle_events_reading`](https://github.com/matthiasdold/mne-bids/blob/5d4320f0239c5efac25261d123834b1cb0357b1e/mne_bids/read.py#L526) as its own function which returns the filtered version of the onset, description, duration and event_ids. Assuming that these columns would form a primary key within the `events.tsv`, we could always map additional meta info to the annotations in the raw unambiguously.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,7 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
-- Data from ``events.tsv`` can now be read into an OrderedDict using :func:`mne_bids.events_file_to_annotation_kwargs()`, by `Matthias Dold` (:gh:`1389`)
+- Data from ``events.tsv`` can now be read into an OrderedDict using :func:`mne_bids.events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
 
 
 🧐 API and behavior changes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,7 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
-- `_handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.read.events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
+- :func:`mne_bids.read._handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.read._events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
 
 
 🧐 API and behavior changes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,7 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
-- `_handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
+- `_handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.read.events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
 
 
 🧐 API and behavior changes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -20,6 +20,7 @@ The following authors contributed for the first time. Thank you so much! 🤩
 * `Christian O'Reilly`_
 * `Berk Gerçek`_
 * `Arne Gottwald`_
+* `Matthias Dold`_
 
 The following authors had contributed before. Thank you for sticking around! 🤘
 
@@ -39,6 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
+- :func:`mne_bids.read._handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.read._events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
 
 
 🧐 API and behavior changes
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -40,7 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
-- :func:`mne_bids.read._handle_events_reading()` was refactored to now exposed the filtering of the `events.tsv` data as a separate function :func:`mne_bids.read._events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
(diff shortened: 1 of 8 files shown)


sappelhoff (commented):
Hi @matthiasdold to make this function public, it would need to get a more extensive docstr (numpydoc), and preferably also be used in a public example.

You may (or not) also need to add it here in some way: https://github.com/mne-tools/mne-bids/blob/main/mne_bids/__init__.py

There are also some test failures here that I haven't looked into yet.

matthiasdold:
> Hi @matthiasdold to make this function public, it would need to get a more extensive docstr (numpydoc), and preferably also be used in a public example.
> 
> You may (or not) also need to add it here in some way: https://github.com/mne-tools/mne-bids/blob/main/mne_bids/__init__.py
> 
> There are also some test failures here that I haven't looked into yet.

Hi @sappelhoff, thanks for checking this out. I actually think it could be kept `private` / `_events_file_to_annotation_kwargs` function as in the [first commit](https://github.com/mne-tools/mne-bids/pull/1389/commits/c2d078ae56a173f5ef9d2db53e0d80d849bb1f64). The reason why I was later putting it to public was that I could not get the docs to build and making it public + introducing it to the `__init__.py` at least had sphynx find it. 

Unfortunately, I still cannot get the docs to build, neither locally nor here on the server. Which is very puzzling to me, as the change really is just separating the existing functions... I will have another try later this afternoon, but any hints would be much appreciated.

sappelhoff (commented):
Yes, it would be great if we could cover any differences introduced via this PR.

So you ended up leaving the function private 🤔 I think that if you want to use this in another library, it may be worth making it public. For that you could take one example function in mne-bids that is public, and see where it has to be declared, and then do the same for your function.

Otherwise, you always take the risk that is inherent in using private functions of another software package: backward incompatible changes may come at any time and without notice 🤔 

If you decide that you do NOT want to make the function public and that you DO want to take the risk of using a private function, then you would at least have to adjust your changelog entry, as we usually do not talk about changes to private functions so explicitly. You could write something a bit more generic, like "refactoring of events filtering", under code health or so.

WDYT @drammock @hoechenberger ?

drammock:
> WDYT @drammock

I wasn't super familiar with the internals of what MNE-BIDS did in terms of reading `events.tsv` prior to looking at this PR.  My feeling (from a quick skim and quick think) is that:

1. there should be a public func that reads `events.tsv` files.
2. I'm conflicted about whether it should return an actual annotations object, a dict suitable for passing to `mne.Annotations`, or a dataframe. I guess the Annotations and the DataFrame can be quite easily constructed from the dict, so I guess the dict is most general/flexible.
3. for users who want an events array, they can use `mne.events_from_annotations` and we should mention that in the docstring.
4. The changelog should only mention this new public function; users won't care / don't need to know that it was part of a private function that was refactored into a separate thing.

matthiasdold:
>     2. I'm conflicted about whether it should return an actual annotations object, a dict suitable for passing to `mne.Annotations`, or a dataframe. I guess the Annotations and the DataFrame can be quite easily constructed from the dict, so I guess the dict is most general/flexible.


Operating on a (pandas) dataframe would also lead to a more readable code as we can use standard filtering instead of [`_drop`](https://github.com/mne-tools/mne-bids/blob/5d4320f0239c5efac25261d123834b1cb0357b1e/mne_bids/tsv_handler.py#L92). Although just a subset of the functionality, the dataframe is what I am using in the meantime for [moabb](https://github.com/NeuroTechX/moabb/pull/744/commits/942a52026ca8023d3327bc231a228bb7ed678d00#diff-80115723a6ae060787a6ddfa3b7c8922cfc196a391ef65a7284224a19b99187d).

I didn't go the dataframe route as I wanted to be minimally invasive. Happy to provide a dataframe version as well.

sappelhoff:
Thanks for your view, Dan.

I agree with your points 1, 3 and 4 -- and regarding point 2, I think returning a dict is the way to go. We do not have pandas as a required dependency of mne-bids, and I don't think the current "feature" is enough to warrant it.

matthiasdold:
> @matthiasdold I think you need double backquotes for code samples. This might be why your doc build fails https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

Thx for the suggestion @PierreGtch  - unfortunately this seems not to be the issue. As for the :func: part, all other lines only had single backquotes, so I would not expect this to be a porblem there (the error message states that `mne_bids.events_file_to_annotations_kwargs` cannot be found, however the import works for the pytest... 

I am a bit clueless here...

PierreGtch:
@matthiasdold your issue is not anymore with building the doc but before, when the ci is trying to pull the code
It is probably unrelated to you

doc/whats_new.rst
```diff
@@ -39,6 +40,7 @@ Detailed list of changes
 - Empty-room matching now preferentially finds recordings in the subject directory tagged as `task-noise` before looking in the `sub-emptyroom` directories. This adds support for a part of the BIDS specification for ER recordings, by `Berk Gerçek`_ (:gh:`1364`)
 - Path matching is now implemenented in a more efficient manner within :meth:`mne_bids.BIDSPath.match()` and :func:`mne_bids.find_matching_paths()`, by `Arne Gottwald` (:gh:`1355`)
 - :func:`mne_bids.get_entity_vals()` has a new parameter ``include_match`` to prefilter item matching and ignore non-matched items from begin of directory scan, by `Arne Gottwald` (:gh:`1355`)
+- Data from ``events.tsv`` can now be read into an OrderedDict using :func:`mne_bids.events_file_to_annotation_kwargs()`, by `Matthias Dold` (:gh:`1389`)
```
sappelhoff, line 43:
```suggestion
- Data from ``events.tsv`` can now be read into an OrderedDict using :func:`mne_bids.events_file_to_annotation_kwargs`, by `Matthias Dold` (:gh:`1389`)
```

drammock:
> something is fishy with circleci.

the error looks totally related to me:
https://app.circleci.com/pipelines/github/mne-tools/mne-bids/6995/workflows/6d22be6e-fce2-43b0-a5d2-a0712a9fd896/jobs/9247?invite=true#step-108-0_142

```
/home/circleci/project/doc/whats_new.rst:43: WARNING: py:func reference target not found: mne_bids.events_file_to_annotation_kwargs [ref.func]
```

same error here:
https://github.com/mne-tools/mne-bids/actions/runs/14379592851/job/40319969519?pr=1389#step:6:257

```
/home/runner/work/mne-bids/mne-bids/doc/whats_new.rst:43: WARNING: py:func reference target not found: mne_bids.events_file_to_annotation_kwargs [ref.func]
```

drammock:
add the new public function name to `doc/api.rst`

sappelhoff:
> the error looks totally related to me:

I was talking about this one:

https://app.circleci.com/pipelines/github/mne-tools/mne-bids/6993/workflows/1783535a-8a54-4b94-ac2b-ace4331f6e3f/jobs/9245

image

sappelhoff:
yet https://github.com/mne-tools/mne-bids/pull/1391 is unaffected, so it must be something about this PR.

sappelhoff:
ok fine, the errors that circleci now provides all seem meaningful and related :-) no idea what I was after in https://github.com/mne-tools/mne-bids/pull/1389#issuecomment-2794796715

@matthiasdold this should be possible for you to address. I am +1 for merging once everything is green. thanks a lot!

mne_bids/read.py
```diff
@@ -523,8 +523,70 @@ def _handle_info_reading(sidecar_fname, raw):
...
+            The durations of the events in seconds.
+        - 'description' : np.ndarray
+            The descriptions of the events.
+        - 'event_id' : dict
+            A dictionary mapping event descriptions to integer event IDs.
+
+    Notes
+    -----
+    The function handles the following cases:
+    - If the `trial_type` column is available, it uses it for event descriptions.
```
drammock, line 554: add blank line before bullet list starts

mne_bids/read.py
```diff
@@ -523,8 +523,70 @@ def _handle_info_reading(sidecar_fname, raw):
...
+    >>> df.to_csv(events_file, sep='\t', index=False)
+    >>>
+    >>> # Read the events file using the function
+    >>> events_dict = events_file_to_annotation_kwargs(events_file)
+    >>> events_dict
+    {'onset': array([0.1, 0.2, 0.3]),
+    'duration': array([0.1, 0.1, 0.1]),
+    'description': array(['event1', 'event2', 'event1'], dtype='<U6'),
+    'event_id': {'event1': 1, 'event2': 2}}
+
```
drammock, line 588: remove blank line at end of example

mne_bids/read.py
```diff
@@ -523,8 +523,70 @@ def _handle_info_reading(sidecar_fname, raw):
...
+    ----------
+    events_fname : str
+        The file path to the `events.tsv` file.
+
+    Returns
+    -------
+    dict
+        A dictionary containing the following keys:
+        - 'onset' : np.ndarray
+            The onset times of the events in seconds.
```
drammock, line 543: this line is getting foramtted as a blockquote, that's the source of the doc build error.

drammock:
here's the rendered docstring:
https://output.circle-artifacts.com/output/job/1e8c4f31-cedf-4fcd-a9ef-94de47d8437b/artifacts/0/dev/generated/mne_bids.events_file_to_annotation_kwargs.html

see how the notes section bullet list isn't formatted right, and the formatting in the Returns section is inconsistent.

matthiasdold:
Thanks a lot @drammock  for the detailed explanation! I should get it sorted out now. My first time working with circleci (as you of course must have guessed), so I was not aware that I could look at the artifacts. Thanks for your patients @drammock  and @sappelhoff 😅

mne_bids/tests/test_read.py
```diff
@@ -1466,3 +1468,76 @@ def test_gsr_and_temp_reading():
...
+        subject="01", session="eeg", task="rest", datatype="eeg", root=tiny_bids_root
+    )
+    events_fname = _find_matching_sidecar(bids_path, suffix="events", extension=".tsv")
+
+    # ---------------- plain read --------------------------------------------
+    df = pd.read_csv(events_fname, sep="\t")
+    ev_kwargs = events_file_to_annotation_kwargs(events_fname=events_fname)
+    assert (ev_kwargs["onset"] == df["onset"].values).all()
+    assert (ev_kwargs["duration"] == df["duration"].values).all()
+    assert (ev_kwargs["description"] == df["trial_type"].values).all()
```
drammock, line 1485: [nitpick] this is a case where I'd normally use `np.testing.assert_array_equal`.

mne_bids/tests/test_read.py
```diff
@@ -1466,3 +1468,76 @@ def test_gsr_and_temp_reading():
...
+
+    assert (ev_kwargs_filtered["onset"] == dext_f["onset"].astype(float).values).all()
+    assert (
+        ev_kwargs_filtered["duration"]
+        == dext_f["duration"].replace("n/a", "0.0").astype(float).values
+    ).all()
+    assert (ev_kwargs_filtered["description"] == dext_f["trial_type"].values).all()
+    assert (
+        ev_kwargs_filtered["duration"][0] == 0.0
+    )  # now idx=0, as first row is filtered out
```
drammock, line 1528: idem. Could also consider `pd.testing.assert_frame_equal` but that would require converting the dict to dataframe first.

mne_bids/tests/test_read.py
```diff
@@ -1466,3 +1468,76 @@ def test_gsr_and_temp_reading():
...
+
+    # ---------------- default if missing trial_type  ------------------------
+    tmp_tsv_file = tmp_path / "events.tsv"
```
drammock, line 1531: shouldn't need to redefine this here
```suggestion
```

drammock (approved):
LGTM, thanks for the thorough test. Left a few nitpick comments that you can feel free to ignore or implement.

mne_bids/tests/test_read.py
matthiasdold: As we are using a dict for the return type, I would just stick to comparing the iterables here

matthiasdold:
Once https://github.com/mne-tools/mne-python/pull/13213 is implemented in `mne`, we could also add the additional metadata to the `Annotations` object. This could be part of this PR or a separate one. Let me know what you prefer @drammock and @sappelhoff

sappelhoff (approved):
Thanks @matthiasdold!

sappelhoff:
> This could be part of this PR or a separate one.

a separate one, please! :)

PR #1404, opened 2025-05-24
Gracefully handle absence of `bids-validator`, and allow running it through `npx`

hoechenberger:
@sappelhoff I also made it such that tests are skipped if neither the validator nor `npx` are installed. Is this too lenient? WDYT?

Also wondering if running through `npx` shouldn't be the default, as this would ensure we're always using the latest version of the validator.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/conftest.py
+++ b/mne_bids/tests/conftest.py
@@ -21,7 +21,7 @@ def _bids_validate():
         shell = False
 
     # If neither bids-validator nor npx are available, we cannot validate BIDS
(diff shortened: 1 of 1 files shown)


hoechenberger:
I'll change it to raising of the validator cannot be found or run

Regarding the schema-based validator, this is a JS tool too, no? so we could eventually just call the new validator, but keep the existing tooling that currently calls the legacy validator

PR #1359, opened 2024-12-20
FIX:  Use "recording" for recording entity, not "rec"

scott-huberty:
Fixes #1348

See: https://bids-specification.readthedocs.io/en/stable/modality-specific-files/physiological-recordings.html

As @kaare-mikkelsen pointed out, [BEP 020](https://github.com/bids-standard/bids-specification/pull/1128) will make use of the `recording` entity for eyetracking data.  And since I'm working on #1342 , this should be fixed first 🙂 

Merge checklist
---------------

Maintainer, please confirm the following before merging.
If applicable:

- [ ] All comments are resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [x] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/main/doc/whats_new.rst) is updated
- [ ] New contributors have been added to [CITATION.cff](https://github.com/mne-tools/mne-bids/blob/main/CITATION.cff)
- [x] PR description includes phrase "closes <#issue-number>"

PR #1417, opened 2025-07-02
FIX, DOC: Dont pair recording parameter w/ rec BIDS entity

scott-huberty:
Following up on some things I missed in #1359, I found two places where the API doc suggested that the `recording` parameter refers to the "rec" entity in BIDS (for example in the [BIDSPath](https://mne.tools/mne-bids/stable/generated/mne_bids.BIDSPath.html#mne_bids.BIDSPath.recording) constructor). As we realized before, "rec" actually stands for reconstruction, and there is a stand alone "recording" entity in BIDS.


hoechenberger (approved):

hoechenberger:
LGTM, thanks @scott-huberty

PR #1427, opened 2025-08-12
add optional task- to electrodes.tsv

### LINKED ISSUE #1424: electrodes.tsv file should have task- entity?
The `electrodes.tsv` file that mne-bids writes with `write_raw_bids()` contains impedance values (in my case, from the BrainVision files). If one has two or more recordings in the same session, but for different tasks (e.g. for resting-state open and closed eyes recordings) the impedance values may differ, and this leads to an error when calling `write_raw_bids()` on the second file:

`RuntimeError: Trying to write electrodes.tsv, but it already exists at /bidsroot/rawdata/sub-ML01/ses-01/eeg/sub-ML01_ses-01_space-CapTrak_electrodes.tsv and the contents do not match. You must differentiate this electrodes.tsv file from the existing one, or set "overwrite" to True.`

In `mne_bids/dig.py`, at line 498, the following dict is defined which is used to make the file path:

```python
    coord_file_entities = {
        "root": bids_path.root,
        "datatype": bids_path.datatype,
        "subject": bids_path.subject,
        "session": bids_path.session,
        "acquisition": bids_path.acquisition,
        "space": None if bids_path.datatype == "nirs" else coord_frame,
    }
```

I believe the issue could be solved by adding `"task": bids_path.task` to this dict (perhaps optionally?). This would be [BIDS-compliant](https://bids-specification.readthedocs.io/en/latest/modality-specific-files/electroencephalography.html#electrodes-description-_electrodestsv). 

In case it's relevant, I found a related but different issue: #766.

alm180:
PR Description
--------------

This PR adds the `task-` entity to the `electrodes.tsv` files as discussed in #1424 

- fixes #1424 

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -1,0 +1,6 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot[bot]
+      - pre-commit-ci[bot]
+      - github-actions[bot]
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out a copy of the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Check whether the citation metadata from CITATION.cff is valid
         uses: citation-file-format/cffconvert-github-action@2.0.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,11 +18,11 @@ jobs:
   package:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
           fetch-tags: true
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
         with:
           python-version: '3.13'
       - name: Install dependencies
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -21,12 +21,12 @@ jobs:
       matrix:
         python-version: ["3.13"]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/cache@v4
@@ -54,7 +54,7 @@ jobs:
         python-version: ["3.10", "3.13"]  # Oldest and newest supported versions
     steps:
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
 
@@ -69,7 +69,7 @@ jobs:
         pip check
         python -m pip install build twine
 
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
@@ -88,7 +88,7 @@ jobs:
     - name: Remove sdist install
       run: python -m pip uninstall -y mne-bids
 
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
@@ -107,7 +107,7 @@ jobs:
     - name: Remove wheel install
       run: python -m pip uninstall -y mne-bids
 
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
@@ -174,18 +174,18 @@ jobs:
       TZ: Europe/Berlin
       FORCE_COLOR: true
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
 
     - name: Set up Node.js
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v5
       with:
         node-version: 20
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
 
@@ -304,12 +304,12 @@ jobs:
       matrix:
         python-version: ["3.13"]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
         fetch-tags: true
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/cache@v4
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.7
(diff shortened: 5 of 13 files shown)


hoechenberger:
Thanks for your contribution, @alm180!

I'm wondering if we even need a switch to control this behavior – why not always write the task entity? @sappelhoff WDYT?

sappelhoff:
> why not always write the task entity? @sappelhoff WDYT?

I think it should remain optional as it's optional in the BIDS spec, too and for many people, task specific sensor specs do not make sense, I think 🤔 but I am not hard set on this

scott-huberty:
@alm180 looks like the test suite is passing which is a good start! I think the biggest thing left to do is add a test (or update a currently existing test). Ideally all new code is covered by test. Personally I also think that tests can hep devs understand the new/updated API (but maybe that's just me).

Others can chime in if they think this is overkill, but I think that I would add something like this to `mne_bids/tests/test_dig` or `mne_bids/tests/test_write`


```python
# raw_fname is already defined in many of our test modules, but putting it below to be explicit
raw_fname = mne.datasets.testing.data_path() / "MEG" / "sample" / "sample_audvis_trunc_raw.fif"


def test_task_specific_channels_sidecar(tmp_path):
    raw = mne.io.read_raw_fif(raw_fname).pick("eeg")
    raw_foo = raw.copy().load_data().pick(slice(None, len(raw.ch_names) // 2))
    raw_bar = raw.copy().load_data().pick(slice(len(raw.ch_names) // 2, None))

    bpath_kwargs = dict(
        root=tmp_path,
        subject="01",
        session="01",
        run="01",
        acquisition="01",
        task="foo",
    )

    write_kwargs = dict(
        allow_preload=True,
        electrodes_tsv_task=True,
        format="BrainVision",
    )


    bpath = mne_bids.BIDSPath(**bpath_kwargs)

    bpath_foo =mne_bids.write_raw_bids(raw_foo, bpath, **write_kwargs)

    # Read the raw data back from BIDS format

    bpath_kwargs.update(task="bar")
    bpath = mne_bids.BIDSPath(**bpath_kwargs)
    bpath_bar = mne_bids.write_raw_bids(raw_bar, bpath, **write_kwargs)
    # assert that the electrodes.tsv / channels.tsv filenames are
    # what we expect and their content is correct

    raw_foo = mne_bids.read_raw_bids(bpath_foo)
    raw_bar = mne_bids.read_raw_bids(bpath_bar)
    # assert that the ch_names and pos are being loaded from the
    # correct coord file.
```

alm180:
Hi, thank you so much for the kind welcome and especially @scott-huberty for the help. I've made a commit with the test based on what you sent. I think it covers the expected behavior, but I'm open to adding more if needed.

mne_bids/tests/test_dig.py
```diff
@@ -378,3 +378,87 @@ def test_electrodes_io(tmp_path):
...
+
+    bpath_kwargs.update(task="bar")
+    bpath = mne_bids.BIDSPath(**bpath_kwargs)
+
+    # without specifying a task, writing a second electrodes.tsv with the same
+    # filename but different contents fails
+    with pytest.raises(
+        RuntimeError, match="Trying to write electrodes.tsv, but it already exists"
+    ):
+        bpath_bar = mne_bids.write_raw_bids(raw_bar, bpath, **write_kwargs)
```
scott-huberty, line 464: I am not sure that you need to confirm that an error is raised if you try to overwrite a task-agnostic channels sidecar,  because that case seems to already be tested here:
 https://github.com/mne-tools/mne-bids/blob/59e61e3a2edbc358bd926a65e69f6ef3a3b81603/mne_bids/tests/test_write.py#L3060-L3063

mne_bids/write.py
```diff
@@ -1432,6 +1432,7 @@ def write_raw_bids(
     event_id=None,
     event_metadata=None,
     extra_columns_descriptions=None,
+    electrodes_tsv_task=False,
```
scott-huberty, line 1435: You placed the new parameter `electrodes_tsv_task=False` before the asterisk `*` - but I would actually move `electrodes_tsv_task=False`  _after it_. A bare asterisk in the function signature forces any parameters that are defined after it to be passed as keyword arguments when calling the function. And since this parameter is new, I think that it makes sense to force the user to pass it as a keyword-argument (`electrodes_tsv_task=False`), instead of as a positional one (`False`)

mne_bids/write.py
```diff
@@ -1535,6 +1536,9 @@ def write_raw_bids(
     extra_columns_descriptions : dict | None
         A dictionary that maps column names of the ``event_metadata`` to descriptions.
         Each column of ``event_metadata`` must have a corresponding entry in this.
+    electrodes_tsv_task : bool
+        Add the ``task-`` entity to the ``BIDSPath`` of the
+        ``electrodes.tsv`` file. Defaults to ``False``.
```
scott-huberty, line 1541:
```suggestion
        Add the ``task-`` entity to the ``electrodes.tsv`` filename. Defaults to ``False``.
```

I think that this makes more sense but let me know if you disagree or if I am misunderstanding something.

scott-huberty (changes requested):
Thanks for adding a test @alm180 - almost there!!!! Just a few comments.

After these are addressed @sappelhoff I think that this is good to merge. I've also interacted with this new parameter locally and everything appears to work as expected.

mne_bids/tests/test_dig.py
```diff
@@ -378,3 +378,87 @@ def test_electrodes_io(tmp_path):
...
+
+    write_kwargs = dict(
+        allow_preload=True,
+        electrodes_tsv_task=False,
+        format="BrainVision",
+    )
+
+    bpath = mne_bids.BIDSPath(**bpath_kwargs)
+    bpath_foo = mne_bids.write_raw_bids(raw_foo, bpath, **write_kwargs)
+
```
scott-huberty, line 455:
```suggestion

elpath_foo = bpath_foo.find_matching_sidecar(suffix="electrodes", extension=".tsv")
    assert (
        elpath_foo.name == "sub-01_ses-01_acq-01_space-CapTrak_electrodes.tsv"
    )
    
```

Maybe you can explicitly check that the task wasn't written to the electrodes sidecar.

mne_bids/tests/test_dig.py
```diff
@@ -378,3 +378,87 @@ def test_electrodes_io(tmp_path):
...
+    coordpath_foo = bpath_foo.find_matching_sidecar(
+        suffix="coordsystem", extension=".json"
+    )
+    coordpath_bar = bpath_bar.find_matching_sidecar(
+        suffix="coordsystem", extension=".json"
+    )
+
+    assert coordpath_foo.name == "sub-01_ses-01_acq-01_space-CapTrak_coordsystem.json"
+    assert coordpath_bar.name == "sub-01_ses-01_acq-01_space-CapTrak_coordsystem.json"
+
```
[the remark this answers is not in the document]
scott-huberty: Per my [newfound realization](https://github.com/mne-tools/mne-bids/pull/1427/files#r2274988275), the test I suggested above isn't quite right. Since this PR deals with `electrodes.tsv`, we actually want to check that the channel locations are what we expect:

```python
raw_foo_want_pos = raw_foo.get_montage().get_positions()["ch_pos"]["EEG 029"]
raw_foo = mne_bids.read_raw_bids(bpath_foo)
np.testing.assert_allclose(
    raw_foo.get_montage().get_positions()["ch_pos"]["EEG 029"], raw_foo_want_pos
)
```

alm180:
Hi @scott-huberty I've integrated the suggested changes. Thanks for your help!

scott-huberty (approved):
Thanks @alm180 !

@sappelhoff ready for your review.

sappelhoff (approved):
> @sappelhoff ready for your review.

very sorry about that, but I completely missed this.

I just rebased this branch and resolved conflicts. Changes look good to me -- if everything passes, we can merge.

Thanks for the submission @alm180 and thanks for the review @scott-huberty 

@scott-huberty I have given you "write" access for mne-bids, so that at least your reviews are shown in "green" when you do them 🙂

PR #1428, opened 2025-08-12
FIX: sphinx-gallery installation

scott-huberty:
We want to install `sphinx-gallery` in a way that ensures `.git` metadata is included, otherwise installation will fail because `setuptools-scm` won't be able to infer the sphinx-gallery package version.

see: https://github.com/mne-tools/mne-bids/issues/1424#issuecomment-3180838806

Let's see if this fixes the doc CI.. and @sappelhoff @hoechenberger let me know either of you still think that this is an upstream packaging bug.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ doc = [
   "seaborn",
   "sphinx-copybutton",
   "sphinx>=7.4.7",
-  "sphinx_gallery @ https://github.com/sphinx-gallery/sphinx-gallery/archive/refs/heads/master.zip",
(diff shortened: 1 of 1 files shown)


hoechenberger:
No, we shouldn't have to do this. The setuptools-scm team needs to fix this upstream. They need to add a `.git_archival.txt` file to their repository.

scott-huberty:
> No, we shouldn't have to do this. The sphinx gallery team needs to fix this upstream. They need to add a `.git_archival.txt` file to their repository.
> 
> Edit: And a `.gitattributes` file.

Ah OK. feel free to close this, or merge it as a temporary fix.

hoechenberger:
I'll let sb else take action, just wanted to give you a pointer :)

PR #1430, opened 2025-08-25
Add motion to BIDSPath

JuliusWelzel:
PR Description
--------------

This PR adds motion as an allowed datatype for the BIDSPath with relevant changes in the config.py.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -6,7 +6,7 @@
 from mne import io
 from mne.io.constants import FIFF
 
-BIDS_VERSION = "1.7.0"
+BIDS_VERSION = "1.9.0"
 
 PYBV_VERSION = "0.7.3"
 EEGLABIO_VERSION = "0.0.2"
@@ -15,17 +15,19 @@ DOI = """https://doi.org/10.21105/joss.01896"""
 
 EPHY_ALLOWED_DATATYPES = ["meg", "eeg", "ieeg", "nirs"]
 
-ALLOWED_DATATYPES = EPHY_ALLOWED_DATATYPES + ["anat", "beh"]
+ALLOWED_DATATYPES = EPHY_ALLOWED_DATATYPES + ["anat", "beh", "motion"]
 
 MEG_CONVERT_FORMATS = ["FIF", "auto"]
 EEG_CONVERT_FORMATS = ["BrainVision", "auto"]
 IEEG_CONVERT_FORMATS = ["BrainVision", "auto"]
 NIRS_CONVERT_FORMATS = ["auto"]
+MOTION_CONVERT_FORMATS = ["tsv", "auto"]
 CONVERT_FORMATS = {
     "meg": MEG_CONVERT_FORMATS,
     "eeg": EEG_CONVERT_FORMATS,
     "ieeg": IEEG_CONVERT_FORMATS,
     "nirs": NIRS_CONVERT_FORMATS,
+    "motion": MOTION_CONVERT_FORMATS,
 }
 
 # Orientation of the coordinate system dependent on manufacturer
@@ -147,12 +149,17 @@ allowed_extensions_nirs = [
     ".snirf",  # SNIRF
 ]
 
+allowed_extensions_motion = [
+    ".tsv",  # Tab-separated values
+]
+
 # allowed extensions (data formats) in BIDS spec
 ALLOWED_DATATYPE_EXTENSIONS = {
     "meg": allowed_extensions_meg,
     "eeg": allowed_extensions_eeg,
     "ieeg": allowed_extensions_ieeg,
     "nirs": allowed_extensions_nirs,
+    "motion": allowed_extensions_motion,
 }
 
 # allow additional extensions that are not BIDS
@@ -190,6 +197,7 @@ ALLOWED_FILENAME_SUFFIX = [
     "physio",
     "stim",  # behavioral
     "nirs",
+    "motion",  # motion
 ]
 
 # converts suffix to known path modalities
@@ -227,6 +235,7 @@ ALLOWED_PATH_ENTITIES = (
     "description",
     "suffix",
     "extension",
+    "tracking-system",
 )
 ALLOWED_PATH_ENTITIES_SHORT = {
     "sub": "subject",
@@ -239,6 +248,7 @@ ALLOWED_PATH_ENTITIES_SHORT = {
     "recording": "recording",
     "split": "split",
     "desc": "description",
+    "track-sys": "tracking_system",
 }
 
 # Annotations to never remove during reading or writing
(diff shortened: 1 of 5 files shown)


JuliusWelzel:
The CI's fail for two example scripts [anonymize_dataset.py](https://github.com/mne-tools/mne-bids/blob/main/examples/anonymize_dataset.py) and [bidspath.py](https://github.com/mne-tools/mne-bids/blob/main/examples/bidspath.py) [(Error here)](https://app.circleci.com/pipelines/github/mne-tools/mne-bids/7265/workflows/9d3f0e58-4606-4618-8e3d-1803ee414206/jobs/9567/parallel-runs/0/steps/0-107).

If I run the same scripts locally, I cannot recreate the same error. @sappelhoff, maybe you have an idea how to solve this?

scott-huberty:
@JuliusWelzel If I checkout your branch locally, I do get the same error that the CI's are hitting:

```python

RuntimeError: Found more than one matching data file for the requested recording. While searching:
    BIDSPath(
    root: /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids
    datatype: eeg
    basename: sub-01_ses-eeg)
Found 5 paths:
    /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_electrodes.tsv
    /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_space-CapTrak_electrodes.tsv
    /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_channels.tsv
    /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_eeg.vhdr
    /Users/scotterik/devel/repos/mne-bids/mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_events.tsv
```

I think I see the problem.

mne_bids/config.py
```diff
@@ -147,12 +149,17 @@
     ".snirf",  # SNIRF
 ]
 
+allowed_extensions_motion = [
+    ".tsv",  # Tab-separated values
+]
+
```
scott-huberty, line 155: This seems to be the line causing the issue, which makes sense because the EEG BIDS datasets have sidecar files that are just text files. So adding `.tsv` to the allowed datafile extensions seems to be confusing some of mne_bids internal functions..

Motion data might be the first case where data are stored in simple text files, which is why we are hitting this problem now. But I know that Eye-tracking BIDS will have the same issue..  So supporting these data might require more refactoring of the codebase.

scott-huberty: As a grumpy sidenote, I never understood why BIDS encourages storing some modalities of data in text files... !

mne_bids/config.py
```diff
@@ -190,6 +197,7 @@
     "physio",
     "stim",  # behavioral
     "nirs",
+    "motion",  # motion
```
drammock, line 200: redundant comment

```suggestion
    "motion",
```

drammock:
[`49b5c02` (#1430)](https://github.com/mne-tools/mne-bids/pull/1430/commits/49b5c02d114852c3bfdbada8aa0cbfc152336a15) should fix the tests. Honestly I'm not sure it's the right fix, but to me the way it was before was buggy / coincidentally working. Adding `motion` revealed a bug IMO.  But we should check with @hoechenberger or @sappelhoff, whether my fix is the right one.

mne_bids/config.py
sappelhoff: without having looked into it too deeply, that could be something that needs refactoring, yes.

sappelhoff:
> Honestly I'm not sure it's the right fix, but to me the way it was before was buggy / coincidentally working. 

the fix looks good to me

JuliusWelzel:
Hi,

thanks for the help with fixing the tests. All CI's are passing now. One question from my side:
The `tracksys` path entity was introduced with Motion BIDS. In https://github.com/mne-tools/mne-bids/pull/1430/commits/0373d1888011ee7fd22b6e4286ed19e7a1da1eee I changed all `tracking-system` to `tracking_system`. One reason for this was that the  [`BIDSPath.__init__`](https://github.com/JuliusWelzel/mne-bids/blob/8f1ff0eb8a72d22e8d27d9524c0d9efeef24e552/mne_bids/path.py#L388) function cannot take `tracking-system` with a hyphen.  Another option would be that I simply rename it to `trackingsystem` to follow the current format. Any thoughts on this?

drammock:
> Another option would be that I simply rename it to `trackingsystem` to follow the current format. Any thoughts on this?

Underscore seems ok to me.

sappelhoff (commented):
@JuliusWelzel could you please also add a what's new entry and resolve the conflicts?

sappelhoff (approved):
Cool, LGTM! Thanks @JuliusWelzel.

If this is doing what you need for mne-bids to be helpful for motion data, then I think this can be merged as is.

JuliusWelzel:
Thanks @sappelhoff.

The changes would now allow me to use mne-bids to handle the path to `motion`, but not write motion data. 
Once the class is created, I would "manually" update it to write my data in BIDS-like manner into the dir:

```python
import numpy as np
from mne_bids import BIDSPath

# define motion-bids relevant information
subject_id = "01"
session = "01"
task = "walking"
tracking_system = "imu1"

# initialize BIDSPath for motion data
motion_raw_bids_path = BIDSPath(subject=subject_id, task=task, datatype='motion', tracking_system=tracking_system, root=bids_root)
print(motion_raw_bids_path)
motion_raw_bids_path.mkdir()

# generate random data for 9 markers and 10 seconds of data at 200Hz srate
raw_motion_data = np.random.rand(9, 10 * 200) # 9 channels, 10 seconds, 200 Hz

#store as tsv without headers
np.savetxt(str(motion_raw_bids_path), raw_motion_data, delimiter='\t', header='', comments='')

# update path for motion.json
motion_json_bids_path = motion_raw_bids_path.update(extension='.json')
print(motion_json_bids_path)

# update path for channels
motion_channels_bids_path = motion_raw_bids_path.update(suffix='channels', extension='.tsv')
print(motion_channels_bids_path)
```

As long as MNE does not want/allow motion data the current approach is already helpful. However, I am open to extend the write_raw_bids function to allow motion data, but would need a little support.

If you think this the current approach is to much of an edge case and not relevant to *Brain Imaging Data*, I am happy not to merge :)

sappelhoff:
Thanks for the explanation @JuliusWelzel I am happy with how you aim to use mne-bids for motion data. And will merge this, thanks!

Looking forward to future enhancements (to mne or mne-bids) for motion data compatibility. 🚀

PR #1470, opened 2025-11-04
MAINT: Sort in pre-commit

larsoner:
A comment over in #1466 made me realize we should add this sorter. We use it in MNE-Python. Nice it doesn't catch anything new here! (I deleted the duplicate entry same is in #1466, though.)


drammock:
more filelock failures in this PR @bruAristimunha

https://github.com/mne-tools/mne-bids/actions/runs/19075239403/job/54488868684?pr=1470#step:22:359

PR #1475, opened 2025-11-06
add curryreader as full dep

drammock:
PR Description
--------------

https://github.com/mne-tools/mne-python/pull/13176 switched MNE-Python to use https://github.com/mne-tools/curry-python-reader/ for reading Neuroscan Curry files.  CIs that use MNE main have started failing for not having curryreader installed.

closes #1477 (supersedes)
closes #1478 (supersedes)

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -10,6 +10,7 @@ BIDS_VERSION = "1.9.0"
 
 PYBV_VERSION = "0.7.3"
 EEGLABIO_VERSION = "0.0.2"
+CURRYREADER_VERSION = "0.1.2"
 
 DOI = """https://doi.org/10.21105/joss.01896"""
 
--- a/mne_bids/tests/test_write.py
+++ b/mne_bids/tests/test_write.py
@@ -47,6 +47,7 @@ from mne_bids import (
 )
 from mne_bids.config import (
     BIDS_COORD_FRAME_DESCRIPTIONS,
+    CURRYREADER_VERSION,
     EEGLABIO_VERSION,
     PYBV_VERSION,
     REFERENCES,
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -62,6 +62,7 @@ doc = [
 ]
 # Dependencies for using all mne_bids features
 full = [
+  "curryreader <= 0.1.2",
```
drammock:
```suggestion
  "curryreader >= 0.1.2",
```

drammock:
progress so far: old curry reader code and new code do slightly different things. With https://github.com/mne-tools/mne-testing-data/blob/master/curry/test_bdf_stim_channel%20Curry%208.cdt:

- old reader: each channel in `raw.info['chs']` has `cal=1e-6, range=1.0`, and `raw._orig_units = {}`. Sets `cal` directly.
- new reader: each channel in `raw.info['chs']` has `cal=1.0, range=1.0`, and `raw._orig_units = {'C3': 'µV', 'C4': 'µV', 'Cz': 'µV'}`. Sets `raw._cals` directly after `BaseRaw.__init__` is finished.

the difference in `_orig_units` is what causes MNE-BIDS to write `µV` into the "units" column of channels.tsv now (which it didn't before), and that is the direct cause of the test failure. That said, the data arrays end up the same (to within `assert_allclose` tolerance), and after a round-trip through `write_raw_bids` (as BrainVision format) and `read_raw_bids`, raws read with both old and new curry readers end up as `cal=0.1`, `range=1e-6`, and `_orig_units={'C3': 'µV', 'C4': 'µV', 'Cz': 'µV'}`

I think the behavior of the new reader is correct w/r/t `_orig_units`, because:

1. the curry header does say "uV" in it
2. the curry data (before it's converted to SI units) is on the order of 7000-16000 (plausible as uV for EEG data)

Therefore, I think (?) it is MNE-BIDS that is incorrect here, in that it is writing the `channels.tsv` "unit" column based on original unit, rather than the unit that is actually being written to disk; using original unit might make sense for cases where the file is not preloaded / only copied / no format conversion occurs, but in this case it's reading Curry and writing BrainVision so IIUC the BrainVision data will be in V (right?)

dominikwelke:
hi @drammock - 
sorry for late reply!

just to confim - 
you are correct, my new curry reader sets `orig_units` to what is written in the file header. 
in all the files I saw so far this was either `µV` or `fT`. should there be any other unit, this will be copied verbatim with `cal=1.0` 

cals are defined as:
```
    cals = [
        1.0 / 1e15 if (u == "fT") else 1.0 / 1e6 if (u == "uV") else 1.0 for u in units
    ]
```

and then set directly via:
```
    self._cals = np.array(cals)
```
(before loading any data)

dominikwelke:
> Therefore, I think (?) it is MNE-BIDS that is incorrect here, in that it is writing the `channels.tsv` "unit" column based on original unit, rather than the unit that is actually being written to disk;

i also looked into the mne-bids code, and yes -  this seems to be the case.
unit written to `channels.tsv` is taken from `raw._orig_units`, and `pybv.write_brainvision` is handed units (`µV`) to allow appropriate scaling but there seems to be no feedback to reflect this unit conversion in the tsv

drammock:
edit for correctness/posterity: mne-bids isn't using mne's export functionality, it is calling pybv directly. But it does the same thing that MNE-Python does (always write V data as uV):

https://github.com/mne-tools/mne-bids/blob/1e0a96e132fc904ba856d42beaa9ddddb985f1ed/mne_bids/write.py#L1275-L1283

PR #1477, opened 2025-11-10
fix units in channels.tsv for EEG (& other Volt-based) channels

drammock:
PR Description
--------------
All valid EEG export formats (BrainVision, EDF, and EEGLAB .set) convert V -> µV when writing, but µV wasn't getting written to channels.tsv units column in many cases.

see https://github.com/mne-tools/mne-bids/pull/1475#issuecomment-3513669141

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -27,6 +27,7 @@ The following authors contributed for the first time. Thank you so much! 🤩
 The following authors had contributed before. Thank you for sticking around! 🤘
 
 * `Stefan Appelhoff`_
+* `Daniel McCloy`_
 
 
 Detailed list of changes
@@ -60,6 +61,7 @@ Detailed list of changes
 - Updated MEG/iEEG writers to satisfy the stricter checks in the latest BIDS validator releases: BTi/4D run folders now retain their ``.pdf`` suffix (falling back to the legacy naming when an older validator is detected), KIT marker files encode the run via the ``acq`` entity instead of ``run``, datasets lacking iEEG montages receive placeholder ``electrodes.tsv``/``coordsystem.json`` files, and the ``AssociatedEmptyRoom`` entry stores dataset-relative paths  by `Bruno Aristimunha`_ (:gh:`1449`)
 - Made the lock helpers skip reference counting when the optional ``filelock`` dependency is missing, preventing spurious ``AttributeError`` crashes during reads, by `Bruno Aristimunha`_ (:gh:`1469`)
 - Fixed a bug in :func:`mne_bids.read_raw_bids` that caused it to fail when reading BIDS datasets where the acquisition time was specified in local time rather than UTC only in Windows, by `Bruno Aristimunha`_ (:gh:`1452`)
+- Fixed bug in :func:`~mne_bids.write_raw_bids` where incorrect unit was sometimes written into ``channels.tsv`` file when converting data to BrainVision, EDF, or EEGLAB formats, by `Daniel McCloy`_ (:gh:`1477`)
 
 ⚕️ Code health
 ^^^^^^^^^^^^^^


mne_bids/write.py
```diff
@@ -193,10 +193,21 @@ def _channels_tsv(raw, fname, overwrite=False):
...
+    # If `raw._orig_units` is missing (or if data are being *converted*),
+    # unit is determined by destination format:
+    #   - `eeglabio.raw.export_set` always converts V to uV
+    #   - `mne.export._edf_bdf._export_raw_edf_bdf` always converts V to uV
+    #   - `pybv.write_brainvision` converts V to uV by default (and we don't alter that)
     else:
-        units = [_unit2human.get(ch_i["unit"], "n/a") for ch_i in raw.info["chs"]]
+        units = [
+            "µV"
+            if ch_i["unit"] == FIFF.FIFF_UNIT_V
```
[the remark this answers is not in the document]
drammock: @agramfort I expanded the comment, and also fixed the logic a bit, but I realize now it's still not correct. I need to check not just whether we're converting formats, but also what destination format is...  as it's written now it will write µV for EEG channels going into a FIFF file (which is obv. wrong). This logic should only apply when converting to .set, .vhdr, .edf

mne_bids/write.py
```diff
@@ -193,11 +197,32 @@ def _channels_tsv(raw, fname, overwrite=False):
...
+    # https://github.com/bids-standard/pybv/blob/2832c80ee00d12990a8c79f12c843c0d4ddc825b/pybv/io.py#L40
+    # https://github.com/mne-tools/mne-bids/blob/1e0a96e132fc904ba856d42beaa9ddddb985f1ed/mne_bids/write.py#L1279-L1280
+    if convert_fmt:
+        volt_like = "µV" if convert_fmt in ("BrainVision", "EDF", "EEGLAB") else "V"
+        units = [
+            volt_like
+            if ch_i["unit"] == FIFF.FIFF_UNIT_V
+            else _unit2human.get(ch_i["unit"], "n/a")
+            for ch_i in raw.info["chs"]
+        ]
```
drammock: @agramfort I think the logic is OK now, do you agree?

agramfort (approved):

mne_bids/write.py
```diff
@@ -193,11 +197,32 @@ def _channels_tsv(raw, fname, overwrite=False):
...
+    # https://github.com/jackz314/eeglabio/blob/3961bb29daf082767ea44e7c7d9da2df10971c37/eeglabio/raw.py#L57
+    #
+    #   - `mne.export._edf_bdf._export_raw_edf_bdf` always converts V to µV, cf:
+    # https://github.com/mne-tools/mne-python/blob/1b921f4af5154bad40202d87428a2583ef896a00/mne/export/_edf_bdf.py#L61-L63
+    #
+    #   - `pybv.write_brainvision` converts V to µV by default (and we don't alter that)
+    # https://github.com/bids-standard/pybv/blob/2832c80ee00d12990a8c79f12c843c0d4ddc825b/pybv/io.py#L40
+    # https://github.com/mne-tools/mne-bids/blob/1e0a96e132fc904ba856d42beaa9ddddb985f1ed/mne_bids/write.py#L1279-L1280
+    if convert_fmt:
+        volt_like = "µV" if convert_fmt in ("BrainVision", "EDF", "EEGLAB") else "V"
```
dominikwelke: this logic implies that all file formats except bv, edf and eeglab store voltages in `V`.
is this correct? 
or is it rather correct, that all other formats dont do any conversion? then we would still need to access `orig_unit` here

maybe practically irrelevant if `mne_bids` only supports these export formats (for eeg they are currently the only formats allowed in BIDS), but couldnt there be MEG formats that store both MEG and EEG in the same file (like curry does)?

drammock: Also the user can request/force conversion in spite of what the bids standard requires

PR #669, opened 2020-12-29
[MRG] Fix anonymization in EDF files to be EDF/EDFBrowser-compliant

adam2392:
PR Description
--------------

Closes: #667 

Working on this + unit tests.

Merge checklist
---------------

Maintainer, please confirm the following before merging:

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy
- [ ] PR title starts with [MRG]
- [ ] [whats_new.rst](https://github.com/mne-tools/mne-bids/blob/master/doc/whats_new.rst) is updated
- [ ] PR description includes phrase "closes <#issue-number>"

PR #1479, opened 2025-11-13
make EDF writer handle anonymized meas_date

drammock:
PR Description
--------------

EDF/BDF formats restrict the `startdate` (equivalent of meas_date) to be 1985-01-01 or later, which conflicts with the BIDS standard for anonymization (which requires anonymized dates to be before 1920-01-01). In #669 a decision was made to write EDF(+) files that violate the BIDS standard but honor the EDF standard (making them viewable in EDFBrowser), while writing the BIDS-compliant anonymized date in `scans.tsv`. However, it was only implemented in cases where the anonymized data started out in EDF format (i.e., only `mne_bids/copyfiles.py` was changed).  In cases where the data start in another format, are anonymized, and then written to EDF(+), the result will be an invalid EDF file. This PR fixes that.  As with #669, the BIDS-compliant anonymized date will still be written to `scans.tsv`; only the EDF file itself will be modified.

The PR also does a small internal refactor to allow both EDF and BDF writing; this is needed for #1129, because EMG modality will have BDF(+) as a RECOMMENDED format.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -64,6 +64,7 @@ Detailed list of changes
 - Made the lock helpers skip reference counting when the optional ``filelock`` dependency is missing, preventing spurious ``AttributeError`` crashes during reads, by `Bruno Aristimunha`_ (:gh:`1469`)
 - Fixed a bug in :func:`mne_bids.read_raw_bids` that caused it to fail when reading BIDS datasets where the acquisition time was specified in local time rather than UTC only in Windows, by `Bruno Aristimunha`_ (:gh:`1452`)
 - Fixed bug in :func:`~mne_bids.write_raw_bids` where incorrect unit was sometimes written into ``channels.tsv`` file when converting data to BrainVision, EDF, or EEGLAB formats, by `Daniel McCloy`_ (:gh:`1475`)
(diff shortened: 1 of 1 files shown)


drammock:
> I would be happy to have the min python pushed to 3.11.

will do

> It's still 11 months to go until 3.10's EOL [...]

FWIW [SPEC-0](https://scientific-python.org/specs/spec-0000/) would recommend already setting 3.12 as the minimum. So switching from 3.10 to 3.11 is already a bit conservative from the perspective of our ecosystem peers.

drammock:
OK CIs are happy except for one more case of 
`mne_bids/tests/test_fileio.py:545: in test_lock_refcount_multiprocess` (which we're handling separately in #1480).  The 6 "pending" checks are cruft due to the CI job names changing when I changed the Python version, and the codecov/patch result is [spurious](https://app.codecov.io/gh/mne-tools/mne-bids/pull/1479?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=mne-tools) (no new missed lines).

Ready for review/merge

PR #1503, opened 2026-01-14
Improve read BIDS example

cbrnr:
I just followed along the example and found some opportunities to improve it. Most importantly, the MNE sample dataset (1.5GB) was downloaded although it was never used/needed here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,10 @@ jobs:
                 echo "source /home/circleci/mne_bids_env/bin/activate" >> $BASH_ENV
 
         - run:
+            name: Pre-create MNE config
+            command: mkdir -p /home/circleci/.mne
+
+        - run:
             name: Install Python packages
             command: |
               python -m pip install --upgrade --progress-bar off pip


cbrnr:
Also, macOS jobs are failing because of some old `scipy.special.sph_harm` deprecation that should have been fixed in MNE-Python a long time ago? Very likely because the "stable" version is still pinned at `maint/1.9` (should be fixed in #1504).

larsoner:
Nowadays the GitHub users themselves have to register on CircleCI (@cbrnr you could do this if you want, it's free and quick). It's a minor annoyance introduced a few months ago, so far it has been okay at the MNE-Python end at least just clicking the "run workflow from failed" button on the CircleCI UI as a maintainer when needed

cbrnr:
PS: I did sign up / log in to CircleCI via my GitHub account, but I still cannot restart the job...

sappelhoff:
maybe you don't have sufficient rights for mne-bids. I restarted it for you.

cbrnr:
Thanks @sappelhoff. Whatever it is, it is annoying and I would turn this feature off, otherwise this will hit every contributor who doesn't have a CircleCI account.

sappelhoff:
> Can we make the main area a bit wider so that this doesn't happen?

fine with me

sappelhoff:
> but can we not switch to another provider such as GitHub Actions?

The main reason to stick with circleci is their service to render the built artifacts (docs) accessibly as a website as in https://output.circle-artifacts.com/output/job/f46731b9-1e1e-4d7c-9ee6-7c6ed0d79f56/artifacts/0/dev/index.html

this is convenient because reviewers do not have to download, extract. and open the build artifacts, as is the case with GH actions. 

If we can find a way around this, I would be open to using the alternative.

cbrnr:
The horizontal space can definitely be moved to a new issue. However, I think we should try to resolve the first two issues regarding the warnings (https://github.com/mne-tools/mne-bids/pull/1503#issuecomment-3753907953). @larsoner do you know if we can get rid of those?

PR #1512, opened 2026-02-05
Support BIDS compliant eyetracking data I/O (BEP 020)

scott-huberty:
Seeing what it will take to read/write eyetracking data according to BIDS-specification BEP 020.

I had to hack around the internals a bit because the BIDS spec for eyetracking data differs a bit from what we've seen with other modalities (AFAIK). For example:

1) instead of storing data in a single binary file, the data _for each eye_ gets written to its own separate text file.
2) eyetracking data does not get its own modality/datatype. If eyetracking data is recorded on its own, it gets stored in a behavior directory (`'beh'`). If eyetracking data is collected alongside another modality such as EEG, it gets written to the `'eeg'` directory.
3) regardless of the directory it is written to (e.g. `'beh'`, `'eeg'`, `'func'`), eyetracking text files are given the `_physio` suffix. When reading, to figure out if a `<match>_physio.tsv` file contains eyetracking data (and not some other physiological data type), you must open its corresponding JSON sidecar file and inspect the `PhsyioType` field.

### Needs

- [ ] #1528
- [ ] #1536


mne_bids/tests/test_copyfiles.py
```diff
@@ -212,6 +212,7 @@ def test_copyfile_mef(tmp_path):
     assert (dest / "session" / "segment.dat").read_bytes() == b"\x00\x01\x02"
 
 
+@pytest.mark.xfail(reason="Need to discuss this with devs.")
```
scott-huberty: per my [comment on _parse_ext](https://github.com/mne-tools/mne-bids/pull/1512#discussion_r2866587657), I refactored `_parse_ext` to handle multi-extension filenames. However this test copies data from `<match>.set` to a file names `<match>.set.set`

However on L232, we seem to test that `copyfile_eeglab(raw_fname, new_name)` will be able to copy `<match>.set` to `<match>.set.set` without error. .

I think this is because on main:

```python
_parse_ext("/CONVERTED_test_raw.set.set") 
```

Will (incorrectly!) return `('/CONVERTED_test_raw.set', '.set')` .. Which just happened to play well with `copyfile_eeglab`.

if I can change one line in this test:

```python
new_name = bids_root / f"CONVERTED_{fname}.set"
```

To

```python
new_name = bids_root / f"CONVERTED_{fname}"
```

Then this test will pass again.

mne_bids/tsv_handler.py
```diff
@@ -183,7 +198,40 @@ def _from_tsv(fname, dtypes=None):
     return data_dict
 
 
-def _to_tsv(data, fname):
+def _from_compressed_tsv(fname, dtypes=None):
+    """Wrap _from_tsv and then read column names from corresponding JSON."""
+    fname = Path(fname)
```
scott-huberty: Physiological (including eyetracking) data files need to be saved into gzipped text files. So I added functionality to be able to read/write such files.

mne_bids/write.py
```diff
@@ -387,7 +388,9 @@ def _events_tsv(
     raw,
     fname,
     trial_type,
+    *,
     event_metadata=None,
+    include_column_names=True,
```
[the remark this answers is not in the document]
scott-huberty: Change of plans. I realized that gzipped text generally files cannot have headers (according to BIDS spec). I added stand alone functions to write compressed TSVs.

mne_bids/write.py
```diff
@@ -2509,7 +2552,8 @@ def write_raw_bids(
 
     # otherwise if the BIDSPath currently exists, check if we
     # would like to overwrite the existing dataset
-    if bids_path.fpath.exists():
+    # TODO: Why did I gatekeep eyetracking from this path. Remove and see if it breaks.
+    if bids_path.fpath.exists() and not is_eyetracking_only:
```
scott-huberty: The reason that I gatekeep this path from eyetracking-only data is because, `bids_path.fpath` will point to the` _physio.tsv.gz` file, which already exists at this point because it gets created earlier in this function (by `write_eyetrack_tsvs`).

Maybe there is a cleaner way to go about this but I will need to think about it.

scott-huberty:
Note to self and reviewers:

This PR is huge. I am going to split out parts of this into standalone PR's, to reduce the diff here and aid focused reviews (e.g.  the compressed TSV I/O code, the `_parse_ext` refactoring code, etc).

PR #1520, opened 2026-02-16
FIX: Detect and autofix missing underscore separators in get_entities_from_fname

### LINKED ISSUE #1518: get_entities_from_fname silently returns mangled values when entity separators are missing
### Describe the problem

When parsing a BIDS filename with a missing underscore separator between entities (e.g., `task-ECONrun-1` instead of `task-ECON_run-01`), `get_entities_from_fname` with `on_error="warn"` returns `task="ECONrun"` and silently drops the `run` entity, without emitting any warning.

The function currently warns about unknown entity keys and out-of-order entities, but doesn't detect when a parsed value appears to contain an embedded entity key.

### Steps to reproduce

```python
from mne_bids import get_entities_from_fname

# Missing underscore between task and run entities
entities = get_entities_from_fname(
    "sub-01_task-ECONrun-1_eeg.set", on_error="warn"
)
print(entities["task"])  # 'ECONrun' — no warning emitted
print(entities["run"])   # None — silently lost

# Compare with correctly formatted filename
entities = get_entities_from_fname(
    "sub-01_task-ECON_run-01_eeg.set", on_error="warn"
)
print(entities["task"])  # 'ECON'
print(entities["run"])   # '01'
```

This pattern was encountered in OpenNeuro datasets during automated processing of ~200 BIDS datasets.

### Describe your solution

When `on_error="warn"`, `get_entities_from_fname` could check whether a parsed entity value contains a substring matching a known BIDS entity key (e.g., a `task` value ending with `run\d+`) and emit a warning like:

> Entity value 'ECONrun' for key 'task' may contain an embedded entity key 'run'. The filename may have a missing underscore separator.

This would help users catch malformed filenames during data validation without changing parsing behavior.

1. **Do nothing** — the filename is non-BIDS-compliant and should be fixed at the source. Users of `on_error="warn"` may not expect the function to validate value contents beyond basic parsing.
2. **Only check for common embedded patterns** (values ending with `run\d+`, `ses\w+`, `acq\w+`) to minimize false positives.
3. **Add a separate validation function** rather than expanding the scope of `get_entities_from_fname`.

### Additional context

- MNE-BIDS version: 0.18.0
- The regex pattern `([^-_\.\\\/]+)-([^-_\.\\\/]+)` correctly captures `task-ECONrun` as a key-value pair — the issue is that the value `ECONrun` silently absorbs what was intended to be a separate `run` entity, with no indication to the user.

bruAristimunha:
## Summary
- Detect filename segments with multiple hyphens (e.g., `task-ECONrun-1`) that indicate missing underscore separators between BIDS entities
- Add `on_error='autofix'` mode that warns and applies unambiguous fixes (e.g., `task-ECON_run-1`)
- Existing modes (`raise`, `warn`, `ignore`) also benefit from improved multi-hyphen detection

Closes #1518

## Test plan
- [x] Parametrized tests for `warn`/`autofix`/`raise`/`ignore` modes across multiple filenames
- [x] Three-entity merge detection (`sub-01task-ECONrun-1`)
- [x] Non-fixable segments preserve best-effort parsing
- [x] Repeated malformed segments produce correct progressive suggestions
- [x] No false positives on valid filenames (including compound extensions like `.nii.gz`, `.tsv.gz`)
- [x] `ruff check` and `ruff format` pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.0
+    rev: v0.15.1
     hooks:
       - id: ruff
         name: ruff mne_bids/


larsoner:
Based on coverage, looks like some of the ambiguous cases are not caught, could you add a `test_..._ambiguous` or so that hits them?

https://app.codecov.io/gh/mne-tools/mne-bids/pull/1520#7602b6afa3fd982e9028ecfc38dd8dd4-R1762

PR #1523, opened 2026-02-27
Fix coordenation

bruAristimunha:
PR Description
--------------

According to BIDS, if it is a derivative, it should include the coordinate system for EEG, besides this:

  - For iEEG: both files are unconditionally REQUIRED -- the current RuntimeError is correct;
  - For EEG: electrodes.tsv is only RECOMMENDED, but if you include it, coordsystem.json MUST accompany it -- our warning is appropriate (the dataset is technically non-compliant but we can still read it)
  - For MEG: coordsystem.json is OPTIONAL at baseline, but the same conditional rule applies -- if electrodes.tsv exists, coordsystem.json MUST exist too

This was overly strict for EEG and MEG, where datasets in the wild often lack coordsystem.json, causing reads to fail unnecessarily.

  This PR differentiates the behavior by modality:
  - iEEG: keeps the strict RuntimeError (coordinates are essential for intracranial data)
  - EEG / MEG: emits a RuntimeWarning referencing the BIDS specification and continues without applying a montage (dig is set to None for EEG, preserved from the raw file for MEG)

  Also fixes the on_error parameter for the sidecar search to use the datatype instead of the suffix, and restructures the logic to remove redundant checks.

  Changes:

  - mne_bids/read.py: use datatype (not suffix) for on_error selection; warn instead of raise for EEG/MEG; clean up with else branch instead of duplicate condition
  - mne_bids/tests/test_read.py: add test_read_eeg_missing_coordsystem_warns, test_read_eeg_root_electrodes_no_coordsystem, test_read_meg_missing_coordsystem_warns; update iEEG test match
  string
  - doc/whats_new.rst: add changelog entry

**Note:** The datasets listed previously are **not** failing due to a regression in mne-bids. I got confused and incorrectly associated the errors with this change. There is no regression — those datasets have other issues unrelated to this PR.

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
           python -m pip install build twine
       - run: python -m build --sdist --wheel
       - run: twine check --strict dist/*
-      - uses: actions/upload-artifact@v6
+      - uses: actions/upload-artifact@v7
         with:
           name: dist
           path: dist
@@ -48,7 +48,7 @@ jobs:
       name: pypi
       url: https://pypi.org/p/mne-bids
     steps:
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@v8
         with:
           name: dist
           path: dist
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -359,7 +359,7 @@ jobs:
       run: |
         make build-doc
     - name: Upload artifacts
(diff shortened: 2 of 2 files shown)


bruAristimunha:
If you want to be really generous, I think we could apply a warning to IEEG as well, but I think that might be too much, because the field is mandatory.

PR #1527, opened 2026-03-03
Allow task=None in read_raw_bids for BIDS paths without task entity

AmanJaiswal1503:
PR Description
--------------

## Summary
Allow `read_raw_bids` to accept BIDSPath instances with `task=None`, supporting BIDS datasets that omit the task entity in the path (e.g. some anatomical or non-task recordings).

## Changes
- **mne_bids/read.py**
  - Require only `root` and `subject` on `bids_path` (removed `task` from required attributes).
  - When deciding whether to ignore missing events, guard `bids_path.task.startswith("rest")` with `bids_path.task is not None` to avoid AttributeError when task is None.

## Datasets this fixes
OpenNeuro datasets that previously failed with `RuntimeError: BIDSPath missing task` and are fixed by this change:
- ds002791
- ds003420
- ds003498
- ds004017

## Minimal self-contained example

This example downloads one subject from the ds002791 OpenNeuro dataset, then reads the data using a `BIDSPath` that does **not** set the `task` entity (only `subject`, `session="spelling"`, `run=1`, and `datatype="eeg"` are set). On **main**, this results in `RuntimeError: "bids_path" must contain 'root', 'subject', and 'task'`; with this PR it runs successfully.

```python
# Minimal example: read without setting task (fails on main, passes with this PR)
import warnings
warnings.filterwarnings("ignore")

import tempfile
from pathlib import Path

import openneuro

# Download one subject from a dataset that can be read without task in the path
dataset = "ds002791"
target = Path(tempfile.mkdtemp(prefix="bids_")) / dataset
openneuro.download(dataset=dataset, target_dir=str(target), include="sub-01")

bids_root = target
bids_path = BIDSPath(
    root=bids_root, subject="01", session="spelling", run=1, task=None, datatype="eeg"
)

raw = read_raw_bids(bids_path, verbose=False)
print(f"Read {raw.filenames} successfully with task=None.")
```

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -1051,6 +1051,14 @@ def read_raw_bids(
                 f"attributes but it's missing `{required}`."
             )
 
+    if bids_path.task is None:
+        warn(
+            "bids_path has no task entity. The BIDS specification requires "
+            "the task entity for M/EEG data; this path may refer to a "
+            "legacy or non-compliant dataset.",
+            UserWarning,
+        )
+
     bids_path = bids_path.copy()
     sub = bids_path.subject
     ses = bids_path.session
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -1164,12 +1164,10 @@ def read_raw_bids(
 
     # Try to find an associated events.tsv to get information about the
     # events in the recorded data
-    if (bids_path.subject == "emptyroom" and bids_path.task == "noise") or (
-        bids_path.task is not None and bids_path.task.startswith("rest")
-    ):
-        on_error = "ignore"
-    else:
-        on_error = "warn"
+    is_emptyroom = bids_path.subject == "emptyroom" and bids_path.task == "noise"
+    is_rest = bids_path.task is not None and bids_path.task.startswith("rest")
+
+    on_error = "ignore" if is_emptyroom or is_rest else "warn"
 
     events_fname = _find_matching_sidecar(
         bids_path, suffix="events", extension=".tsv", on_error=on_error
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -977,11 +977,10 @@ def read_raw_bids(
     ----------
     bids_path : BIDSPath
         The file to read. The :class:`mne_bids.BIDSPath` instance passed here
-        **must** have the ``.root`` and ``.subject`` attributes set. The
-        ``.task`` attribute is optional (e.g. for datasets without a task
-        entity). The ``.datatype`` attribute **may** be set. If ``.datatype``
-        is not set and only one data type (e.g., only EEG or MEG data) is
-        present in the dataset, it will be selected automatically.
+        **must** have the ``.root`` attribute set. The ``.datatype`` attribute
+        **may** be set. If ``.datatype`` is not set and only one data type
+        (e.g., only EEG or MEG data) is present in the dataset, it will be
+        selected automatically.
 
         .. note::
            If ``bids_path`` points to a symbolic link of a ``.fif`` file
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -977,10 +977,11 @@ def read_raw_bids(
     ----------
     bids_path : BIDSPath
         The file to read. The :class:`mne_bids.BIDSPath` instance passed here
-        **must** have the ``.root`` attribute set. The ``.datatype`` attribute
-        **may** be set. If ``.datatype`` is not set and only one data type
-        (e.g., only EEG or MEG data) is present in the dataset, it will be
-        selected automatically.
+        **must** have the ``.root`` and ``.subject`` attributes set. The
+        ``.task`` attribute is optional (e.g. for datasets without a task
+        entity). The ``.datatype`` attribute **may** be set. If ``.datatype``
+        is not set and only one data type (e.g., only EEG or MEG data) is
+        present in the dataset, it will be selected automatically.
 
         .. note::
            If ``bids_path`` points to a symbolic link of a ``.fif`` file
@@ -1043,10 +1044,10 @@ def read_raw_bids(
             '"bids_path" must be a BIDSPath object. Please '
             "instantiate using mne_bids.BIDSPath()."
         )
-    for required in ["root", "subject", "task"]:
+    for required in ["root", "subject"]:
         if not getattr(bids_path, required):
             raise RuntimeError(
-                '"bids_path" must contain `root`, `subject`, and `task` '
+                '"bids_path" must contain `root` and `subject` '
                 f"attributes but it's missing `{required}`."
             )
 
@@ -1156,9 +1157,9 @@ def read_raw_bids(
 
     # Try to find an associated events.tsv to get information about the
     # events in the recorded data
-    if (
-        bids_path.subject == "emptyroom" and bids_path.task == "noise"
-    ) or bids_path.task.startswith("rest"):
+    if (bids_path.subject == "emptyroom" and bids_path.task == "noise") or (
+        bids_path.task is not None and bids_path.task.startswith("rest")
+    ):
         on_error = "ignore"
     else:
         on_error = "warn"
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -215,6 +215,25 @@ def test_read_correct_inputs():
     "ignore:No events found or provided:RuntimeWarning",
     "ignore:Found no extension for raw file.*:RuntimeWarning",
 )
+def test_read_raw_bids_task_none_warns(tmp_path):
+    """Test that read_raw_bids emits UserWarning when bids_path has no task entity."""
+    _write_parallel_dataset(str(tmp_path), subject="01", run="01")
+    bids_path_with_task = BIDSPath(
+        subject="01", task="rest", run="01", datatype="meg", root=tmp_path
+    )
+    bids_path_no_task = BIDSPath(
+        subject="01", task=None, run="01", datatype="meg", root=tmp_path
+    )
+    sh.copy(bids_path_with_task.fpath, bids_path_no_task.fpath)
+    with pytest.warns(UserWarning, match="bids_path has no task entity"):
+        raw = read_raw_bids(bids_path_no_task, verbose=False)
+    assert raw is not None
+
+
+@pytest.mark.filterwarnings(
+    "ignore:No events found or provided:RuntimeWarning",
+    "ignore:Found no extension for raw file.*:RuntimeWarning",
+)
 def test_parallel_participants_multiprocess(tmp_path):
     """Ensure parallel reads keep all participants entries visible."""
     bids_root = tmp_path / "parallel_multiprocess"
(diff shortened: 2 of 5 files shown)


scott-huberty:
Hi @AmanJaiswal1503  would you mind sharing a minimally working example that downloads one subject from one of the aforementioned datasets and tries to read them without setting the task entity? It would be nice to have an example that fails on main but passes here.

AmanJaiswal1503:
Hey @scott-huberty, attaching an image of minimally working example for ds002791 dataset. Please let me know if this is okay, will be happy to add any more information if needed. Thanks.

Screenshot 2026-03-02 at 8 25 11 PM

scott-huberty:
> Hey @scott-huberty, attaching an image of minimally working example for ds002791 dataset. 

@AmanJaiswal1503 can you please share some self-contained code, such that others can copy and paste it in their terminal to run it? You can use the package `openneuro-py` to download an openneuro dataset from Python.

AmanJaiswal1503:
@scott-huberty I have added the self-contained code to the PR description. Hope that helps. Thank you for the suggestion.

hoechenberger:
Wait, since when is `task` not a required entity anymore? Even for MEG empty-room recordings, `task-noise` is mandated.

Can you point to the respective section or change in the specs? Thanks!

hoechenberger:
https://bids-specification.readthedocs.io/en/stable/modality-specific-files/electroencephalography.html#eeg-recording-data

`task` is not marked as optional here…

bruAristimunha:
I have the impression that was the other way around, it was not mandatory in the first place, and few year from now is mandatory. But I couldn't confirm in the documentation because the old version of the specification is not online.

bruAristimunha:
Here we have another problem regarding bid datasets that are on OpenNeuro, but are not real bids.

Any input or maybe some historical background?

larsoner:
@hoechenberger from an office hours chat with @bruAristimunha we were thinking that reading functions should be permissive (probably with a warning) about allowing to read files like this. There are a lot of datasets on OpenNeuro which passed the validator when they were uploaded but would fail now (and/or the validator remains incomplete), it would be a shame not to allow users to read them at all.

On the other hand, at the writing stage, we should enforce the standard to the best of our ability, and not allow people to create new datasets / BIDS files that violate the standard.

WDYT?

hoechenberger:
@larsoner I agree. However when reading, one needs to pass a BIDSPath. How would we handle this? ask users to pass check=False?

larsoner:
In the traceback above, the failure looks like it's in `read_raw_bids` not `BIDSPath` construction. Double checking:
```
>>> from mne_bids import BIDSPath, read_raw_bids
>>> path = BIDSPath(root=".", subject="01", session="a", task=None, run=None)
>>> read_raw_bids(path)
  File "<python-input-3>", line 1, in <module>
    read_raw_bids(path)
    ~~~~~~~~~~~~~^^^^^^
  File "<decorator-gen-402>", line 12, in read_raw_bids
  File "/Users/larsoner/python/mne-bids/mne_bids/read.py", line 1048, in read_raw_bids
    raise RuntimeError(
    ...<2 lines>...
    )
RuntimeError: "bids_path" must contain `root`, `subject`, and `task` attributes but it's missing `task`.
```
So no need to net `check=False`

AmanJaiswal1503:
Hi @larsoner , @hoechenberger and @bruAristimunha  — thanks for the discussion on making the reader permissive.

To align with the “permissive with a warning” idea, I’d be happy to add a warning in `read_raw_bids` when `task` is `None`, e.g. that the path has no task entity and the dataset may not be fully BIDS compliant. That would make it clear we’re allowing reads for legacy/incomplete datasets without implying that omitting task is correct for new data.

If you’re on board with that, I can add the warning in the next commit.

drammock:
@larsoner this issue is different than the one we discussed with @bruAristimunha in office hours.  That was about absence of `coordsystem.json` files; we agreed that letting folks *read* the ephys data (with a warning) even when coordsystem file was missing was a good idea (and seems @hoechenberger agrees at least in principle).

This issue is about `task` entity, which AFAIK has always been mandatory for MEG and EEG, and was made mandatory for EMG when that modality was recently added to BIDS.  There has been talk about making it optional (e.g., [here](https://github.com/bids-standard/bids-specification/pull/1998#issuecomment-3335575946)) but I don't think it's happened yet.

drammock:
Quoting from the MWE:

> This example downloads one subject from the ds002791 OpenNeuro dataset, then reads the data using a `BIDSPath` that does **not** set the `task` entity (only `subject`, `session="spelling"`, `run=1`, and `datatype="eeg"` are set).

Without scrutinizing the dataset too closely, I'd wager that `session="spelling"` really should have been `task="spelling"`, and the dataset ought to be corrected.

That said, I think I would argue that (as in the coordsystem case) we should probably let users read datasets that are not 100% valid BIDS (but raise warnings when we do). In this case, if specifying a BIDSPath with root, subj, and session is enough to unambiguously select a desired file, we shouldn't stand on ceremony to demand a `task` entity that wasn't used in the dataset (even if it should have been).

hoechenberger:
Oh my 😃
Whatever helps our users read more real-world datasets, take my +1 😃
We could also be strict by default but issue a helpful message and instruct users to pass a certain parameter (like, check=False) explicitly

hoechenberger:
@drammock 
> That said, I think I would argue that (as in the coordsystem case) we should probably let users read datasets that are not 100% valid BIDS (but raise warnings when we do).

I'd be on board with that!

scott-huberty:
I agree that it seems the task entity has always been mandatory (even as far back as [v1.1.2](https://bids-specification.readthedocs.io/en/v1.1.2/04-modality-specific-files/02-magnetoencephalography.html)). It seems like a gap in the `bids-validator` coverage is what allowed these aforementioned datasets to pass validation on openneuro:

e.g. The legacy validator won't complain about the missing `task` entity:

`npx bids-validator@latest /path/to/ds002791`

But the new and improved schema validator will raise an error:

`deno -A jsr:@bids/validator /path/to/ds002791`


I imagine that many users likely depend on and trust the validator to tell them if their datasets are compliant. At the time of upload, these datasets all passed validation. So +1 from me to gracefully handle this scenario (when possible)!

AmanJaiswal1503:
Hey @scott-huberty @hoechenberger @larsoner and @bruAristimunha,

To align with the “permissive with a warning” approach we discussed, I’ve made two updates:

**Warning in read_raw_bids**
When bids_path.task is None, we now emit a UserWarning that the path has no task entity, that the BIDS spec requires the task entity for M/EEG data, and that the dataset may be legacy or non-compliant. Reading still proceeds so users can load these datasets.

**Test configuration**
The test suite uses filterwarnings = ["error", ...], so the new warning was causing a failure. I added an ignore for this specific UserWarning in pyproject.toml so existing tests that don’t set a task (or any new tests that use task=None) don’t fail, while the warning still appears for users in normal use.

Please let me know if you’d prefer the warning worded differently or if you want a dedicated test that explicitly checks for this warning (e.g. with pytest.warns).

AmanJaiswal1503:
I have also added tests!

mne_bids/tests/test_read.py
```diff
@@ -211,6 +211,26 @@ def test_read_correct_inputs():
...
+        subject="01",
+        run="01",
+        datatype="meg",
+        suffix="meg",
+        extension=".fif",
+        root=tmp_path,
+    )
+    bids_path.fpath.parent.mkdir(parents=True, exist_ok=True)
+    raw = _make_parallel_raw("01")
+    raw.save(bids_path.fpath)
```
larsoner: This is the part that we shouldn't allow. You shouldn't be allowed to write an invalid file, just read it.

So ideally the code here would be in a `with pytest.raises(...)`, then you change the path to be valid, then write, then rename/adjust files on disk as if they had been written with `task=None`, then read (which should warn) as below.

mne_bids/read.py
```diff
@@ -977,10 +977,11 @@ def read_raw_bids(
     ----------
     bids_path : BIDSPath
         The file to read. The :class:`mne_bids.BIDSPath` instance passed here
-        **must** have the ``.root`` attribute set. The ``.datatype`` attribute
-        **may** be set. If ``.datatype`` is not set and only one data type
-        (e.g., only EEG or MEG data) is present in the dataset, it will be
-        selected automatically.
+        **must** have the ``.root`` and ``.subject`` attributes set. The
+        ``.task`` attribute is optional (e.g. for datasets without a task
```
larsoner: Hmm... maybe we should revert this part. We don't really want to advertise that `task` is optional, even if it is.

pyproject.toml
```diff
@@ -121,6 +121,8 @@ filterwarnings = [
   # numba with NumPy dev
   "ignore:`np.MachAr` is deprecated.*:DeprecationWarning",
   "ignore:`product` is deprecated as of NumPy.*:DeprecationWarning",
+  # read_raw_bids with task=None (legacy / non-compliant BIDS)
+  "ignore:bids_path has no task entity.*:UserWarning",
```
larsoner: -1 for adding it here

I'd rather add any `with pytest.warns(...)` anywhere we need them in tests (which is hopefully nowhere other than in the test added here!)

AmanJaiswal1503:
Hey @larsoner! Thank you for the suggestions, I have incorporated all of them in my latest commit.

bruAristimunha (approved):

bruAristimunha:
For me, it is good! no point here, what about you @scott-huberty, @sappelhoff, @drammock?

larsoner (approved):
LGTM, I'll let @scott-huberty or @drammock merge if they're happy

scott-huberty:
> LGTM, I'll let @scott-huberty or @drammock merge if they're happy

I don't have time to review today but If no one else gets to it before then I will look at this tomorrow afternoon!

mne_bids/read.py
```diff
@@ -1156,9 +1164,9 @@ def read_raw_bids(
...
     # events in the recorded data
-    if (
-        bids_path.subject == "emptyroom" and bids_path.task == "noise"
-    ) or bids_path.task.startswith("rest"):
+    if (bids_path.subject == "emptyroom" and bids_path.task == "noise") or (
+        bids_path.task is not None and bids_path.task.startswith("rest")
+    ):
         on_error = "ignore"
     else:
         on_error = "warn"
```
```suggestion
    is_emptyroom = bids_path.subject == "emptyroom" and bids_path.task == "noise"
    is_rest = bids_path.task is not None and bids_path.task.startswith("rest")

    on_error = "ignore" if is_emptyroom or is_rest else "warn"
```

scott-huberty: The syntax was correct but honestly for me that condition was becoming difficult to mentally unpack. Are folks ok with splitting this up like so and giving each of those concepts names?

scott-huberty (approved):
Just 1 nitpick. The test is nicely done. Thanks @AmanJaiswal1503 !

mne_bids/read.py
AmanJaiswal1503: yes, it is easier to comprehend this way

scott-huberty: committed.

PR #1528, opened 2026-03-03
ENH: read/write gzipped tsv

scott-huberty:
In order to support physiological data, MNE-BIDS needs to be able to read and write gzipped TSVs, which are headerless and must store their column names in a corresponding JSON sidecar file.

## Needs

- [x] #1531


mne_bids/tsv_handler.py
```diff
@@ -193,15 +235,23 @@ def _to_tsv(data, fname):
     fname : str
         Path to the file being written.
     """
+    include_header = False if compress else True
+    encoding = "utf-8-sig"
```
hoechenberger, line 239: This adds a behavior change, no? Personally I don't want BOM headers in my TSVs. What does the BIDS spec say? If it merely mentions UTF8 and doesn't say anything about a BOM, we should not add one.

scott-huberty: @hoechenberger On `main` TSV files are written with this encoding as well:

https://github.com/mne-tools/mne-bids/blob/bbc83e7b074561598e36f7c3311013cdbee69044/mne_bids/tsv_handler.py#L186-L199

I just took a quick look at the BIDS spec for [tabular files.](https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files) and it just says UTF8. So probably they don't intend to require writing BOM?

It looks like writing `utf-8-sig` in all TSV files started in #625 - specifically see https://github.com/mne-tools/mne-bids/pull/625#issuecomment-726902640

hoechenberger: Haha funny, so I initially didn't want to have a BOM but then we decided otherwise :)

I did a bit of digging online and it seems that NOT having the BOM is the "correct" way of doing things. And newer versions of Excel (at least in M365) read non-BOM'd TSV files correctly! In fact, a quick test shows actually that the situation has sort of **reversed!** – not having a BOM is BETTER for Excel, at least when reading TSVs, as can be seen here:

```python
# test.py
#
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "pandas>=3.0.1",
# ]
# ///
import pandas as pd

data = pd.DataFrame.from_dict({
    "Foo": [1, 2, 3],
    "Bar": [4, 5, 6],
    "Baz": [7, 8, 9]
})

data.to_csv("data.tsv", index=False, encoding="utf-8", sep="\t")
data.to_csv("data-bom.tsv", index=False, encoding="utf-8-sig", sep="\t")
```
Run via:
```
uv run test.py
```

No BOM – loads correctly upon a double-click on the file:
image


With BOM – does not load correctly:
image

So I would say we drop the BOM. How to handle this potentially breaking change (then again, who would really be affected?) – I leave up to @sappelhoff to decide 😃

scott-huberty: > Haha funny, so I initially didn't want to have a BOM but then we decided otherwise :)

Well at least you have been consistent 🙂 So we would switch to `"utf-8"` for writing but keep `"utf-8-sig"` when reading for backwards compatibility? I agree with you, and I think writing the BOM actually might make it such that other BIDS apps have trouble ingesting datasets written by MNE-BIDS! e.g. see https://github.com/sccn/EEG-BIDS/issues/78 !

I am going to elevate your comment into a separate issue so that we can discuss there!

sappelhoff: very interesting. 

> So we would switch to "utf-8" for writing but keep "utf-8-sig" when reading for backwards compatibility?

sounds reasonable to me!

scott-huberty:
@bruAristimunha and @larsoner finally ready for review. can you take a look at this next week when you have some time?

scott-huberty:
I will self-merge this today so that we can continue to make progress on Eyetracking BIDS support!

mne_bids/tsv_handler.py
```diff
@@ -227,7 +247,24 @@ def _from_tsv(fname, dtypes=None):
     return data_dict
 
 
-def _to_tsv(data, fname, *, lock=True):
+def _get_column_names_from_json(fname, dtypes=None):
+    """Wrap _from_tsv and then read column names from corresponding JSON."""
```
scott-huberty:
```suggestion
    """Read column names for gzipped TSV files from corresponding JSON."""
```

PR #1531, opened 2026-03-04
FIX: Write text files with utf-8 encoding instead of utf-8-sig

### LINKED ISSUE #1530: MNE-BIDS should not include a BOM when writing TSV files
From #1528 .. in #625 MNE-BIDS started writing text files with encoding `"utf-8-sig"`. However the BIDS spec only requires `'utf-8'` for tabular files:

Haha funny, so I initially didn't want to have a BOM but then we decided otherwise :)

I did a bit of digging online and it seems that NOT having the BOM is the "correct" way of doing things. And newer versions of Excel (at least in M365) read non-BOM'd TSV files correctly! In fact, a quick test shows actually that the situation has sort of **reversed!** – not having a BOM is BETTER for Excel, at least when reading TSVs, as can be seen here:

```python
# test.py
#
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "pandas>=3.0.1",
# ]
# ///
import pandas as pd

data = pd.DataFrame.from_dict({
    "Foo": [1, 2, 3],
    "Bar": [4, 5, 6],
    "Baz": [7, 8, 9]
})

data.to_csv("data.tsv", index=False, encoding="utf-8", sep="\t")
data.to_csv("data-bom.tsv", index=False, encoding="utf-8-sig", sep="\t")
```
Run via:
```
uv run test.py
```

No BOM – loads correctly upon a double-click on the file:
image

With BOM – does not load correctly:
image

So I would say we drop the BOM. How to handle this potentially breaking change (then again, who would really be affected?) – I leave up to @sappelhoff to decide 😃

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-bids/pull/1528#discussion_r2879179427_

scott-huberty:
Fixes #1530 cc @sappelhoff @hoechenberger 

~~Hope you don't mind that I added 2 new constants to `config.py`. I think this makes the code intent clearer and will make updating these encodings easier in the future.~~


hoechenberger (commented):
Hi, great so far!

Except for JSON reading: we should NOT support reading JSON with a BOM, as this simply isn't valid JSON. We should fail hard in this case.

As for the changelog update: I would think that the only relevant change for users is TSV writing. I would be specific about this and omit that other text files are affected as well. But this is just my personal view

scott-huberty:
> Except for JSON reading: we should NOT support reading JSON with a BOM, as this simply isn't valid JSON. We should fail hard in this case.

OK! addressed in [a2f71e2](https://github.com/mne-tools/mne-bids/pull/1531/commits/a2f71e2c77762068b79fe4751a2658a3c804bb1b) . I expanded my two encoding constants into a little class, now that we have different encoding rules for TSV vs JSON I/O. For me this feels like a clean approach but If folks think it is overkill feel free to let me know.

mne_bids/dig.py
```diff
@@ -178,7 +178,7 @@ def _handle_coordsystem_reading(coordsystem_fpath, datatype):
     Handle reading the coordinate frame and coordinate unit
     of each electrode.
     """
-    with _open_lock(coordsystem_fpath, encoding="utf-8-sig") as fin:
+    with _open_lock(coordsystem_fpath, encoding="utf-8") as fin:
```
scott-huberty, line 181: @hoechenberger and I discussed always reading JSON files with utf-8 encoding (because `'utf-8-sig'` is not permitted for JSON per the JSON spec). 

I could not find any evidence that MNE-BIDS ever wrote JSON with `utf-8-sig`, but the fact that we previously used `"utf-8-sig"` to read JSON files here makes me a little nervous that maybe at some point we wrote this coordsystem.json as utf-8-sig

mne_bids/path.py
```diff
@@ -1493,7 +1493,8 @@ def _print_lines_with_entry(file, entry, folder, is_tsv, line_numbers, outfile):
         prints to the console, else a string is printed to.
     """
     entry_lines = list()
-    with _open_lock(file, encoding="utf-8-sig") as fid:
+    encoding = _detect_file_encoding(file)
+    with _open_lock(file, encoding=encoding) as fid:
```
scott-huberty, line 1497: It appears that this line was missed by #1593 , maybe because that PR was strictly for TSV files and this function may be used on more file types (as suggested by the `is_tsv` param).

mne_bids/write.py
```diff
@@ -608,7 +614,8 @@ def _readme(datatype, fname):
     # observe a partially written file.
     with _open_lock(fname):
         if fname.is_file():
-            text = fname.read_text("utf-8-sig")
+            encoding = _detect_file_encoding(fname)
+            text = fname.read_text(encoding)
```
scott-huberty, line 618: Doesn't hurt to use `_detect_file_encoding` on README.{txt,md} files right?

drammock: should be fine

larsoner (approved):
Looks good to me, if nobody else asks to look @scott-huberty feel free to self-merge in a couple of days!

bruAristimunha (approved):
LGTM

PR #1532, opened 2026-03-09
FIX: Warn instead of crash on malformed scans.tsv timestamps

bruAristimunha:
## Summary

- `_handle_scans_reading` in `read.py` calls `datetime.strptime(acq_time, date_format)` without error handling. Malformed timestamps (seconds >= 60, truncated strings, non-ISO formats) cause an unhandled `ValueError` that crashes `read_raw_bids()`.
- Since mne_bids already handles `"n/a"` gracefully (skipping `meas_date` assignment), a malformed timestamp should degrade the same way rather than crash.
- Wraps `datetime.strptime` in `try/except ValueError`, emits a warning, and returns `raw` with its measurement date unchanged.
- Discovered processing OpenNeuro datasets during batch ingestion with [eegdash](https://github.com/bruAristimunha/eegdash).

## Test plan

- [ ] Added test cases with `"not-a-date"`, `"2020-01-01T25:61:61"`, and `"2020-13-01T00:00:00"` timestamps
- [ ] Verify warning is emitted and `meas_date` is unchanged
- [ ] Run `make pep` to confirm style compliance
- [ ] Run `make test` for full test suite


bruAristimunha:
sorry folks, i will create the issue first.

PR #1534, opened 2026-03-10
Add HED read support from events.json sidecar

bruAristimunha:
## Summary

- Read HED tags from JSON sidecar annotations (`"trial_type": {"HED": {"show_face": "Sensory-event"}}`) and/or direct HED column in `events.tsv`
- Create `HEDAnnotations` (from MNE-Python) when available, with graceful fallback for older MNE versions
- Add `_assemble_hed_from_sidecar()` to resolve categorical and template HED specs per row
- Add `_read_hed_version()` to read `HEDVersion` from `dataset_description.json`
- Add `_safe_iadd_annotations()` to handle merging `HEDAnnotations` with regular `Annotations`
- Update tiny_bids fixture with HED sidecar data and HEDVersion

## Test plan

- [x] `test_events_file_to_annotation_kwargs` — HED column separated from extras
- [x] `test_handle_events_reading_hed` — column HED, default version fallback, n/a fallback
- [x] `test_assemble_hed_from_sidecar` — categorical + template HED assembly
- [x] `test_handle_events_reading_sidecar_hed` — sidecar HED → HEDAnnotations


mne_bids/tests/test_read.py
```diff
@@ -2036,3 +2036,103 @@ def test_events_file_to_annotation_kwargs(tmp_path):
...
+    ]
+    if ev_kwargs_hed["extras"] is not None:
+        for extra in ev_kwargs_hed["extras"]:
+            assert "HED" not in extra
+
+
+def test_handle_events_reading_hed(tmp_path):
+    """Test HEDAnnotations from column HED, sidecar HED, and fallbacks."""
+    if not hasattr(mne, "HEDAnnotations"):
+        pytest.skip("HEDAnnotations requires a recent MNE-Python version")
```
drammock:
```suggestion
        pytest.skip("HEDAnnotations requires MNE-Python 1.12 or newer")
```

mne_bids/read.py
```diff
@@ -118,6 +120,23 @@ def _read_raw(
...
+        annotations += new_annotations
+    except TypeError:
+        # HEDAnnotations can't merge with regular Annotations.
+        # Convert to regular Annotations to allow merge.
+        annotations = mne.Annotations(
+            onset=annotations.onset,
+            duration=annotations.duration,
+            description=annotations.description,
+            orig_time=annotations.orig_time,
+        )
```
drammock: shouldn't you convert the HED information into `extras` here, so the info isn't lost?

bruAristimunha: true!

mne_bids/read.py
```diff
@@ -118,6 +120,23 @@ def _read_raw(
     return raw
 
 
+def _safe_iadd_annotations(annotations, new_annotations):
```
drammock: also, wondering if it would be better to add this functionality to MNE-Python instead (e.g., add code in the `__iadd__` of plain Annotations to handle the special case of trying to add HEDAnnotations to them)?

bruAristimunha:
One dataset showing working:

You will need to install this branch:

```shell
pip install git+https://github.com/bruAristimunha/mne-bids.git@feat/hed-read-support
```

```python
import openneuro                                                                                                                                                                           
from mne_bids import read_raw_bids, BIDSPath
                                                                                                                                                                                           
# Download one subject (~120MB) from ds002893 (Auditory-Visual Attention Shift)                                                                                                          
openneuro.download(                                                                                                                                                                      
    dataset="ds002893",
    target_dir="/tmp/ds002893",
    include=["dataset_description.json", "participants.tsv",
             "participants.json", "task-*_events.json", "sub-001/**"],
    verify_hash=False,
)

# Read with MNE-BIDS — HED is assembled from the JSON sidecar automatically
bids_path = BIDSPath(
    subject="001", task="AuditoryVisualShift", run="01",
    datatype="eeg", root="/tmp/ds002893",
)
raw = read_raw_bids(bids_path, verbose="warning")

# HEDAnnotations created successfully
print(f"Type: {type(raw.annotations).__name__}")  # HEDAnnotations
print(f"Events: {len(raw.annotations)}")           # 5856
print(f"HED version: {raw.annotations._hed_version}")  # 8.0.0

# Each annotation has a full HED string assembled from 7 sidecar columns
for desc, hed in zip(raw.annotations.description[:3],
                     raw.annotations.hed_string[:3]):
    print(f"\n  {desc}:\n    {hed}")
```

mne_bids/read.py
bruAristimunha: fixed

mne_bids/read.py
bruAristimunha: It is a very short function; you decide what is best here. If you think this solves more at the root, I'm happy to open the PR.

drammock: my feeling is: people might want to do this with data that isn't BIDS-formatted, so why not support that too?  There's nothing BIDS-specific about wanting to combine HED with non-HED annotations.

bruAristimunha: Issue created @drammock, I will create the PR at MNE-Python: https://github.com/mne-tools/mne-python/issues/13735

bruAristimunha:
it is necessary to merge first the mne-tools/mne-python#13736

bruAristimunha:
hey @drammock,

I was wondering, gentil ping here

bruAristimunha:
hey @larsoner, can we merge this please?

mne_bids/read.py
```diff
@@ -744,44 +768,138 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+
+def _assemble_hed_from_sidecar(events_dict, events_json_fname):
+    """Assemble per-row HED strings from column annotations in events.json."""
+    if events_json_fname is None:
+        return None
+
+    try:
+        with open(events_json_fname, encoding="utf-8-sig") as f:
+            sidecar = json.load(f)
+    except (json.JSONDecodeError, OSError):
```
larsoner: Seems like we should at least emit a warning if the file exists but cannot be read. If that's already done elsewhere, can you add a comment saying so?

mne_bids/read.py
```diff
@@ -744,44 +768,138 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+        try:
+            annot_from_events = mne.HEDAnnotations(
+                onset=annotations_info["onset"],
+                duration=annotations_info["duration"],
+                description=annotations_info["description"],
+                hed_string=hed_strings,
+                hed_version=_read_hed_version(bids_root) or _DEFAULT_HED_VERSION,
+                extras=extras,
             )
+        except (AttributeError, ValueError, ImportError):
```
larsoner: Seems like we should emit a warning saying HED was detected but not able to be parsed

mne_bids/read.py
bruAristimunha: Fixed in f63a93ff. `_assemble_hed_from_sidecar` now emits a `RuntimeWarning` when the sidecar exists but cannot be decoded (JSONDecodeError / OSError). `FileNotFoundError` stays silent since a missing sidecar is expected.

mne_bids/read.py
bruAristimunha: Fixed in f63a93ff. `_handle_events_reading` now emits a `RuntimeWarning` ("HED annotations were detected but could not be parsed") before falling back to regular `Annotations`. HED strings remain preserved in `extras`.

bruAristimunha:
@larsoner your turn now :)

drammock:
lots of failing CIs @bruAristimunha

mne_bids/read.py
```diff
@@ -43,6 +50,8 @@
     warn,
 )
 
+_DEFAULT_HED_VERSION = "8.3.0"
```
scott-huberty: Should this live in `mne_bids.config`?

mne_bids/tests/test_read.py
```diff
@@ -2095,3 +2099,233 @@ def test_events_file_to_annotation_kwargs(tmp_path):
...
+    return bids_root, events_tsv, events_json, raw
+
+
+def test_handle_events_reading_hed(_hed_tiny_bids):
+    """Test HEDAnnotations from column HED, sidecar HED, and fallbacks."""
+    pytest.importorskip("hed")
+    try:
+        from mne.annotations import HEDAnnotations
+    except ImportError:
+        pytest.skip("HEDAnnotations not available in this MNE version")
```
scott-huberty: Would `pytest.importorskip("mne", minversion=min_supported_version)` work?

mne_bids/read.py
scott-huberty: Should this be moved to `mne_bids.config`?

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+                continue
+
+            if isinstance(hed_spec, dict):
+                # Categorical: {"famous_face": "Def/Famous-face-cond", ...}
+                tag = hed_spec.get(val_str)
+                if tag:
+                    per_row_tags[i].append(tag)
+            elif isinstance(hed_spec, str) and "#" in hed_spec:
+                # Value template: "Item-interval/#"
+                per_row_tags[i].append(hed_spec.replace("#", val_str))
```
scott-huberty, line 827: Do we need an `else` here?

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+                warn(
+                    "The version of MNE-Python you are using (<1.10) does "
+                    "not support the extras argument in mne.Annotations. "
+                    f"The extra column(s) {list(extras[0].keys())} will be "
+                    "ignored."
+                )
+            annot_from_events = mne.Annotations(
+                onset=annotations_info["onset"],
+                duration=annotations_info["duration"],
+                description=annotations_info["description"],
```
```suggestion
        kwargs = dict(
            onset=annotations_info["onset"],
            duration=annotations_info["duration"],
            description=annotations_info["description"],
        )
        try:
            annot_from_events = mne.Annotations(**kwargs, extras=extras)
        except TypeError:
            if extras:
                warn(
                    "The version of MNE-Python you are using (<1.10) does "
                    "not support the extras argument in mne.Annotations. "
                    f"The extra column(s) {list(extras[0].keys())} will be "
                    "ignored."
                )
            annot_from_events = mne.Annotations(**kwargs)
```

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
             )
```
scott-huberty, line 929:
```suggestion
```

mne_bids/read.py
```diff
@@ -17,6 +17,13 @@
 from mne.transforms import apply_trans
 from mne.utils import get_subjects_dir, logger
 
+try:
+    # HEDAnnotations was added in MNE 1.12; importing from mne.annotations
+    # works regardless of whether the top-level lazy-loader stub re-exports it.
+    from mne.annotations import HEDAnnotations as _HEDAnnotations
+except ImportError:
+    _HEDAnnotations = None
+
```
```suggestion
```

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+    # Try to create HEDAnnotations when all HED strings are valid.
+    # Skip silently when MNE is too old to expose HEDAnnotations — that is a
+    # feature-availability fallback, not a parse failure (HED is still kept in
+    # extras). Only warn when construction is actually attempted and fails.
+    annot_from_events = None
+    if (
+        hed_strings is not None
+        and all(s and s != "n/a" for s in hed_strings)
+        and _HEDAnnotations is not None
```
```suggestion
```

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+        try:
+            annot_from_events = _HEDAnnotations(
+                onset=annotations_info["onset"],
+                duration=annotations_info["duration"],
+                description=annotations_info["description"],
+                hed_string=hed_strings,
+                hed_version=_read_hed_version(bids_root) or _DEFAULT_HED_VERSION,
+                extras=extras,
+            )
+        except (AttributeError, ValueError, ImportError) as exc:
```
scott-huberty: If you use my pattern above, I think this `ImportError` can be removed from this tuple?

mne_bids/read.py
scott-huberty: more DRY?

mne_bids/read.py
```diff
@@ -744,45 +775,178 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+        )
+        return None
+
+
+def _handle_events_reading(
+    events_fname, raw, *, bids_root=None, events_json_fname=None
+):
     """Read associated events.tsv and convert valid events to annotations on Raw."""
-    annotations_info = events_file_to_annotation_kwargs(events_fname)
+    events_dict = _from_tsv(events_fname)
```
```suggestion
    mne_supports_hed = mne.utils.check_version(mne, min_version="1.12")
    events_dict = _from_tsv(events_fname)
```

mne_bids/read.py
```suggestion
        and mne_supports_hed
```

scott-huberty (commented):
First pass, let me know if I misunderstood anything, @bruAristimunha !

mne_bids/read.py
bruAristimunha: Moved to `mne_bids.config` as `DEFAULT_HED_VERSION` in 77c6efb4, alongside the other version constants.

bruAristimunha: Done — moved to `mne_bids.config.DEFAULT_HED_VERSION` in 77c6efb4.

mne_bids/tests/test_read.py
bruAristimunha: Good call — switched to `pytest.importorskip('mne', minversion='1.12')` in 77c6efb4 (both HED tests).

mne_bids/read.py
bruAristimunha: Applied in 77c6efb4 — shared `kwargs = dict(onset=..., duration=..., description=...)` used by both the HEDAnnotations call and the plain Annotations fallback.

mne_bids/read.py
bruAristimunha: Dropped the module-level `try: from mne.annotations import HEDAnnotations / except ImportError` in 77c6efb4 in favor of a runtime `mne.utils.check_version('mne', min_version='1.12')` gate.

mne_bids/read.py
bruAristimunha: Dropped `AttributeError` (version gate eliminates it) and kept `(ValueError, ImportError)` — `ImportError` still fires when `hedtools` is missing inside `HEDAnnotations.__init__`. Done in 77c6efb4.

mne_bids/read.py
bruAristimunha: Added `mne_supports_hed = check_version('mne', min_version='1.12')` in 77c6efb4 (passing the string `'mne'`; passing the module object tripped MNE's lazy-loader on `startswith`).

mne_bids/read.py
bruAristimunha: Replaced the `_HEDAnnotations is not None` guard with `mne_supports_hed` in 77c6efb4.

mne_bids/read.py
bruAristimunha: No `else` needed here — the BIDS spec only defines column-level HED as a dict (categorical map) or a string containing `#` (value template). Anything else (e.g. a plain string) is not a documented shape, so silently skipping matches the BIDS-compliant behavior. Happy to add an `else: pass  # BIDS does not define` comment if that would be clearer.

mne_bids/read.py
bruAristimunha: On reflection, reverted in 95fd798f: `_DEFAULT_HED_VERSION` is a default data-schema fallback used by one function, not a dependency/spec pin like `BIDS_VERSION` / `PYBV_VERSION`. Grouping them by name suffix only is superficial, and the leading underscore signalled module-private. Kept it at the top of `read.py`. Happy to reconsider if there's a second consumer in flight.

bruAristimunha: Same as above — reverted in 95fd798f. Reasoning: the other `*_VERSION` constants in config are package pins; `_DEFAULT_HED_VERSION` is a module-private fallback for a data schema, so it stays with its single consumer.

bruAristimunha: Actually better: dropped the constant entirely in f11ec12d. `mne.HEDAnnotations.__init__` already defaults `hed_version='8.3.0'` (mne/annotations.py:1054), so duplicating it here adds nothing. We now only pass `hed_version=` when the dataset specifies one; otherwise we omit the kwarg and let MNE's default apply.

mne_bids/read.py
bruAristimunha: You were right — my earlier justification was wrong. MNE's `_soft_import` actually raises `RuntimeError` (not `ImportError`), so the original tuple missed the missing-hedtools case. Fixed in fc8db7eb: now `except (RuntimeError, ValueError)`, with a comment noting which MNE call raises which, plus a parametrized test covering both failure modes.

mne_bids/read.py
bruAristimunha: Added a one-line comment in fc8db7eb: `# Other shapes are not defined by the BIDS HED spec; silently skip.` Happy to switch to an explicit `else: warn(...)` if you'd prefer louder feedback for malformed sidecars.

bruAristimunha:
should be better @scott-huberty :)

mne_bids/read.py
```diff
@@ -744,45 +766,188 @@ def events_file_to_annotation_kwargs(events_fname: str | Path, *, verbose=None)
...
+        duration=annotations_info["duration"],
+        description=annotations_info["description"],
+    )
+
+    # HED path: only when MNE >=1.12 and every row has a usable HED string.
+    if (
+        hed_strings is not None
+        and all(s and s != "n/a" for s in hed_strings)
+        and check_version("mne", min_version="1.12")
+    ):
```
scott-huberty, line 902: @bruAristimunha do you think that we should warn the user if their raw object has HED annotations but we can't save them using the new and improved HED path?

scott-huberty (approved):
Still hitting a CI failure due to OpenNeuro

mne_bids/read.py
bruAristimunha: solved now @scott-huberty :)

scott-huberty:
@larsoner and/or @drammock CI's are green, do either of you want to look at this again?

larsoner:
Nope feel free to merge if you're happy @scott-huberty !

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_events.json
```diff
@@ -6,6 +6,10 @@
         "Description": "The event code (also known as trigger code or event ID) associated with the event."
     },
     "trial_type": {
-        "Description": "The type, category, or name of the event."
+        "Description": "The type, category, or name of the event.",
+        "HED": {
+            "start_experiment": "Experiment-structure",
+            "show_stimulus": "Sensory-event, Visual-presentation"
+        }
```
drammock, line 13: I may be naive about how HED+BIDS is supposed to work, but...  does this even make sense?  We're in a JSON sidecar which is supposed to just define the meanings / units of the columns in events.tsv.  Why are there nested keys here, and why do the values look like experimental data... and why no corresponding columns in the `.tsv` file?

drammock (changes requested):

mne_bids/tests/data/tiny_bids/sub-01/ses-eeg/eeg/sub-01_ses-eeg_task-rest_events.json
drammock: never mind. I read up on it. Sorry for the noise.

drammock (approved):

bruAristimunha:
can we merge @scott-huberty, @larsoner and @scott-huberty ? I would like to move on to the writing part, and I need this one to be merged.

larsoner:
Yeah looks good, merging `main` into this branch and marking for merge-when-green, thanks in advance @bruAristimunha !

PR #1535, opened 2026-03-10
Add HED write support to events.json sidecar

bruAristimunha:
## Summary

- Extract HED tags from `HEDAnnotations` and write as sidecar mapping in `events.json` (`"trial_type": {"HED": {...}}`)
- Falls back to HED column in `events.tsv` when same trial_type has different HED strings
- Set `HEDVersion` in `dataset_description.json`
- Fix `make_dataset_description()` overwrite logic to allow adding new fields without overwriting existing ones
- Fix `_events_json` docstring

**Depends on #1534**

## Test plan

- [x] `test_write_hed_annotations` — write + roundtrip: verifies sidecar HED in events.json, HEDVersion in dataset_description.json, no HED column in events.tsv, HEDAnnotations reconstructed on read


bruAristimunha:
suuuper clean PR now :) @drammock, @scott-huberty, @larsoner

mne_bids/write.py
```diff
@@ -481,8 +481,47 @@ def _events_tsv(
...
+        return None
+    strings = list(raw.annotations.hed_string)
+    if len(strings) != n_events:
+        warn(
+            f"Number of HED strings ({len(strings)}) does not match number "
+            f"of events ({n_events}). HED data will not be written."
+        )
+        return None
+    sidecar_map = {}
+    for desc, s in zip(raw.annotations.description, strings):
```
scott-huberty: Is it predetermined that `len(strings) == len(raw.annotations.description)` ? 

I just worry about the silent bugs that would be caused by a length mismatch in these iterables

scott-huberty (approved):
Just one question! 

I did not checkout this branch and test it locally, nor am I very familiar with BIDS HED. 
Maybe @larsoner or @drammock or @sappelhoff can take a look before merge.

mne_bids/write.py
bruAristimunha: Good catch. `HEDAnnotations.__init__` calls `_check_hed_strings(hed_string, len(onset))` (mne/annotations.py:1064) and all mutation paths preserve it, so `len(hed_string) == len(description)` holds. The zip would still silently truncate if that ever broke, though. Added `strict=True` in 7e9b4bc so it fails loud instead.

scott-huberty: Awesome! Great idea to use  `strict=True`, gives us a sanity check without creating a separate path that we have to test ourselves ; )

scott-huberty:
@bruAristimunha I think the one error was just due to a connection timeout.

Also - Do you think it is worth using the `bids_validate` fixture in of these write tests? The CI that uses the Schema validator should be able to handle it (not sure about the legacy validators). But we should be able to triage the validation on validator type, as done in #1540

bruAristimunha:
excellent idea @scott-huberty! i increased the test to validate this :) And I put one extra line to be more semantically correct test line with hedtools too

scott-huberty:
> excellent idea @scott-huberty! i increased the test to validate this :) And I put one extra line to be more semantically correct test line with hedtools too

Woot! Thanks!!

scott-huberty (approved):

bruAristimunha:
ready to merge boss @scott-huberty!

larsoner (approved):
Merging `main` into this branch and marking for merge-when-green, thanks in advance @bruAristimunha !

PR #1536, opened 2026-03-12
Modernize _parse_ext to handle multi-extension filenames

scott-huberty:
Related to #1528 
Towards #1268

`_parse_ext` splits a filename into its stem and extension. It already contained a special case for `.nii.gz` files. here I refactor it to generally handle multi-extension filenames (EDIT: because I need `_parse_ext` to handle `.tsv.gz` files).


mne_bids/tests/test_copyfiles.py
```diff
@@ -221,7 +221,7 @@ def test_copyfile_eeglab(tmp_path, fname):
     bids_root = tmp_path
     data_path = testing_path / "EEGLAB"
     raw_fname = data_path / fname
-    new_name = bids_root / f"CONVERTED_{fname}.set"
+    new_name = bids_root / f"CONVERTED_{fname}"
```
scott-huberty, line 224: Here's why I think this test was weird and why I think it warrants a change.

 [copyfile_eeglab](https://mne.tools/mne-bids/stable/generated/mne_bids.copyfiles.copyfile_eeglab.html) will throw an error if `dest` contains a different extension than `src`.

On main, `fname == "test_raw.set"` and `new_name == "/CONVERTED_test_raw.set.set")`


So on L231, we do `copyfile_eeglab(raw_fname, new_name)`

and inside `copyfile_eeglab`, we do `_parse_ext("/CONVERTED_test_raw.set.set")` which  returns `('/CONVERTED_test_raw.set', '.set')` .. 

I think this just happened to work because now the files end up with the same extension even though `_parse_ext` didn't really do its job, and even though the caller passed a `dest` with a different extension than `src`.

scott-huberty:
Ok @bruAristimunha and @larsoner this one is ready for review too!

bruAristimunha (approved):
For me seems good! I think in some future we should use more the Path library for this, but it is good enough for now

scott-huberty:
> think in some future we should use more the Path library for this

totally agree.

scott-huberty:
Thanks @bruAristimunha ! I'll self-merge EOD unless someone else leaves a review.

mne_bids/path.py
```diff
@@ -1663,21 +1663,30 @@ def print_dir_tree(folder, max_depth=None, return_str=False):
...
-    elif ext == ".gz" and raw_fname.endswith(".nii.gz"):
-        ext = ".nii.gz"
-        fname = fname[:-4]  # cut off the .nii
-    return fname, ext
+    elif len(exts) == 1:
+        ext = exts[0]
+    else:  # >1 extension e.g. .nii.gz, .tsv.gz
+        ext = "".join(raw_fname.suffixes)
+        fname = raw_fname.name[: -len(ext)]  # cut off the .nii.gz, tsv.gz etc.
+    # TODO: Should we return a Path obj, and refactor call sites if needed?
```
larsoner: I would say yes, `pathlib.Path` should be preferred nowadays

mne_bids/path.py
```diff
@@ -70,8 +70,6 @@ def _find_empty_room_candidates(bids_path):
     bids_path = bids_path.copy()
 
     datatype = "meg"  # We're only concerned about MEG data here
-    bids_fname = bids_path.update(suffix=datatype).fpath
-    _, ext = _parse_ext(bids_fname)
```
scott-huberty, line 74: The `ty` extension in my IDE revealed these to be dead LOC

mne_bids/path.py
scott-huberty: OK @larsoner and @bruAristimunha done in [1f32683](https://github.com/mne-tools/mne-bids/pull/1536/commits/1f326830e451f15e5907f5abc4f947d7ba2e8e87) : )

scott-huberty:
I'll self-merge at the end of the work day unless someone wishes to do another pass at reviewing : )

bruAristimunha (approved):

scott-huberty:
Thanks @bruAristimunha !

PR #1539, opened 2026-03-14
Allow writing EEG data as BDF and EEGLAB formats

### LINKED ISSUE #1537: We should allow to write eeg as BDF
### Describe the problem

It is already possible within the EMG, and it is one of the options for the EEG.

I think it is the best format with some compression and good resolution. 

### Describe your solution

allow the writing of the BDF.

To be super honest, I don't know.

### Additional context

_No response_

bruAristimunha:
## Summary
- Add `BDF`, `EDF`, and `EEGLAB` to `EEG_CONVERT_FORMATS` and `EEGLAB` to `IEEG_CONVERT_FORMATS` in `config.py` to match the write logic that already supported these formats
- Update `write.py` to include `eeg` datatype in BDF conversion/writing branches
- Add `("BDF", "eeg")` test case to `test_preload`

Closes #1537

## Test plan
- [x] `pytest mne_bids/tests/test_write.py::test_preload -v` — all 4 cases pass (including new `BDF-eeg`)

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -3,8 +3,9 @@
 # Authors: The MNE-BIDS developers
 # SPDX-License-Identifier: BSD-3-Clause
 
-from mne import io
 from mne.io.constants import FIFF
+
+from mne import io
 
 BIDS_VERSION = "1.9.0"
 
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -3,9 +3,8 @@
 # Authors: The MNE-BIDS developers
 # SPDX-License-Identifier: BSD-3-Clause
 
-from mne.io.constants import FIFF
-
 from mne import io
+from mne.io.constants import FIFF
 
 BIDS_VERSION = "1.9.0"
 
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -21,7 +21,7 @@ ALLOWED_DATATYPES = EPHY_ALLOWED_DATATYPES + ["anat", "beh", "motion"]
 MEG_CONVERT_FORMATS = ["FIF", "auto"]
 EEG_CONVERT_FORMATS = ["BrainVision", "EDF", "BDF", "EEGLAB", "auto"]
 EMG_CONVERT_FORMATS = ["EDF", "BDF", "auto"]
-IEEG_CONVERT_FORMATS = ["BrainVision", "EEGLAB", "auto"]
+IEEG_CONVERT_FORMATS = ["BrainVision", "EDF", "EEGLAB", "auto"]
 NIRS_CONVERT_FORMATS = ["auto"]
 MOTION_CONVERT_FORMATS = ["tsv", "auto"]
 CONVERT_FORMATS = {
@@ -31,6 +31,16 @@ CONVERT_FORMATS = {
     "ieeg": IEEG_CONVERT_FORMATS,
     "nirs": NIRS_CONVERT_FORMATS,
     "motion": MOTION_CONVERT_FORMATS,
+}
+
+# Mapping from format names to file extensions - single source of truth
+# for format-to-extension resolution used in write_raw_bids.
+FORMAT_EXTENSIONS = {
+    "BrainVision": ".vhdr",
+    "EDF": ".edf",
+    "BDF": ".bdf",
+    "EEGLAB": ".set",
+    "FIF": ".fif",
 }
 
 # Orientation of the coordinate system dependent on manufacturer
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -19,9 +19,9 @@ EPHY_ALLOWED_DATATYPES = ["eeg", "emg", "ieeg", "meg", "nirs"]
 ALLOWED_DATATYPES = EPHY_ALLOWED_DATATYPES + ["anat", "beh", "motion"]
 
 MEG_CONVERT_FORMATS = ["FIF", "auto"]
-EEG_CONVERT_FORMATS = ["BrainVision", "auto"]
+EEG_CONVERT_FORMATS = ["BrainVision", "EDF", "BDF", "EEGLAB", "auto"]
 EMG_CONVERT_FORMATS = ["EDF", "BDF", "auto"]
-IEEG_CONVERT_FORMATS = ["BrainVision", "auto"]
+IEEG_CONVERT_FORMATS = ["BrainVision", "EEGLAB", "auto"]
 NIRS_CONVERT_FORMATS = ["auto"]
 MOTION_CONVERT_FORMATS = ["tsv", "auto"]
 CONVERT_FORMATS = {
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -15,10 +15,8 @@ from collections import OrderedDict, defaultdict
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
-import mne
 import mne.preprocessing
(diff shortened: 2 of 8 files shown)


bruAristimunha:
I was creating the BDF writing permission, but then I realized that we support the EEGLab option, but it was not consistently showing.

bruAristimunha:
@scott-huberty some nice PR for you review ;)

mne_bids/config.py
```diff
@@ -19,9 +19,9 @@
 ALLOWED_DATATYPES = EPHY_ALLOWED_DATATYPES + ["anat", "beh", "motion"]
 
 MEG_CONVERT_FORMATS = ["FIF", "auto"]
-EEG_CONVERT_FORMATS = ["BrainVision", "auto"]
+EEG_CONVERT_FORMATS = ["BrainVision", "EDF", "BDF", "EEGLAB", "auto"]
```
scott-huberty, line 22: I am surprised to see that `"EDF"` and `"EEGLAB"` weren't already here.

Because IIRC, on `main` You can for example convert an `*.mff` file to `.edf` format during `write_raw_bids`, yet `'EDF'` does not have an entry in  `EEG_CONVERT_FORMATS`. That's.. confusing 🤔 

I think we are currently maintaining two sources of truth for the convert behavior: here in `CONVERT_FORMATS`, and also in the downstream `if format != "auto"` conversion logic, in the `write` module. I think this is why you had to update code in both of these places for this PR.

I think it would be more DRY if `CONVERT_FORMATS` is the sole source of truth, and then the `if format != "auto"` logic becomes simpler:

```python
if format != "auto":
    if format not in FORMAT_EXTENSIONS:
        raise ValueError(...)

    if format not in CONVERT_FORMATS[datatype]:
        raise ValueError(...)

    convert = True
    bids_path.update(extension=FORMAT_EXTENSIONS[format])
```

Where `FORMAT_EXTENSIONS` is a currently non-existent dictionary like `{"BrainVision": ".vhdr",...}`

BUT I think we can save that for a separate PR 😊

scott-huberty (approved):

scott-huberty:
This LGTM! I will wait until tomorrow to merge, to give others a chance to review if they wish to.

mne_bids/config.py
drammock: > I am surprised to see that "EDF" and "EEGLAB" weren't already here.

I suspect that (prior to #1129) each entry had only the *recommended* format for each modality, not all possible *allowed* formats.

drammock: yep, see here: https://github.com/mne-tools/mne-bids/issues/659#issuecomment-753924819

scott-huberty: huh. Interesting.. It seems that we were able to expand `CONVERT_FORMATS` from "Please convert to `'BrainVision'`", to "here's all allowed formats", _without breaking anything_*, because we don't always defer to `CONVERT_FORMATS` as the authority on the matter. i.e. we just hard code that want to convert to brain vision directly in write_raw_bids:

https://github.com/mne-tools/mne-bids/blob/8e52e3ee4d6689e0f3da14b4f6596204fb6757c2/mne_bids/write.py#L2550-L2556


I think that this was true even at the time that `CONVERT_FORMATS` was introduced. So for me, the code intention was never that clear to begin with.

https://github.com/adam2392/mne-bids/blob/120101603168f3aa4072000399943b882d91c4f9/mne_bids/write.py#L1266-L1296


___________________________________________________________________________________________
*For example, on this PR branch, doing `write_raw_bids(some_curry_file.cdt, fmt="auto")` still converts the data to the "recommended" BrainVision (and not EEGLAB or EDF).

drammock: > for me, the code intention was never that clear to begin with

I'm not saying it actually worked that way, I'm saying that it was *intended* that it should work that way (based on Alex's proposed docstring, not Adam's code).  That intent would explain why there was only one format (plus `auto`) listed in each of the `*_CONVERT_FORMATS` constants (up until I broke that rule when I added EMG support). If the rest of Adam's implementation didn't follow/enforce that idea, well, that may not have been on purpose; I didn't notice any PR comments where the stated intent was discussed further / revised.

Regardless of the history, this would be a good time to:

1. decide if we want to allow user-forced conversion to *any allowed format* or only *recommended formats*
2. refactor things so that those constants actually reflect what is allowed (as you suggested @scott-huberty)

scott-huberty: > 1. decide if we want to allow user-forced conversion to _any allowed format_ or only _recommended formats_

We currently allow user-forced conversion to any allowed formats and I think we should keep it that way!

https://github.com/mne-tools/mne-bids/blob/8e52e3ee4d6689e0f3da14b4f6596204fb6757c2/mne_bids/write.py#L2458-L2488



> 2\. refactor things so that those constants actually reflect what is allowed (as you suggested @scott-huberty)

Hmmm  I think my vote would be +1 to this, but now that I am enlightened on the original intent of those constants, I am not feeling so strongly opinionated.. Open to other peoples opinions.

drammock: my feeling is that we need not be prescriptive here; folks might have good reasons why they want to write in an allowed-but-not-recommended format. The issue motivating Adam's original PR reflects this need: he had downstream scripts that expected BrainVision format, so he wanted to force conversion even though the files were already in a recommended format (EDF). So from me: +1 to allow writing to any allowed format, and +1 to either remove those constants, or update them to include all allowed formats (and also update the code to actually check against them before writing!  IIUC, right now they're only used when constructing the error msg when things go wrong)

scott-huberty: > +1 to either remove those constants, or update them to include all allowed formats (and also update the code to actually check against them before writing! IIUC, right now they're only used when constructing the error msg when things go wrong)

@drammock do you have a preference on whether this update/remove constants + refactor should be done in this PR or in a separate PR?

bruAristimunha: I create one variable that centralizes everything and check here. This way we have the control of what mne bids allow us to write, and what is the default configuration.

bruAristimunha:
hey @scott-huberty, can we merge here?

scott-huberty (approved):

scott-huberty:
LGTM. @drammock did you want to take a look one more time before merge?

mne_bids/write.py
```diff
@@ -2485,12 +2466,26 @@ def write_raw_bids(
                 f"Please use one of {CONVERT_FORMATS[datatype]} "
                 f"for {datatype} datatype."
             )
+        else:
+            convert = True
+            bids_path.update(extension=FORMAT_EXTENSIONS[format])
```
drammock, line 2471: [nitpick] hard to tell in GH UI, but is there a reason this isn't `extension=ext`?

drammock (approved):

mne_bids/write.py
scott-huberty: Inspecting on a test file locally (code below), then at this point in the codebase,  `ext` is `.bin`. But I think we want the extension to reflect the format we are converting to. 

```python
import mne
import mne_bids
fpath = mne.datasets.testing.data_path() / "EGI" / "test_egi.mff"
raw = mne.io.read_raw_egi(fpath)
bpath = mne_bids.BIDSPath(subject="01", task="foo", root=".")
mne_bids.write_raw_bids(raw, bpath, format="BrainVision", overwrite=True)
```

drammock: Ok +1 for merge then!

Issue #1538, opened 2026-03-13
BIDS-Validator


scott-huberty:
Related #1302 and #1288 

Right now we have CI's that test 1) The stable release of the _legacy_ validator (`validator-stable`, 2) the development version of the _legacy_ validator (`validator-main`, and 3) the schema validator (development version?) (`validator-main-schema`)*.

Only the Schema validator will be aware of newly added BIDS extensions** (EMG, eyetracking, etc). However right now it seems that our test suite does not know how to triage tests based on the validator version. For example, we simply skip validation of EMG datasets even though the schema validator can handle them (I verified this locally), probably because the `validator-stable` and `validator-main` CI's would fail:

https://github.com/mne-tools/mne-bids/blob/8e52e3ee4d6689e0f3da14b4f6596204fb6757c2/mne_bids/tests/test_write.py#L3262-L3263

I imagine we all agree that moving forward our CI's should be able to keep up with the BIDS specification?

If so, I am wondering if there is a preference for 1) dropping the legacy validator CI's, or 2) figuring out a way to skip `_bids_validate` for certain validator types (legacy, schema) and/or versions.
______________________________________________________________________________________________________

*I feel that these names are a bit of a misnomer. I would expect `validator-stable` to use the stable version of the schema validator.
**It is my understanding that the legacy validator will no longer be updated alongside updates to the specification (Correct me if I am wrong).

larsoner:
> If so, I am wondering if there is a preference for 1) dropping the legacy validator CI's, or 2) figuring out a way to skip _bids_validate for certain validator types (legacy, schema) and/or versions.

I don't see any harm in keeping a legacy one around, but it would be nice to test the latest and greatest validator.

Skipping for certain validator types seems easy enough... either a `pytest.mark.skipif` if test-wide or some `if/else` conditional guard based on validator version seems useful.

To keep things simple my vote would be to test against whatever the best/most up-to-date validator is, and make sure the code passes with *at least* that. If there are times it doesn't pass with an older validator, just add a `@pytest.mark.skipif(old_validator, reason=...)` to the test and move on...

sappelhoff:
I am fine with the proposal by Eric, but would personally vote for dropping the legacy validator completely, and to instead thoroughly use the schema validator.

drammock:
> I am fine with the proposal by Eric, but would personally vote for dropping the legacy validator completely, and to instead thoroughly use the schema validator.

[...]

PR #1540, opened 2026-03-15
TST: Run BIDS Schema Validator on EMG datasets

scott-huberty:
Toward #1538

### DIFF
--- a/mne_bids/tests/conftest.py
+++ b/mne_bids/tests/conftest.py
@@ -4,21 +4,18 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import platform
+import re
 import shutil
 
 import pytest
 from mne.utils import run_subprocess
+from packaging.version import Version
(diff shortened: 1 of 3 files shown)


larsoner (approved):
Looks like a reasonable approach to me! Thanks @scott-huberty 

Want to try some CI changes next according to https://github.com/mne-tools/mne-bids/issues/1538#issuecomment-4063496472 ?

larsoner:
... I removed the `Closes` comment since I think to close that we should follow what @sappelhoff proposed, and this doesn't quite get us there I think

scott-huberty:
> Looks like a reasonable approach to me! Thanks @scott-huberty
> 
> Want to try some CI changes next according to [#1538 (comment)](https://github.com/mne-tools/mne-bids/issues/1538#issuecomment-4063496472) ?

Yes 🤓

PR #1542, opened 2026-03-17
MAINT: Fully switch to Schema Validator

### LINKED ISSUE #1538: BIDS-Validator

scott-huberty:
Closes #1538 
This drops the legacy validator from our CI's. Since we no longer have to use the legacy validator, I changed the CI setup from installing a `bids-validator` command on PATH, to just installing deno runtime. The `_bids_validate` fixture will now use deno to run the validator, optionally allowing the dev to control the spec version via a bash environment variable. For example:

`BIDS_VALIDATOR_VERSION='2.4.0'`

For this environment variable, we special case two values, (`'stable'`, and `'dev'`), as aliases to run the latest version version on [JSR](https://jsr.io/@bids/validator), and the development spec, respectively. Personally I find this very handy!

 
Finally, if deno is not available, `_bids_validate` will check if the PyPI CLI is installed: `bids-validator-deno`. I think this will be nice for new and/or casual contributors, who don't want to install a non-python testing dependency. They can just do `python -m pip install bids-validator-deno` before running the test suite.


.github/workflows/unit_tests.yml
```diff
@@ -156,52 +160,36 @@ jobs:
...
-      with:
-        node-version: 20
-
-    - name: Set up deno (validator-main-schema)
-      if: matrix.bids-validator-version == 'validator-main-schema'
+    - name: Set up deno
       uses: denoland/setup-deno@v2
       with:
-        deno-version: v1.x
+        deno-version: v1.x  # TODO: Bump deno to v2?
```
larsoner: Assuming it works then I would say yes. Naively from 

https://github.com/denoland/setup-deno?tab=readme-ov-file#version-from-input

I would assume `lts` is the best choice here (should be stable and well supported and not need to be updated or changed hopefully ever after this, whereas a specific version pin might?)

larsoner (commented):
Love the red > green !

On first run of this branch I got

<details>
<summary>Expected failure traceback</summary>

```
$ pytest mne_bids/tests/test_write.py -k anonymize
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.13.9, pytest-9.0.2, pluggy-1.6.0
PySide6 6.10.1 -- Qt runtime 6.10.1 -- Qt compiled 6.10.1
rootdir: /Users/larsoner/python/mne-bids
configfile: pyproject.toml
plugins: anyio-4.12.1, timeout-2.4.0, qt-4.5.0, rerunfailures-16.1, cov-7.0.0
collected 136 items / 127 deselected / 9 selected                                                                                                              

mne_bids/tests/test_write.py FFFFFF.F.                                                                                                                   [100%]

=========================================================================== FAILURES ===========================================================================
______________________________________________________________________ test_fif_anonymize ______________________________________________________________________
mne_bids/tests/test_write.py:1050: in test_fif_anonymize
    _bids_validate(bids_root)
mne_bids/tests/conftest.py:23: in _validate
    raise FileNotFoundError(
E   FileNotFoundError: ⛔️ A BIDS validator runtime is required to run validation tests. Ensure Deno is available or install bids-validator-deno from PyPI
```

</details>

and then
```
pip install bids-validator-deno
```
indeed allowed things to pass!

.github/workflows/unit_tests.yml
scott-huberty: Done in [008f5c3](https://github.com/mne-tools/mne-bids/pull/1542/commits/008f5c3ee651b75f4484fc845894bfda7adb200f) and looks like things are coming back green!

larsoner:
@scott-huberty what's left to do here? Just need me to fix the conflicts, review, and merge?

scott-huberty:
> @scott-huberty what's left to do here? Just need me to fix the conflicts, review, and merge?

@larsoner yep ready for review!

larsoner:
Awesome, thanks @scott-huberty !

Issue #1529, opened 2026-03-03
Expose defacing code in API


nordme:
### Describe the problem

It would be great to have public-facing function for defacing MRI's. This code is implemented privately in MNE-BIDS currently. 

### Describe your solution

We could potentially convert the current function `mne_bids.write._deface` into a public function. 

### Describe possible alternatives

I'm open to suggestions for alternatives.

### Additional context

Creating a public function for defacing would allow us to write a defacing how-to example for the MNE-Python documentation that includes the defacing functionality implemented here in MNE-BIDS as mentioned in [#13695](https://github.com/mne-tools/mne-python/pull/13695).

larsoner:
Looks like a +1 from @hoechenberger and I'll add one as well. @nordme would you be up for making a PR for this?

nordme:
> Looks like a +1 from [@hoechenberger](https://github.com/hoechenberger) and I'll add one as well. [@nordme](https://github.com/nordme) would you be up for making a PR for this?

Yep, I can make a PR for this. :)

PR #1544, opened 2026-03-25
ENH: Make MRI defacing function public (cf issue: #1529) 

nordme:
PR Description
--------------

This PR exposes a previously private function for defacing MRI's to the public API. It closes Issue #1529 

Additionally, tests are added for testing the defacing function, and fixtures necessary for testing are added as well. 

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/tests/conftest.py
+++ b/mne_bids/tests/conftest.py
@@ -121,13 +121,13 @@ def _get_head_fids():
     return head_fids
 
 
-@pytest.fixture(scope=session)
+@pytest.fixture(scope="module")
 def mri_landmarks():
-    trans_name = "sample_audvis_trunc-trans.fif"
-    trans_path = test_path / "MEG" / "sample" / "trans" / trans_name
+    trans_name = 'sample_audvis_trunc-trans.fif'
+    trans_path = test_path / 'MEG' / 'sample' / trans_name
     trans = read_trans(trans_path)
     head_fids = _get_head_fids()
-    t1 = load_t1()
+    t1 = _load_t1()
     head_to_mri_trans = _get_head_to_vox_trans(t1)
     mri_fids = np.zeros(shape=head_fids.shape)
     for hfi, hfid in enumerate(head_fids):
@@ -136,7 +136,7 @@ def mri_landmarks():
     return mri_fids
 
 
-@pytest.fixture(scope=session)
+@pytest.fixture(scope="module")
 def t1_image():
     t1_im = _load_t1()
     return t1_im
--- a/mne_bids/tests/conftest.py
+++ b/mne_bids/tests/conftest.py
@@ -6,9 +6,8 @@
 import platform
 import re
 import shutil
-from pathlib import Path
-import numpy as np
 
+import numpy as np
 import pytest
 from mne import read_trans
 from mne.datasets import testing
@@ -16,7 +15,6 @@ from mne.io import read_raw_fif
 from mne.transforms import apply_trans
 from mne.utils import run_subprocess
 from packaging.version import Version
-
 
 test_path = testing.data_path(download=False)
 
@@ -101,30 +99,32 @@ def close_all():
 
 def _load_t1():
     import nibabel as nib
-    t1_path = test_path / 'subjects' / 'sample' / 'mri' / 'T1.mgz'
+
+    t1_path = test_path / "subjects" / "sample" / "mri" / "T1.mgz"
     t1 = nib.load(t1_path)
     return t1
 
 
 def _get_head_to_vox_trans(t1):
     from numpy.linalg import inv
+
     vox_to_ras = t1.header.get_vox2ras_tkr()
     ras_to_vox_trans = inv(vox_to_ras)
     return ras_to_vox_trans
 
 
 def _get_head_fids():
-    raw_path = test_path / 'MEG' / 'sample' / 'sample_audvis_trunc_raw.fif'
+    raw_path = test_path / "MEG" / "sample" / "sample_audvis_trunc_raw.fif"
     raw_info = read_raw_fif(raw_path).info
-    head_fids = [dig["r"] for dig in raw_info["dig"] if dig["kind"]==1]
+    head_fids = [dig["r"] for dig in raw_info["dig"] if dig["kind"] == 1]
     head_fids = np.array(head_fids)
     return head_fids
 
 
 @pytest.fixture(scope=session)
 def mri_landmarks():
-    trans_name = 'sample_audvis_trunc-trans.fif'
-    trans_path = test_path / 'MEG' / 'sample' / 'trans' / trans_name
+    trans_name = "sample_audvis_trunc-trans.fif"
+    trans_path = test_path / "MEG" / "sample" / "trans" / trans_name
     trans = read_trans(trans_path)
     head_fids = _get_head_fids()
     t1 = load_t1()
@@ -132,7 +132,7 @@ def mri_landmarks():
     mri_fids = np.zeros(shape=head_fids.shape)
     for hfi, hfid in enumerate(head_fids):
         t_fid = apply_trans(trans, hfid, move=True)
-        mri_fids[hfi] = apply_trans(head_to_mri_trans, t_fid*1e3, move=True)
+        mri_fids[hfi] = apply_trans(head_to_mri_trans, t_fid * 1e3, move=True)
     return mri_fids
 
 
--- a/mne_bids/tests/conftest.py
+++ b/mne_bids/tests/conftest.py
@@ -5,9 +5,18 @@
 
 import platform
 import shutil
+from pathlib import Path
+import numpy as np
 
 import pytest
+from mne import read_trans
+from mne.datasets import testing
+from mne.io import read_raw_fif
+from mne.transforms import apply_trans
 from mne.utils import run_subprocess
+
+
+test_path = testing.data_path(download=False)
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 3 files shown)


mne_bids/tests/test_write.py
```diff
@@ -4484,3 +4484,45 @@ def test_parallel_write_many_subjects(tmp_path):
...
+        deface_mri(t1_image, mri_landmarks, dict(deface, theta=bad_theta_val))
+
+
+@testing.requires_testing_data
+def test_deface_mri(t1_image, mri_landmarks):
+    """Test that defacing completes successfully."""
+    from nibabel.spatialimages import SpatialImage
+
+    defaced_mri = deface_mri(t1_image, mri_landmarks, None)
+    assert isinstance(defaced_mri, SpatialImage)
```
larsoner, line 4528: I would go a little farther than this... for example if you `get_fdata()` on the original image and the defaced image, ravel the results, and `np.corrcoef(x.ravel(), y.ravel())[0, 1]`, I'd expect the result to be high but not 1. So maybe between 0.8 and 0.9 or something? (Not sure how many voxels get zeroed, but adjust upper and lower as needed. If you have to adjust upper to 1 it's a problem, if you have to adjust lower to 0 it's a problem). Or maybe better, `np.isclose(x, y, atol=...).mean()` (for a suitable `atol` given the voxel values) gives you the proportion of voxels that remained the same, which should be quite high.

mne_bids/write.py
```diff
@@ -1279,9 +1279,45 @@ def _sidecar_json(
...
+       X / y / z coordinates of left pre-auricular, nasion, and right
+       pre-auricular points in MRI (voxel) coordinate space (3x3 list of
+       lists or array with nasion coords in the middle).
+    deface : None | dict
+        Dict should have two keys with positive numeric values. "Inset"
+        controls how close the blackout plane sits to the nasion point in
+        mm (a value of 5 will set the blackout plane 5 mm behind the
+        nasion); "theta" controls the angle of the blackout plane with
+        respect to vertical. Inset and theta values default to 5 and 15,
+        respectively.
```
larsoner, line 1303: This is a weird public API, let's make it better if we can. When making this public I'd expect something more like
```
def deface_mri(image, *, landmarks, inset=5, theta=15):
```
then adjust internal uses accordingly

larsoner (commented):
Looks like a good start!

mne_bids/write.py
```diff
@@ -1320,32 +1320,61 @@ def _sidecar_json(
...
+    This function blacks out MRI voxels anterior to a nearly vertical plane
+    set by the nasion location. Watershed surfaces made using a defaced
+    MRI will also have no identifying facial detail.
+
+    Parameters
+    ----------
+    image : MRI object
+        Anatomical image as a nibabel object.
+    landmarks : array-like
+       X / y / z coordinates of left pre-auricular, nasion, and right
```
drammock, line 1335:
```suggestion
       x / y / z coordinates of left pre-auricular, nasion, and right
```

mne_bids/write.py
```diff
@@ -1320,32 +1320,61 @@ def _sidecar_json(
...
+       pre-auricular points in MRI (voxel) coordinate space (3x3 list of
+       lists or array with nasion coords in the middle).
+    offset : int
+        Controls how close the blackout plane sits to the nasion point
+        in mm (an offset value of 5 will set the blackout plane 5 mm anterior
+        to the nasion, whereas -5 will move the plane posterior to the nasion).
+        Defaults to -5.
+    theta : int
```
drammock, line 1343: check freesurfer docs, does this *need* to be an integer, or are fractional degrees OK?

mne_bids/write.py
```diff
@@ -2955,11 +2989,22 @@ def write_anat(
...
             landmarks_deface = landmarks.get("deface")
+            offset = -5
+            theta = 15
             if landmarks_deface is None:
                 # Take first one if none is specified for defacing.
                 landmarks_deface = next(iter(landmarks.items()))[1]
             _, landmarks_deface = _get_landmarks(landmarks_deface, image_nii)
-            image_nii = _deface(image_nii, landmarks_deface, deface)
+            if isinstance(deface, dict):
+                if "offset" not in deface.keys() or "theta" not in deface.keys():
```
drammock, line 2999:
```suggestion
                if "offset" not in deface or "theta" not in deface:
```

mne_bids/write.py
```diff
@@ -2955,11 +2989,22 @@ def write_anat(
...
+                    raise RuntimeError(
+                        f"When deface is a dict, keys must "
+                        f"include offset and theta, but got "
+                        f"{deface.keys()})"
```
drammock, line 3003: not backwards compatible? previously the `deface=dict(...)` approach used key `inset` not `offset`

mne_bids/tests/conftest.py
```diff
@@ -91,3 +98,48 @@ def close_all():
...
+
+def _load_t1():
+    import nibabel as nib
+
+    t1_path = test_path / "subjects" / "sample" / "mri" / "T1.mgz"
+    t1 = nib.load(t1_path)
+    return t1
+
+
+def _get_head_to_vox_trans(t1):
```
drammock, line 111: IIUC this isn't actually "HEAD coord sys" to VOX, it's either "scanner RAS" or "freesurfer RAS" to VOX.  if that's right, function name should change accordingly

mne_bids/tests/conftest.py
```diff
@@ -91,3 +98,48 @@ def close_all():
...
+    from numpy.linalg import inv
+
+    vox_to_ras = t1.header.get_vox2ras_tkr()
+    ras_to_vox_trans = inv(vox_to_ras)
+    return ras_to_vox_trans
+
+
+def _get_head_fids():
+    raw_path = test_path / "MEG" / "sample" / "sample_audvis_trunc_raw.fif"
+    raw_info = read_raw_fif(raw_path).info
```
drammock, line 121: use the read_info function, skips reading the actual data

mne_bids/tests/conftest.py
```diff
@@ -91,3 +98,48 @@ def close_all():
...
+    return head_fids
+
+
+@pytest.fixture(scope="module")
+def mri_landmarks():
+    trans_name = "sample_audvis_trunc-trans.fif"
+    trans_path = test_path / "MEG" / "sample" / trans_name
+    trans = read_trans(trans_path)
+    head_fids = _get_head_fids()
+    t1 = _load_t1()
```
drammock, line 133: might be possible to avoid this read by defining the `t1_image` fixture first, and putting `t1_image` as an input to this function

mne_bids/tests/test_write.py
```diff
@@ -4598,3 +4561,63 @@ class RawANT:
...
+
+@testing.requires_testing_data
+def test_deface_mri(t1_image, mri_landmarks):
+    """Test that defacing completes successfully."""
+    from nibabel.spatialimages import SpatialImage
+
+    # reference mri
+    vox_sum = t1_image.get_fdata().sum()
+
+    # Check that a more negative offset leads to more voxels at 0
```
drammock, line 4604: do the defacing w/ default param values first, and then compare larger offsets / larger thetas versus the results of the default-param defacing (instead of against the "no defacing" case)

mne_bids/write.py
```diff
@@ -1353,32 +1353,79 @@ def _sidecar_json(
...
+    ----------
+    image : nibabel.spatialimages.SpatialImage
+        Anatomical image as a nibabel object.
+    landmarks : array-like
+       Coordinates (x / y / z) of left pre-auricular, nasion, and right
+       pre-auricular points in MRI (voxel) coordinate space (3x3 list of
+       lists or array with nasion coords in the middle).
+    offset : int
+        Controls how close the blackout plane sits to the nasion point
+        in mm (an offset value of 5 will set the blackout plane 5 mm anterior
```
larsoner: `mm` or voxels? If it's voxels it's typically but not always `mm`, worth checking

mne_bids/write.py
```diff
@@ -1389,16 +1436,21 @@ def _deface(image, landmarks, deface):
     idxs = np.transpose(idxs, [1, 2, 3, 0])  # (*image_data.shape, 3)
     idxs = idxs.reshape(-1, 3)  # (n_voxels, 3)
 
+    # image.affine is the voxel to RAS transform given the original MRI
     # convert to RAS by applying affine
     idxs = nib.affines.apply_affine(image.affine, idxs)
 
+    # get nasion into RAS
```
larsoner: Add somewhere if not present that nibabel is always operating in `mm` for its physical units (MNE tends to prefer SI `m`)
```suggestion
    # get nasion into RAS (in mm, which is what nibabel uses)
```

mne_bids/write.py
```diff
@@ -1353,32 +1353,79 @@ def _sidecar_json(
     return fname
 
 
-def _deface(image, landmarks, deface):
+def deface_mri(image, landmarks, offset=-5, theta=15):
```
larsoner: Nicer to make these kwarg-only?
```suggestion
def deface_mri(image, landmarks, *, offset=-5, theta=15):
```

mne_bids/write.py
```diff
@@ -3079,11 +3131,24 @@ def write_anat(
...
+                    deface.update(offset=-deface["inset"])
+                if "offset" not in deface or "theta" not in deface:
+                    raise RuntimeError(
+                        f"When deface is a dict, keys must "
+                        f"include offset and theta, but got "
+                        f"{deface.keys()})"
+                    )
+                offset = deface["offset"]
+                theta = deface["theta"]
+            image_nii = deface_mri(image_nii, landmarks_deface, offset, theta)
```
larsoner:
```suggestion
            image_nii = deface_mri(image_nii, landmarks_deface, offset=offset, theta=theta)
```

larsoner (commented):
Just a couple of minor nits, otherwise LGTM!

mne_bids/write.py
```diff
@@ -1353,32 +1353,79 @@ def _sidecar_json(
...
 
-    if not _is_numeric(theta):
-        raise ValueError(f"theta must be numeric (float, int). Got {type(theta)}")
+    if not isinstance(offset, int):
+        raise TypeError(f"offset must be an integer, but got type {type(offset)}.")
 
-    if inset < 0:
-        raise ValueError("inset should be positive, Got {inset}")
+    if not isinstance(theta, (int, float)):
+        raise TypeError(f"theta must be a float or integer, but got type {type(theta)}")
```
scott-huberty: @larsoner is it OK with you if we use `mne.utils._validate_type` here? It's private but we use it throughout the mne-bids code base already and we test against mne main.. Would probably save a few LOC and would save @nordme a bunch of tests that catch type errors : )

larsoner: Yeah that should be fine

Issue #1559, opened 2026-04-15
BIDSPath.match() can be expensive


larsoner:
On some systems, calling `BIDSPath.match` can cause significant overhead. See for example https://github.com/mne-tools/mne-bids-pipeline/issues/1214, where a cached run in MNE-BIDS-Pipeline spends 4 out of 5 minutes (!) in `BIDSPath.match`. This was also I think the motivation behind https://github.com/mne-tools/mne-bids/pull/1450, which helped in at least some cases. I will probably deal with some of this at the MNE-BIDS-Pipeline end, but hopefully we can still do things better here as well.

The current order of operations is:

1. Call `_return_root_paths` to get (almost) all paths in the given `root`
2. Filter the filenames
3. Convert them to bids paths

Ideally step (1) would *not* iterate over all files in the root when it can be much more selective.

I'll see if I can do some refactoring here to help.

dnacombo:
On my end, this type of issue is critical.

I'm dealing with a large dataset of 300+ subjects x 5 sessions x 4 runs + freesurfer derivatives (that may be scanned inadvertently sometimes) and causes massive delays (the 4-5 mins you're talking about would be ok, but the last run of mne-bids-pipeline init on my dataset took 50 min...) I'm tracing down the issue now to some occasional os.walk(root) calls that are an absolute no go for me. Right now, I'm dealing with a get_datatypes call that takes forever (though from what I understand, it should have been cached since it's called by a _get_datatypes_cached)...

I guess that would be a project in itself, but caching the entire file list and refreshing it only occasionally seems like the only option to me. I created a [flister.py](https://gist.github.com/dnacombo/1666e57338e1f4825dfa0d41ff4009bf) file lister at some point that returns a pd.DataFrame for this kind of things. I believe the bids specs could be embedded in a regex to retrieve useful info from the whole tree... Perhaps a different project, though. let me know.

larsoner:
Ouch!

Yeah one problem I hit immediately is that datasets I've structured as
```
bids_root/
bids_root/sub-01
...
bids_root/derivatives
bids_root/derivatives/freesurfer
...
bids_root/derivatives/mne-bids-pipeline
...
```
means that as I add files to `derivatives` the search gets longer, which is problematic. Some of this can be avoided but it would be nice to find a way to avoid it at the MNE-BIDS end, too.

[...]

PR #1560, opened 2026-04-15
MAINT: DRY BIDSPath.match and find_matching_paths

larsoner:
A small start toward #1559: make `BIDSPath.match` use `find_matching_files`. These had almost the same code in them, so route one through the other.

First commit (just pushed) uses the alternate code path, and asserts the two give the same files. (It also exposes that `tracking_systems` was missing from `find_matching_paths`!) Once this comes back green, I'll remove the old code path in favor of the new one.

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -2063,6 +2063,15 @@ def get_datatypes(root, verbose=None):
     return datatypes
 
 
+# Helpers for testing glob accesses
+def _root_glob(root, pattern):
+    return root.glob(pattern)
+
+
+def _root_rglob(root, pattern):
+    return root.rglob(pattern)
+
+
 @verbose
 def get_entity_vals(
     root,
@@ -2253,9 +2262,11 @@ def get_entity_vals(
     search_str = f"**/*{entity_long_abbr_map[entity_key]}-*_*"
     if include_match is not None:
         include_match = _ensure_tuple(include_match)
-        filenames = [f for im in include_match for f in root.glob(im + search_str)]
+        filenames = [
+            f for im in include_match for f in _root_glob(root, im + search_str)
+        ]
     else:
-        filenames = root.glob(search_str)
+        filenames = _root_glob(root, search_str)
 
     for filename in filenames:
         # Skip ignored directories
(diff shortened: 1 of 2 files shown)


bruAristimunha (approved):

larsoner:
Coverage drop is just because I deleted some lines that were covered, increasing the effective proportion of lines in the repo that are un-covered.

@drammock @scott-huberty should be good to go here

mne_bids/path.py
```diff
@@ -1115,21 +1115,20 @@ def match(self, *, ignore_json=True, ignore_nosub=False, check=False):
...
+        kwargs = {
+            f"{key}s": [val] if val is not None else val
+            for key, val in self.entities.items()
+        }
+        kwargs.update(
+            root=self.root,
+            check=check,
             ignore_json=ignore_json,
             ignore_nosub=ignore_nosub,
-            entities=self.entities,
```
larsoner, line 1124: Actually wait... `find_matching_paths` does *not* pass `entities` in its call to `_return_root_paths`. This suggests that the tests do not detect a regression in the number of file paths checked (this PR currently causes the slow path to be used in `_return_root_paths` now!). I'll see if I can add a test for this...

mne_bids/tests/test_path.py
```diff
@@ -224,17 +227,58 @@ def test_path_benchmark(tmp_path_factory):
...
     timed_all = timeit.timeit(
         "mne_bids.find_matching_paths(tmp_bids_root)", setup=setup, number=1
     )
+    assert count == 6878
     timed_ignored_nosub = timeit.timeit(
         "mne_bids.find_matching_paths(tmp_bids_root, ignore_nosub=True)",
         setup=setup,
         number=1,
     )
+    assert count == 360
```
larsoner, line 281: This new testing framework should give us more reliable numbers than `timeit`, and make it clear if we regress (or progress) in terms of the number of files iterated over

mne_bids/tests/test_path.py
```diff
@@ -247,25 +291,55 @@ def test_path_benchmark(tmp_path_factory):
...
     assert out_4 == []
 
+    # BIDSPath.match optimizations as well
+    path = BIDSPath(
+        datatype="meg",
+        extension=".ds",
+        root=tmp_bids_root,
+    )
+    paths = path.match()
+    # TODO: We *should* have n_subjects * n_sessions of these, but we get way more:
```
larsoner, line 326: I will open an issue for this once this PR is merged

larsoner:
Okay ready for real now I think @drammock @scott-huberty !

@bruAristimunha I added a bit more to the benchmarking code in case you're curious

mne_bids/tests/test_path.py
scott-huberty: So ideally it should be ` assert len(paths) == 36` ?😱

larsoner: Yeah...

mne_bids/path.py
```diff
@@ -2604,6 +2615,10 @@ def find_matching_paths(
         may be assigned ``description='cleaned'``.
 
         .. versionadded:: 0.11
+    tracking_systems : str | array-like of str | None
+        The tracking system used for digitization.
+
```
scott-huberty, line 2620: It doesn't have to be here or now, but I think it would be nice to explicitly state somewhere in our documentation that this corresponds to the `tracksys` entity in BIDS, that it is specific to motion capture BIDS (?), and maybe provide some example of a value?

scott-huberty (approved):

mne_bids/path.py
larsoner: Agreed. I just copied this from elsewhere... so good for a follow-up PR when someone is tinkering at some point

scott-huberty: Opened #1562

PR #1563, opened 2026-04-16
Make BIDSPath hashable

larsoner:
As part of optimizing MNE-BIDS-Pipeline, I'd like to make BIDSPath objects hashable so they can be used with `@functools.cache`, for example:
```
@functools.cache
def _find_matching_sidecar(
    bp: BIDSPath,
    ...
):
    return bp.find_matching_sidecar(...)
```
This isn't 100% correct but gets the idea across. Need to double check Python `__hash__` reqs, add class name hash, tests, etc. but wanted to get the ball rolling at least.

cc @dnacombo in case this might help you as well

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -460,6 +460,21 @@ class BIDSPath:
             "tracking_system": self.tracking_system,
         }
 
+    def __getstate__(self):
+        """Get the object state."""
+        state = self.entities
+        for key in ("root", "suffix", "extension", "datatype", "check"):
+            state[key] = getattr(self, key)
+        return state
+
+    def __setstate__(self, state):
+        """Set the object state."""
+        self.update(**state)
+
+    def __hash__(self):
+        """Compute the object hash."""
+        return hash(self.__getstate__().items())
+
     @property
     def basename(self):
         """Path basename."""
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -681,11 +681,11 @@ class BIDSPath:
         """Return the string representation for any fs functions."""
         return str(self.fpath)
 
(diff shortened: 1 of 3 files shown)


mne_bids/path.py
```diff
@@ -665,12 +680,10 @@ def __fspath__(self):
         return str(self.fpath)
 
     def __eq__(self, other):
-        """Compare str representations."""
-        return str(self) == str(other)
```
larsoner, line 669: This wasn't a complete / sufficient check for equality

larsoner: Though given the changes I'm having to make to the tests to be able to change this... I'm less convinced now

mne_bids/tests/test_path.py
```diff
@@ -2080,3 +2082,42 @@ def test_fpath_common_prefix(tmp_path):
         BIDSPath(root=tmp_path, subject="1", run="2").fpath
         == sub_dir / "sub-1_run-2.edf"
     )
+
+
+def test_hash():
+    """Test that BIDSPath is hashable."""
+    bp1 = BIDSPath(subject="01", datatype="eeg", root=Path("foo"))
+    bp2 = BIDSPath(subject="01", datatype="eeg", root=Path("foo"))
+    assert hash(bp1) == hash(bp2)
```
scott-huberty:
```suggestion
```
@larsoner feel free to tell me to buzz off if this is not ready for review yet!

Suggesting deletion here because this assert statement is repeated below?

larsoner: I think it's ready for review! Sure

larsoner:
@scott-huberty ready for review / merge from my end !

drammock (approved):

scott-huberty (approved):

Issue #1557, opened 2026-04-12
Improve sidecar parent checks


larsoner:
> Unrelated to this PR but #1508 already improved `find_matching_sidecar` to search the dataset root for sidecars that are missing from `bids_path.directory`.  I feel like this could be extended to search other parent directories. For example, this `scans.tsv` is in the session directory. 

 _Originally posted by @scott-huberty in [#1556](https://github.com/mne-tools/mne-bids/pull/1556/changes#r3066622948)_

drammock:
Before #1508 it searched everything at or below `sub-*/` (meaning it would already catch sidecars in `ses-*/` directories).  I think the problem for @scott-huberty in #1556 is that `scans.tsv` *is not a sidecar file*, it's a *data summary file* (https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#scans-file)

scott-huberty:
> Before [#1508](https://github.com/mne-tools/mne-bids/pull/1508) it searched everything at or below `sub-*/` (meaning it would already catch sidecars in `ses-*/` directories). I think the problem for [@scott-huberty](https://github.com/scott-huberty) in [#1556](https://github.com/mne-tools/mne-bids/pull/1556) is that `scans.tsv` _is not a sidecar file_, it's a _data summary file_ (https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#scans-file)

I guess you are right about that distinction. I wonder why my little trick in #1556 worked to 'make' `find_matching_sidecar` recognize the file. Might checkout the code before #1508 to see if it would recognize the scans file back then.

drammock:
the trick worked because it returned a single, correct path?  I'm not sure why you think it should find scans.tsv when datatype is set on the bidspath, since the scans.tsv file is at `sub-*/ses-*/*_scans.tsv` (not at `sub-*/ses-*/eeg/*_scans.tsv`).

scott-huberty:
> I'm not sure why you think it should find scans.tsv.  when datatype is set on the bidspath, since the scans.tsv file is at `sub-*/ses-*/*_scans.tsv` (not at `sub-*/ses-*/eeg/*_scans.tsv`)

I thought that because 

[...]

PR #1565, opened 2026-04-16
ENH: Speed up sidecar search

larsoner:
Let's take advantage of BIDS hierarchy to avoid some `**` in `find_matching_sidecar`.

Marking as draft as I need to check to see if it actually speeds up some accesses on a system + dataset I've been testing with.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,25 @@
+comment: false
+github_checks:  # too noisy, even though "a" interactively disables them
+  annotations: false
+
+codecov:
+  notify:
+    require_ci_to_pass: false
+
+coverage:
+  status:
+    patch:
+      default:
+        informational: true
+        target: 95%
+        if_no_uploads: error
+        if_not_found: success
+        if_ci_failed: failure
+    project:
+      default: false
+      library:
+        informational: true
+        target: 90%
+        if_no_uploads: error
+        if_not_found: success
+        if_ci_failed: failure
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -1987,31 +1987,46 @@ def _find_matching_sidecar(bids_path, suffix=None, extension=None, on_error="rai
 
     # We only use subject and session as identifier, because all other
     # parameters are potentially not binding for metadata sidecar files
-    search_str_filename = f"sub-{bids_path.subject}"
-    if bids_path.session is not None:
-        search_str_filename += f"_ses-{bids_path.session}"
+    candidate_list = []
 
-    # Find all potential sidecar files from bids_root/sub-* and immediate subdirs,
+    # Start with searches using subject as root
+    subj_base = f"sub-{bids_path.subject}"
+
+    # Find all potential sidecar files from bids_root/sub-* and session subdirs,
     # potentially taking into account the data type
     search_dir = bids_root / f"sub-{bids_path.subject}"
-    candidate_list = []
-    leaf_node = bids_path.datatype or "*"
-    search_dirs = [
-        search_dir,
-        search_dir / leaf_node,
-    ]
-    if bids_path.session is not None:
-        # Need to special-case  to look for sub-*/ses-*/sub-*_ses-*_scans.tsv
-        # when there is a datatype defined
-        if suffix == "scans" and leaf_node != "*":
-            search_dirs.append(search_dir / "ses-*")
-        search_dirs.append(search_dir / "ses-*" / leaf_node)
-    search_strs_complete = []
-    for search_dir in search_dirs:
+    this_name = f"{subj_base}*{search_suffix}"
+    search_strs_complete = [str(search_dir / this_name)]
+    subj_sessions_wildcard = False
+    if bids_path.datatype is not None:
         search_strs_complete.append(
-            str(search_dir / f"{search_str_filename}*{search_suffix}")
+            str(search_dir / bids_path.datatype / f"{subj_base}*{search_suffix}")
         )
-        candidate_list.extend(glob.iglob(search_strs_complete[-1]))
+    else:
+        subj_sessions_wildcard = True
+        search_strs_complete.append(
+            str(search_dir / "*" / f"{subj_base}*{search_suffix}")
+        )
+    if bids_path.session is None:  # could be any session subdir
+        this_dir = search_dir / "ses-*"
+        if not subj_sessions_wildcard:  # no need as the broader one will catch it
+            this_name = f"{subj_base}_ses-*{search_suffix}"
+            search_strs_complete.append(str(this_dir / this_name))
+        if bids_path.datatype is not None:
+            search_strs_complete.append(str(this_dir / bids_path.datatype / this_name))
+        else:
+            search_strs_complete.append(str(this_dir / "*" / this_name))
+    else:
+        this_dir = search_dir / f"ses-{bids_path.session}"
+        if not subj_sessions_wildcard:
+            this_name = f"{subj_base}_ses-{bids_path.session}*{search_suffix}"
+            search_strs_complete.append(str(this_dir / this_name))
+        if bids_path.datatype is not None:
+            search_strs_complete.append(str(this_dir / bids_path.datatype / this_name))
+        else:
+            search_strs_complete.append(str(this_dir / "*" / this_name))
+    for search_str in search_strs_complete:
+        candidate_list.extend(glob.iglob(str(search_str)))
     best_candidates = _find_best_candidates(bids_path.entities, candidate_list)
 
     # If no candidates found within subject directory, search at dataset root
@@ -2087,11 +2102,11 @@ def _find_matching_sidecar_shortcut(bids_path, suffix=None, extension=None):
     hierarchy_paths.append((bids_root, ""))
 
     # Now do the heavy lifting: look for the files
-    if suffix in ("scans", "coordsystem"):
+    if suffix in ("scans", "coordsystem", "electrodes"):
         path_end = f"{suffix}{extension or '.json'}"
         for dir_, entity_str in hierarchy_paths:
-            # for coordsystem, we need to check with task as well
-            if suffix == "coordsystem" and bids_path.task is not None:
+            # we need to check with task as well
+            if suffix != "scans" and bids_path.task is not None:
                 shortcut_file = dir_ / f"{entity_str}task-{bids_path.task}_{path_end}"
                 if shortcut_file.is_file():
                     return shortcut_file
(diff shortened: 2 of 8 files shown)


mne_bids/path.py
```diff
@@ -1944,45 +1944,58 @@ def _find_matching_sidecar(bids_path, suffix=None, extension=None, on_error="rai
...
-
-    search_str_complete = str(search_dir / f"{search_str_filename}*{search_suffix}")
-
-    candidate_list = glob.glob(search_str_complete, recursive=True)
+        search_dirs = [
+            search_dir,
+            search_dir / bids_path.datatype,
+        ]
+        if bids_path.session is not None:
+            search_dirs.append(search_dir / "ses-*" / bids_path.datatype)
```
drammock:
```py
    leaf_node = bids_path.datatype or "*"
    search_dirs = [search_dir, search_dir / leaf_node]
    if bids_path.session is not None:
        search_dirs.append(search_dir / "ses-*" / leaf_node)
```

can't submit as suggestion through UI (b/c first line isn't in diff I guess?)

larsoner:
Pushed the `leaf_node`, nice and DRY :heart: 

@drammock does this seem correct / reasonable to you given the discussion in #1557 ?

drammock:
> does this seem correct / reasonable to you given the discussion in #1557 ?

yes.  At least, it doesn't make anything worse than it already is :)

I think part of the difficulty is that users might have different desires when looking for sidecars. It might be
1. get me a specific sidecar that applies to only this data file
2. get me whichever sidecar is most specific, even if it applies to multiple files; i.e., there might be a file `root/sub-*/sub-*_whatever.json` that is the most specific/granular/applicable sidecar (no sidecar exists at `root/sub-*/ses-*/[datatype/]sub-*_ses-*_whatever.json`), and maybe we should return that even if the bidspath includes a session (currently we won't).
3. get me all the associated metadata at all levels (and merge them); BIDS has rules for how that merging should happen.
 
AFAICT the function originally meant to do only (1), but the later code that searches dataset root (*if and only if* lower-level metadata aren't found) partially addresses (2) (but I think still leaves out the intermediate-level sidecar in my example above).

larsoner:
Okay good news! This takes time down from ~13s to ~9s in a real-world test

| main | PR |
| -- | -- |
| image | image |

I still don't think it should take 10s for these sidecar checks, though!

Currently what we do in `find_matching_sidecar` is that we go through every reasonable search path, glob with `*`, then filter the resulting files to choose the best one. For well-formed / complete `BIDSPath`s and datasets written by MNE-BIDS, this seems like overkill in many cases. I suspect we could get away with a shortcut: construct the best-matched file (the one that would be written by `write_raw_bids`!) and just call `is_file()` on it, and take it if it exists. I'm going to poke around to see if this is feasible as well...

drammock:
> Currently what we do in `find_matching_sidecar` is that we go through every reasonable search path, glob with `*`, then filter the resulting files to choose the best one. For well-formed / complete `BIDSPath`s and datasets written by MNE-BIDS, this seems like overkill in many cases.

agree!  If we have the data file's BIDSPath, we can construct from that all the possible sidecar paths that would be relevant via the inheritance principle. With a valid BIDSPath I think there should never be any need to glob.

mne_bids/tests/test_path.py
```diff
@@ -297,22 +325,24 @@ def test_path_benchmark(bids_test_dir_dense, monkeypatch, path_counter):
     # This benchmark is to verify the speed-up in function call get_entity_vals with
     # `include_match=sub-*/` in face of a bids tree hosting derivatives and sourcedata.
     fnames = mne_bids.path._return_root_paths(tmp_bids_root)
-    assert len(fnames) == 6878
-    assert path_counter.count == 6878
+    assert len(fnames) == len(set(fnames))
+    assert len(fnames) == 10956
```
larsoner, line 385: These numbers went up because I modified the `bids_test_dir_dense` to have usable EEG files. Then corresponding updates happened below

mne_bids/tests/test_path.py
```diff
@@ -100,7 +103,6 @@ def bids_test_dir(tmp_path_factory):
     return bids_root
 
 
-@testing.requires_testing_data
```
larsoner, line 103: Add the mark to `bids_test_dir` so it's not needed all these places now (DRY)

mne_bids/path.py
```diff
@@ -1956,57 +1956,108 @@ def _find_matching_sidecar(bids_path, suffix=None, extension=None, on_error="rai
...
+    # 4. Check in datatype subdirs within session subdirs:
+    #
+    #    sub-N/ses-*/<datatype>/sub-N_ses-*<search_suffix>
+    #
+    search_strs_complete.append(str(this_dir / datatype_dir / search_name))
+
+    # Actually search now!
+    candidate_list = []
+    for search_str in search_strs_complete:
+        candidate_list.extend(glob.iglob(str(search_str)))
```
larsoner: This logic was rewritten because I think our logic was incorrect before and some files where missed (discovered by new tests). Hopefully the comments in the code now describe what should be searched and why.

larsoner (commented):
Okay I think this is finally ready for review/merge @drammock @scott-huberty 😅

codecov.yml
```diff
@@ -0,0 +1,25 @@
+comment: false
```
larsoner, line 1: Added this from MNE-Python. It's a bit annoying that every PR commit I push is red as soon as a single CI returns. It makes it very difficult to monitor for real CI failures just by looking at the tab icon at the top of the browser.

Since `codecov` isn't a required check anyway, okay to make these informative. Maintainers can look at coverage reports when merging (like they'd have to anyway if relying on "merge when green" behaviors)

scott-huberty (approved):
Reviewed the code and ran the tests locally. my main concern was whether the console output would be polluted with many many path strings upon failure of `find_matching_sidecar`, but the newline characters between each path ensure that everything is still readable!

```
(Pdb) path.find_matching_sidecar(suffix=suffix, extension=extension)
*** RuntimeError: Did not find any electrodes.tsv associated with sub-01_ses-01_task-testing_run-01.

The search strings were:
/private/var/folders/z1/jw4bwv7s13b66b4cz2_m6pzw0000gn/T/pytest-of-scotterik/pytest-632/mnebids_utils_test_bids_ds0/sub-01/sub-01*electrodes.tsv
/private/var/folders/z1/jw4bwv7s13b66b4cz2_m6pzw0000gn/T/pytest-of-scotterik/pytest-632/mnebids_utils_test_bids_ds0/sub-01/meg/sub-01*electrodes.tsv
/private/var/folders/z1/jw4bwv7s13b66b4cz2_m6pzw0000gn/T/pytest-of-scotterik/pytest-632/mnebids_utils_test_bids_ds0/sub-01/ses-01/sub-01_ses-01*electrodes.tsv
/private/var/folders/z1/jw4bwv7s13b66b4cz2_m6pzw0000gn/T/pytest-of-scotterik/pytest-632/mnebids_utils_test_bids_ds0/sub-01/ses-01/meg/sub-01_ses-01*electrodes.tsv
````

**Comment by larsoner** [comment]
Okay this one is ready for merge @drammock !

**Review by drammock — APPROVED** [review]

**Comment by drammock** [comment]
thanks @larsoner

**Comment by larsoner** [comment]
After this PR I get an error [in MNE-BIDS-Pipeline](https://app.circleci.com/pipelines/gh/mne-tools/mne-bids-pipeline/5572/workflows/99363790-eb42-4437-93fe-8e03abc8ee9c):
```
mne_bids_pipeline/_import_data.py:283: in _load_data
    raw = read_raw_bids(
<decorator-gen-411>:12: in read_raw_bids
    ???
../.pyenv/versions/3.13.11/lib/python3.13/site-packages/mne_bids/read.py:1367: in read_raw_bids
    raw = _handle_channels_reading(
../.pyenv/versions/3.13.11/lib/python3.13/site-packages/mne_bids/read.py:1106: in _handle_channels_reading
    raw.set_channel_types(
<decorator-gen-23>:12: in set_channel_types
    ???
../.pyenv/versions/3.13.11/lib/python3.13/site-packages/mne/_fiff/meas_info.py:581: in set_channel_types
    _check_set(info["chs"][c_ind], info["projs"], ch_type)
../.pyenv/versions/3.13.11/lib/python3.13/site-packages/mne/_fiff/meas_info.py:458: in _check_set
    raise RuntimeError(
E   RuntimeError: Cannot change channel type for channel MEG0113 in projector "mag_ssp_upright.fif : PCA-mags-v1"
```
Putting back our original logic, plus an assertion that the results match (with some code higher up to avoid shortcutting if the shortcut code actually finds a file):

<details>
<summary>Original code (ish)</summary>

```
    search_str_filename = f"sub-{bids_path.subject}"
    if bids_path.session is not None:
        search_str_filename += f"_ses-{bids_path.session}"
    # Find all potential sidecar files, doing a recursive glob
    # from bids_root/sub-*, potentially taking into account the data type
    search_dir = Path(bids_root) / f"sub-{bids_path.subject}"
    # ** -> don't forget about potentially present session directories
    if bids_path.datatype is None:
        search_dir = search_dir / "**"
    else:
        search_dir = search_dir / "**" / bids_path.datatype
    search_str_complete = str(search_dir / f"{search_str_filename}*{search_suffix}")
    candidate_list_orig = glob.glob(search_str_complete, recursive=True)
    best_candidates_orig = _find_best_candidates(bids_path.entities, candidate_list_orig)

    if len(best_candidates) == 1:
        out = Path(best_candidates[0])
        if shortcut_file is not None:
            assert out == shortcut_file
        assert len(best_candidates_orig) == 1, len(best_candidates_orig)
        assert out == Path(best_candidates_orig[0])
        return out
```

</details>


You can see the problem:
```
larsoner:~/python/mne-bids-pipeline$ pytest mne_bids_pipeline/ -k 117 --pdb
...
> /home/larsoner/python/mne-bids/mne_bids/path.py(2093)_find_matching_sidecar()
-> assert len(best_candidates_orig) == 1
(Pdb) p best_candidates
['/home/larsoner/mne_data/ds000117/sub-01/ses-meg/sub-01_ses-meg_task-facerecognition_channels.tsv']
(Pdb) p best_candidates_orig
[]
(Pdb) p bids_path
BIDSPath(
root: /home/larsoner/mne_data/ds000117
datatype: meg
basename: sub-01_ses-meg_task-facerecognition_run-02)
```
In other words, for `sub-01/ses-meg/sub-01_ses-meg_task-facerecognition_run-02` prior to this PR we didn't find the sidecar `sub-01/ses-meg/sub-01_ses-meg_task-facerecognition_channels.tsv`, and now we do. (Incidentally, this would explain why after this PR some MNE-BIDS-Pipeline runs had to completely re-run, because it found new sidecars, invalidating the cache!)

I'll fix the resulting "bug" with handling this particular dataset, but wanted to document this change of behavior. I'll include an updated changelog entry in the upcoming PR to point out that now some sidecar files may be found when they weren't previously.

drammock:
this is not actually a bug though, right?  I mean, a `run-02` file *should* find the run-agnostic sidecar that matches its sub, ses, and task. Or am I missing something?

larsoner:
Correct, it was a bug that we didn't find it before and we do now

PR #1573, opened 2026-04-20
ENH: Speed up lock context

larsoner:
Working on https://github.com/mne-tools/mne-bids/pull/1565 I saw a potential speedup in avoiding an expensive `inspect.stack()` call. Unfortunately, this was causing errors in CI of the form:
```
mne_bids/tests/test_write.py:4467: in test_parallel_write_many_subjects
    assert set(participants["participant_id"]) == set(expected_ids)
E   AssertionError: assert {'sub-001', '...sub-007', ...} == {'sub-001', '...sub-006', ...}
E     
E     Extra items in the right set:
E     'sub-261'
E     'sub-092'
E     'sub-144'
E     'sub-004'
E     Use -v to get more diff
```
which I can now reproduce *sometimes* on my machine.

I think this is due to the reentrant logic. @bruAristimunha I don't think it's working as intended... IIUC the reentrant stuff is meant to make it so that nested calls can reenter, like:
```
def _participants_tsv():
    with open_lock(a_file):
        func_b(a_file)

def func_b(b_file):
    with open_lock(b_file):
        fid...
```
However, the logic for how/when to do this is just to check to see if a lock has been opened for that file already. But for `write_raw_bids` doing `participants.tsv`, if any other thread has already acquired that lock with its own `func_a` call ... the current thread will proceed in reentrant mode, thus potentially "missing" some data if it calls `fid.read()` before the other thread has managed to append its data.

... at least I think that's what's happening. First commit is to show the problem. I don't think the test should be flaky, so I removed that decorator. And reducing the time spent in `stack.inspect` also helps show the issue I think.

Then I'll try to get rid of the reentrant stuff to see if it can fix the problem, but @bruAristimunha feel free to take over or open another separate PR if it's clear to you what the problem is already. I could be totally wrong on this.


larsoner:
See failures for example [in this CI run](https://github.com/mne-tools/mne-bids/actions/runs/24685147964/job/72192421976?pr=1573#step:18:738).

larsoner:
And the simplification seems to have worked.

Okay for you @bruAristimunha ?

mne_bids/_fileio.py
```diff
@@ -75,21 +71,24 @@ def _get_lock_context(path, timeout=None):
...
         # [1] = contextlib __enter__
         # [2] = _open_lock
         # [3] = contextlib __enter__
         # [4] = caller of _open_lock
-        where = inspect.stack()[4]
-        stack = f"{where.filename}:{where.lineno} {where.function}"
-        del where
+        # Using inspect.stack is expensive, so traverse directly instead
+        where = inspect.currentframe().f_back.f_back.f_back.f_back
+        stack = f"{where.f_code.co_filename}:{where.f_lineno} {where.f_code.co_name}"
```
bruAristimunha, line 84: Good idea!

mne_bids/_fileio.py
```diff
@@ -145,201 +146,22 @@ def _open_lock(path, *args, lock_timeout=None, **kwargs):
...
-                    if refcount_path.exists():
-                        refcount_path.unlink()
-                except OSError as exp:
-                    logger.debug(f"Could not remove refcount {refcount_path}: {exp}")
+                warn(
+                    f"Could not acquire lock for {canonical_path} "
+                    f"({exp}); proceeding without a lock."
+                )
+            if args or kwargs:
+                fid = stack.enter_context(open(canonical_path, *args, **kwargs))
```
bruAristimunha, line 164: Seems okay for me

bruAristimunha (approved):
I made the incremental and decremental because I was facing this error that you catch early. Happy that simplify is working well

mne_bids/tests/test_write.py
```diff
@@ -4443,7 +4443,6 @@ def test_write_bids_with_age_weight_info(tmp_path, monkeypatch):
     write_raw_bids(raw, bids_path=bids_path)
 
 
-@pytest.mark.flaky(reruns=3, reruns_delay=5)
```
bruAristimunha, line 4446: Just to make absolutely sure here, can we put some repetition of the test?

bruAristimunha: An marker to repeat

larsoner: We could, but implicitly we've gotten it by the number of CI runs (10ish?) times the number of commits here (one that actually implemented the change plus two more where I merged `main` into the branch) plus the times I ran tests locally to make sure (probably another 10). So we're at about 40 runs of this test across multiple OSes already. Given no failures (and the fact that the test itself is already a bit long at 3s locally) I don't think we need to add reruns here. But if it starts failing we can add a flaky decorator again, or better, make sure we fix whatever is causing the problem!

bruAristimunha: Fair point @larsoner , okay to merge this!

drammock (approved):

Issue #1562, opened 2026-04-15
document tracking_system parameter


scott-huberty:
I think it would be nice to explicitly state somewhere in our documentation that this corresponds to the `tracksys` entity in BIDS, that it is specific to motion capture BIDS (?), and maybe provide some example of a value?

_Originally posted by @scott-huberty in https://github.com/mne-tools/mne-bids/pull/1560#discussion_r3088469921_

viranovskaya:
Hi,

I’d be interested in working on this issue. I have experience with MNE-Python and BIDS/HED annotations, so the topic is close to my current work.

I could update the documentation to explain how `tracking_system` relates to the BIDS `tracksys` entity, clarify its use in Motion-BIDS, and add a concrete example.

If this is the intended scope and the issue is still available, I’d be happy to prepare a PR.

PR #1580, opened 2026-04-23
MAINT: Validate docs

larsoner:
I shouldn't have to make the comment https://github.com/mne-tools/mne-bids/pull/1577#discussion_r3132248253 , let's have CIs do it :smile: 

We did have some incorrect orders, a missing param, mis-rendered parameters (due to them being `acquisition: ` instead of `acquisition : ` for example), mis-ordered sections. Nothing really new added here other than the missing description for `tracking_system` (which I got wrong previously, it's for motion tracking :facepalm: ).

### DIFF
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -258,7 +258,7 @@ class BIDSPath:
         The acquisition session. Corresponds to "ses".
     task : str | None
         The experimental task. Corresponds to "task".
-    acquisition: str | None
+    acquisition : str | None
         The acquisition parameters. Corresponds to "acq".
     run : int | None
         The run number. Corresponds to "run".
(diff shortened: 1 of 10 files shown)


scott-huberty (approved):
Thanks @larsoner !!

mne_bids/path.py
```diff
@@ -281,21 +281,25 @@ class BIDSPath:
         may be assigned ``description='cleaned'``.
 
         .. versionadded:: 0.11
+    tracking_system : str | None
+        The motion tracking system.
+
+        .. versionadded:: 0.18
+    root : path-like | None
+        The root directory of the BIDS dataset.
```
scott-huberty, line 289: Just a note that this is a step towards #1562 , but I think I'll keep it open for a later PR that adds a bit more detail (e.g. example of an acceptable value)?

larsoner: Yeah I just put in a bare minimum to make the numpydoc validator happy, agreed we should keep #1562 open

larsoner:
@scott-huberty on uncontroversial PRs (like this, hopefully) I think you can safely start clicking "Enable auto-merge (squash)" if you're happy / approving

scott-huberty:
> @scott-huberty on uncontroversial PRs (like this, hopefully) I think you can safely start clicking "Enable auto-merge (squash)" if you're happy / approving

😅

Issue #1582, opened 2026-04-24
Tests fail on main branch on local computer


scott-huberty:
Locally on my Intel Mac, a large portion of the `mne_bids/tests/test_path.py` tests are now failing for me on the `main` branch. My main branch is up to date ( e8a8c49f01568c9d862a518b3cac3084c4c92ebe ).

I haven't been able to do a proper investigation yet, but sharing this here in case anyone else hits it.

According to `git bisect` the first bad commit is from #1573

```
git bisect start HEAD 2ddfbf7193b8d745fd6d719f3b02c4342105577e && git bisect run sh -c 'pytest -x mne_bids/tests/test_path.py -k test_get_entity_vals'
```

<details><summary>Output</summary>

```
031db3c3114e9ca6815a60998f815da3c4fc686c is the first bad commit
commit 031db3c3114e9ca6815a60998f815da3c4fc686c
Author: Eric Larson <larson.eric.d@gmail.com>
Date:   Tue Apr 21 07:02:26 2026 -0400

    ENH: Speed up lock context (#1573)
    
    * ENH: Speed up lock context
    
    * FIX: Simplify

 .github/workflows/unit_tests.yml |   1 +
 mne_bids/_fileio.py              | 230 +++++-------------------------------
 mne_bids/dig.py                  |   2 +-
 mne_bids/tests/test_fileio.py    | 246 +--------------------------------------
 mne_bids/tests/test_write.py     |   1 -
 mne_bids/tsv_handler.py          |   4 +-
 mne_bids/utils.py                |   8 +-
 mne_bids/write.py                |  16 +--
 8 files changed, 43 insertions(+), 465 deletions(-)
bisect found first bad commit
```

</details>

Here is the full output of `pytest mne_bids/tests/test_path.py > failure_log.txt`

[failure_log.txt](https://github.com/user-attachments/files/27065565/failure_log.txt)

larsoner:
Weird that no `macOS` CIs have caught this! But I can replicate on my arm64 macOS machine as well.

There is some stuff we should do better like be careful about modifying the state of `bids_test_dir`... I'll take a look soon

scott-huberty:
> Weird that no `macOS` CIs have caught this!

Right!! 🤔 🤔

scott-huberty:
FYI tests in `test_write` are also failing for me (locally), and it looks like stale/persisting lock files are tripping up the BIDS validator. Again it seems that the CI's do not hit this issue 🤔 

[...]

PR #1583, opened 2026-04-27
MAINT: Make testing data read-only

larsoner:
One little step toward #1582

1. Make `mne-testing-data` read-only. We were writing a few files there, which isn't great (I always wondered why sometimes I had an `ernoise_raw.fif` show up locally!)
2. rename `unit_tests` to `tests` for GHA (match MNE)
3. set default shell for all test commands on GHA for uniformity and DRY

### DIFF
--- a/mne_bids/copyfiles.py
+++ b/mne_bids/copyfiles.py
@@ -22,13 +22,13 @@ from scipy.io import loadmat, savemat
 
 from mne_bids._fileio import _open_lock
 from mne_bids.path import BIDSPath, _mkdir_p, _parse_ext
-from mne_bids.utils import _check_anonymize, _get_mrk_meas_date, warn
+from mne_bids.utils import _check_anonymize, _chmod_rw_R, _get_mrk_meas_date, warn
 
 
 def _copytree(src, dst, **kwargs):
     """See: https://github.com/jupyterlab/jupyterlab/pull/5150."""
     try:
-        sh.copytree(src, dst, copy_function=sh.copy, **kwargs)
+        sh.copytree(src, dst, **kwargs)
     except sh.Error as error:
         # ``copytree`` throws an error if copying to + from NFS even though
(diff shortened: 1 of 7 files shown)


mne_bids/copyfiles.py
```diff
@@ -63,7 +75,7 @@ def _get_brainvision_encoding(vhdr_file):
         either 'UTF-8' (default) or whatever encoding scheme is specified
         in the header.
     """
-    with _open_lock(vhdr_file, "rb") as ef:
+    with open(vhdr_file, "rb") as ef:
```
larsoner, line 74: @bruAristimunha do we really need to protect all our reads with `_open_lock`? I think we can assume that things like getting BV recording don't need it... people shouldn't be trying to do this at the same time this file could be being written for example. And simultaneous reads from multiple threads should be fine

bruAristimunha: i think the most complicated file that should not change is in the writing part. So, okay in this read part.

larsoner:
Okay finally green here. So far code like `write_raw_bids` had been untested on read-only paths, and often emitted errors. The behavior now is to write writeable paths, even if the source data is read-only. I think this is most consistent with what `raw.save(...)` and writing in Python in general will do. I think if people want MNE-BIDS *outputs* to be read-only, they can set them that way after the fact.

scott-huberty:
> Okay finally green here. So far code like `write_raw_bids` had been untested on read-only paths, and often emitted errors. The behavior now is to write writeable paths, even if the source data is read-only. I think this is most consistent with what `raw.save(...)` and writing in Python in general will do. I think if people want MNE-BIDS _outputs_ to be read-only, they can set them that way after the fact.

redownloading the testing data now, then will check out this PR and test things out!

scott-huberty (approved):
Thanks @larsoner !

PR #1587, opened 2026-04-29
FIX: more generous events.tsv reading

### LINKED ISSUE #947: 'trial_type' column with only 'n/a' will fail to extract any events even if 'value' column have the trigger codes
https://github.com/mne-tools/mne-bids/blob/2ef3516ba6ba54506901a773204f5f88adff224f/mne_bids/read.py#L378

Hi,

I have a BIDS dataset which has events.tsv looks something like this:
MNE-BIDS fails to extract any events/annotation from the events.tsv file:

onset	duration	sample	trial_type	response_time	stim_file	value
111.6972656250	0.0000000000	57189	n/a	n/a	n/a	1040
114.6972656250	0.0000000000	58725	n/a	n/a	n/a	2049
117.7617187500	0.0000000000	60294	n/a	n/a	n/a	1040
......

I think its's bacause the 'trial_type' column  has only 'n/a', so the code will drop all the events and will not look further for the 'value' columns.

From the BIDS document here :https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html
both trial type and value are optional, so I guess even if the trial_type only has 'n/a', we still should check the 'value' column for possible events?

### LINKED ISSUE #1547: BUG: NaN onset values in events.tsv cause silent corruption in read_raw_bids
## Description of the problem

`events_file_to_annotation_kwargs` drops `"n/a"` onsets via `_drop()` but does not handle `"nan"`, `"NaN"`, or `""` values (common in EEGLAB exports). These invalid onsets silently pass through and are converted to float, causing corrupt annotation data.

This was discovered processing OpenNeuro datasets during batch ingestion with [eegdash](https://github.com/bruAristimunha/eegdash).

## Steps to reproduce

```python
from mne_bids.read import events_file_to_annotation_kwargs
from pathlib import Path

# Create events.tsv with NaN onsets (as EEGLAB exports produce)
tsv = Path("/tmp/events.tsv")
tsv.write_text("onset\tduration\ttrial_type\nnan\t1.0\tstart\n0.5\t1.0\tstim\n")

kwargs = events_file_to_annotation_kwargs(tsv)
print(kwargs["onset"])  # Contains nan, should have been filtered
```

## Expected results

Rows with `"nan"`, `"NaN"`, or `""` onsets should be dropped, matching the existing handling of `"n/a"` onsets.

## Actual results

`"nan"` and `"NaN"` onsets pass through the filter and get converted to `float("nan")`, silently corrupting the annotations.

## Proposed fix

Extend the `_drop()` call to filter additional invalid values:

```python
# Current:
events_dict = _drop(events_dict, "n/a", "onset")

# Fixed:
events_dict = _drop(events_dict, ["n/a", "nan", "NaN", ""], "onset")
```

This requires `_drop()` to accept a list of values to match against (or multiple calls).

## Additional information

- Originally discussed at https://github.com/bruAristimunha/mne-bids/issues/23 with feedback from @drammock

bruAristimunha:
Closes #1547, closes #947.

Two small fixes in `events_file_to_annotation_kwargs`:

1. Drop rows with invalid onset markers (`n/a`, `nan`, `NaN`, `""`) before float conversion. Previously `nan`/`NaN` leaked through as `float("nan")` and `""` raised `ValueError`.
2. When `trial_type` is entirely `n/a` but `value` has trigger codes, fall back to `value` instead of dropping every row.

## Reproducer (end-to-end `read_raw_bids`)

```python
import openneuro

openneuro.download(
    dataset="ds004843",
    target_dir="/tmp/ds004843",
    include=["sub-001/eeg/sub-001_task-VisualSituationalAwareness_run-2_*",
             "dataset_description.json", "participants.tsv"],
)
bp = BIDSPath(subject="001", task="VisualSituationalAwareness", run="2",
              datatype="eeg", suffix="eeg", extension=".set",
              root="/tmp/ds004843")
raw = read_raw_bids(bp)
print(len(raw.annotations), "annotations")
```

| | result |
|--|--|
| `main` | `ValueError: cannot convert float NaN to integer` (in `mne/annotations.py:crop`) |
| this PR | 4048 annotations, all onsets finite |

Other affected datasets: `ds004841`, `ds004842`.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -65,7 +65,7 @@ Detailed list of changes
 - Fix :func:`mne_bids.read_raw_bids` and related read paths failing with ``PermissionError`` on datalad/git-annex datasets by keeping the file lock next to the symlink instead of its (read-only) target, and gracefully continuing without a lock when one cannot be created, by `Bruno Aristimunha`_ (:gh:`1569`)
 - Avoid modifying calibration files by making :func:`mne_bids.write_meg_calibration` copy instead of parsing and rewriting, by `Marijn van Vliet`_ (:gh:`1576`)
 - Fix bug with :meth:`mne_bids.BIDSPath.find_matching_sidecar` not searching parent directories properly, by `Eric Larson`_ (:gh:`1565`)
-- Fix :func:`mne_bids.read.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
+- Fix :func:`mne_bids.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
 - Fix :func:`mne_bids.read.events_file_to_annotation_kwargs` to fall back to the ``value`` column when ``trial_type`` is entirely ``n/a`` but ``value`` contains trigger codes, instead of dropping all events, by `Bruno Aristimunha`_ (:gh:`947`)
 
 ⚕️ Code health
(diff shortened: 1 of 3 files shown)


bruAristimunha:
pinging @larsoner and @scott-huberty because GitHub doesn't seem to be very happy notifying things today

doc/whats_new.rst
```diff
@@ -65,6 +65,8 @@ Detailed list of changes
 - Fix :func:`mne_bids.read_raw_bids` and related read paths failing with ``PermissionError`` on datalad/git-annex datasets by keeping the file lock next to the symlink instead of its (read-only) target, and gracefully continuing without a lock when one cannot be created, by `Bruno Aristimunha`_ (:gh:`1569`)
 - Avoid modifying calibration files by making :func:`mne_bids.write_meg_calibration` copy instead of parsing and rewriting, by `Marijn van Vliet`_ (:gh:`1576`)
 - Fix bug with :meth:`mne_bids.BIDSPath.find_matching_sidecar` not searching parent directories properly, by `Eric Larson`_ (:gh:`1565`)
+- Fix :func:`mne_bids.read.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
```
larsoner: This doesn't link properly, have to use public API endpoint

https://mne.tools/mne-bids/stable/generated/mne_bids.events_file_to_annotation_kwargs.html


```suggestion
- Fix :func:`mne_bids.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
```

This should have raised an error, it's bad that it didn't. We must not build with warnings-as-errors or something... I'll open an issue

larsoner (approved):

larsoner:
Committed the doc fix and marking for merge-when-green, thanks in advance @bruAristimunha !

PR #1588, opened 2026-04-29
MAINT: Modernize CI infrastructure and caching with uv

### LINKED ISSUE #1586: Reduce cache use in GHA
Looking at

Image

This is a lot more than even MNE-Python, which is currently just at ~4 GB. I'll investigate what we're caching, not sure we need to do it...

### LINKED ISSUE #1590: Doc build warnings not caught
See

https://github.com/mne-tools/mne-bids/pull/1587#discussion_r3161982983

The `:func:` didn't link properly but CIs were fine with it

larsoner:
Closes #1586
Closes #1590

Going to see if `setup-uv` can be smarter about caching. If not, I'll just get rid of the caching altogether. Also:

- Simplified some `Makefile` targets and moved some config from there to `pyproject.toml` while I was in there.
- No need for `make pep` anymore since pre-commit takes care of this stuff nowadays, so it's gone
- Bumping max Python version to 3.14 since it should be doable nowadays.
- Simplifying test data cache, should not depend on MNE version (just testing data version)
- Removing `build_docs` job from GHA to reduce usage, CircleCI already does this better
- Made GHA jobs depend on `style` job to save some CI cycles
- Reduced matrix from 3.11 + 3.14 on all jobs to just 3.14, plus one run with 3.11
- Speed up macOS Intel and Windows CIs by skipping some slow tests there (it was taking > 6m whereas others were ~3m)
- Deletes some unneeded files from `mne-testing-data` to speed up caching step


.github/workflows/tests.yml
```diff
@@ -189,69 +168,75 @@ jobs:
...
-
-    - name: Install MNE (stable)
-      if: matrix.mne-version == 'mne-stable'
+    - name: Workaround for macOS behavior on login shells
+      if: startsWith(matrix.os, 'macos')
       run: |
-        git clone --single-branch --branch maint/1.11 https://github.com/mne-tools/mne-python.git
-        python -m pip install -e ./mne-python
+        echo "export PATH=\"${{ github.workspace }}/.venv/bin:$PATH\"" | tee -a ~/.bash_profile
+        echo "PATH=\"${{ github.workspace }}/.venv/bin:$PATH\"" | tee -a $GITHUB_ENV
```
scott-huberty: is this work around because on a login shell, macOS, the virtual environment activation is not persisting (i.e., `activate-environment: true` is not working as expected)? If so, @drammock and I have hit the same issue in https://github.com/mne-tools/mne-python/pull/13704

larsoner: Yep, see https://github.com/actions/setup-python/issues/132

scott-huberty: Ah ok. Our work around over at https://github.com/mne-tools/mne-python/pull/13704 is to prepend any `python` / `pytest` commands with `Pixi run`. Feel free to chime in over there if you would advocate for an approach similar to the one you've taken here.

larsoner:
@scott-huberty I think this one is finally ready for review/merge !

larsoner:
(Once you're happy feel free to cilck the auto-merge button and I'll adjust the branch protection rules to actually allow it to be merged)

scott-huberty (approved):

PR #1556, opened 2026-04-10
FIX: scans.tsv: resolve the timestamp timezone from the historical date if you can

scott-huberty:
Fixes #1555 

- When parsing a timezone-naive timestamp from a `scans.tsv` file, by default we try to attach the computers local timezone, but using the timezone rules for ~that date~ the date of the timestamp (this was the behavior prior to https://github.com/mne-tools/mne-bids/commit/9c550e1252353f1f693b6e7ecb9021f591001714 )

- But if we hit this Windows platform pre-epoch timestamp error, then we fallback to just attaching the computers *current* timezone to the timestamp (even though this might be incorrect sometimes).

@larsoner @drammock once the CI's are green can I ask one of you to take a close look at this?

This one was tricky because the pre-epoch datetime test had the same problematic behavior as the function we were patching (i.e., it queried the computer to get the current local timezone information). In order to make the test more deterministic I had to do some monkey patching...

PR #1591, opened 2026-04-29
FIX: more generous scans.tsv reading

bruAristimunha:
Mirrors #1587 for `scans.tsv`. Three small fixes in `_handle_scans_reading`:

1. Malformed `acq_time` (`""`, `"NaN"`, non-ISO strings) now warns and leaves `meas_date` unchanged instead of crashing on `strptime`.
2. Recording absent from the `filename` column now warns and skips instead of raising. The close-matches diagnostic from `_verbose_list_index` is preserved in the warning text.
3. Header-only `scans.tsv` short-circuits instead of `IndexError` on `fnames[0].suffix`.

Also: parse `acq_time` via `datetime.fromisoformat` to accept ISO 8601 variants (space separator, explicit `+HH:MM` offsets, microsecond precisions other than 6 digits). Naive parses still defer to `_convert_dt_to_utc` (post-#1556); tz-aware parses skip straight to `.astimezone(UTC)`.

## Reproducer (end-to-end `read_raw_bids`)

```python
import openneuro

openneuro.download(
    dataset="ds006377",
    target_dir="/tmp/ds006377",
    include=["sub-225/sub-225_scans.tsv",
             "sub-225/nirs/sub-225_task-001RestingLH_run-1_nirs*",
             "dataset_description.json", "participants.tsv"],
)
bp = BIDSPath(subject="225", task="001RestingLH", run="1",
              datatype="nirs", suffix="nirs", extension=".snirf",
              root="/tmp/ds006377")
raw = read_raw_bids(bp)
print("meas_date:", raw.info["meas_date"])
```

`sub-225_scans.tsv` has empty `acq_time` cells (`filename\tacq_time\nnirs/...\t\n`).

| | result |
|--|--|
| `main` | `ValueError: time data '.0' does not match format '%Y-%m-%dT%H:%M:%S.%f'` (in `datetime.strptime`) |
| this PR | `RuntimeWarning: ... contains an invalid acq_time value: ''. Leaving raw.info['meas_date'] unchanged.` — read continues, `meas_date` retained from the SNIRF file (`2022-05-17 12:37:15+00:00`) |

Other affected dataset: `ds003844` has `acq_time=1934-09-01T13:42:17.9008789` (7-digit microseconds) — this PR truncates via `fromisoformat` rather than crashing. Discovered ingesting OpenNeuro datasets with [eegdash](https://github.com/bruAristimunha/eegdash). Supersedes #1532 (closed by me as too broad).

### DIFF
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -960,14 +960,8 @@ def test_handle_scans_reading_brainvision(tmp_path):
             ("acq_time", ["2000-01-01T12:00:00.000000Z"]),
         ]
     )
-    test_scan_edf = OrderedDict(
-        [
-            ("filename", [Path("eeg/sub-01_ses-eeg_task-rest_eeg.edf")]),
-            ("acq_time", ["2000-01-01T12:00:00.000000Z"]),
-        ]
-    )
     os.mkdir(tmp_path / "eeg")
-    for test_scan in [test_scan_eeg, test_scan_vmrk, test_scan_edf]:
+    for test_scan in [test_scan_eeg, test_scan_vmrk]:
         _to_tsv(test_scan, tmp_path / test_scan["filename"][0])
 
     bids_path = BIDSPath(
@@ -978,12 +972,6 @@ def test_handle_scans_reading_brainvision(tmp_path):
 
     for test_scan in [test_scan_eeg, test_scan_vmrk]:
         _handle_scans_reading(tmp_path / test_scan["filename"][0], raw, bids_path)
-
-    with pytest.warns(RuntimeWarning, match="is not listed in"):
-        out = _handle_scans_reading(
-            tmp_path / test_scan_edf["filename"][0], raw, bids_path
-        )
-    assert out is raw
 
 
 def _scans_setup(tmp_path):
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -1025,10 +1025,11 @@ def test_handle_scans_reading_filename_not_listed(tmp_path):
     """Recording absent from ``scans.tsv`` warns and leaves ``meas_date`` unchanged."""
     raw, bids_path, scans_path = _scans_setup(tmp_path)
     initial = raw.info["meas_date"]
+    fake_fname = Path("eeg/sub-foo_task-bar_eeg.fif")
     _to_tsv(
         OrderedDict(
             [
-                ("filename", [Path("eeg/sub-99_task-test_eeg.fif")]),
+                ("filename", [fake_fname]),
                 ("acq_time", ["2024-02-04T17:47:18.000000Z"]),
             ]
         ),
--- a/mne_bids/tests/test_read.py
+++ b/mne_bids/tests/test_read.py
@@ -984,46 +984,11 @@ def test_handle_scans_reading_brainvision(tmp_path):
     assert out is raw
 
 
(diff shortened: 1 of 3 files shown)


larsoner (approved):
Seems reasonable to me. @scott-huberty do you want to look and merge if you're happy?

mne_bids/tests/test_read.py
```diff
@@ -977,8 +977,104 @@ def test_handle_scans_reading_brainvision(tmp_path):
...
+
+
+def test_handle_scans_reading_filename_not_listed(tmp_path):
+    """Recording absent from ``scans.tsv`` warns and leaves ``meas_date`` unchanged."""
+    raw, bids_path, scans_path = _scans_setup(tmp_path)
+    initial = raw.info["meas_date"]
+    _to_tsv(
+        OrderedDict(
+            [
+                ("filename", [Path("eeg/sub-99_task-test_eeg.fif")]),
```
scott-huberty:
```suggestion
    fake_fname = Path("eeg/sub-foo_task-bar_eeg.fif")
    _to_tsv(
        OrderedDict(
            [
                ("filename", [fake_fname]),
```

scott-huberty: I just wanted to make the code intent really explicit (it took me a second to reason about).

scott-huberty: On another note, and this doesn't have to be fixed here.

 The full runtime error is:

```
*** RuntimeWarning: eeg/sub-01_task-test_eeg.fif is not listed in sub-01_scans.tsv (PosixPath('eeg/sub-01_task-test_eeg.fif') is not in list. Did you mean one of ['eeg/sub-foo_task-bar_eeg.fif']?). Leaving raw.info['meas_date'] unchanged.
```

Which to me is kind of confusing 😵‍💫

At the very least there should probably be a colon between `sub-01_scans.tsv` and `(PosixPath('eeg/sub-01_task-test_eeg.fif')`?

mne_bids/tests/test_read.py
```diff
@@ -977,8 +977,104 @@ def test_handle_scans_reading_brainvision(tmp_path):
     for test_scan in [test_scan_eeg, test_scan_vmrk]:
         _handle_scans_reading(tmp_path / test_scan["filename"][0], raw, bids_path)
 
-    with pytest.raises(ValueError, match="not in list"):
-        _handle_scans_reading(tmp_path / test_scan_edf["filename"][0], raw, bids_path)
+    with pytest.warns(RuntimeWarning, match="is not listed in"):
+        out = _handle_scans_reading(
+            tmp_path / test_scan_edf["filename"][0], raw, bids_path
+        )
+    assert out is raw
```
scott-huberty: I think that this case (passing a non-existent filepath to `_handle_scans_reading`) is now covered by the new unit test `test_handle_scans_reading_filename_not_listed`

@larsoner is it OK with you if we remove this?

scott-huberty (approved):
Thanks! A few nitpicks. I have not looked deeply into why the CI's are failing (might be unrelated to the changes here).

mne_bids/tests/test_read.py
bruAristimunha: i removed, and i aggree with you @scott-huberty ;)

bruAristimunha:
can we merge? @larsoner and @scott-huberty

larsoner:
Yep, thanks @bruAristimunha !

PR #1592, opened 2026-04-29
[MRG] ENH: Preserve enriched dataset_description.json and user-authored README

### LINKED ISSUE #1548: ENH: write_raw_bids() should not overwrite enriched dataset_description.json
## Describe the problem

`write_raw_bids()` creates or overwrites `dataset_description.json` with minimal info (just `Name` and `BIDSVersion`). If a user calls `make_dataset_description()` first to populate rich metadata (authors, funding, DOI, etc.), a subsequent `write_raw_bids()` call clobbers that enriched version.

This forces users to carefully order their calls — `make_dataset_description()` must come **after** all `write_raw_bids()` calls — which is unintuitive and error-prone.

Related: #302

## Describe your solution

Any of these approaches would fix the issue:

1. **Merge rather than overwrite**: `write_raw_bids()` should read the existing `dataset_description.json` (if present) and only add missing required fields (`Name`, `BIDSVersion`) without removing existing keys.
2. **Don't overwrite if richer content exists**: If `dataset_description.json` already exists and contains more than the minimal fields, skip writing it.
3. **Accept a parameter**: Add a `write_dataset_description=True` parameter to `write_raw_bids()` so users can opt out of the automatic dataset_description creation.

Option 1 (merge) would be the most user-friendly and backwards-compatible.

## Describe possible alternatives

The current workaround is to always call `make_dataset_description()` **after** all `write_raw_bids()` calls with `overwrite=True`. This works but is fragile and unintuitive.

## Current downstream workaround (MOABB)

```python
# Write all raw files first
for subject, session, run in data:
    mne_bids.write_raw_bids(raw, bids_path, ...)

# Then write dataset_description AFTER all write_raw_bids calls
# so our enriched version overwrites the minimal one
mne_bids.make_dataset_description(path=str(root), overwrite=True, **desc_kwargs)
```

## Additional context

- Originally discussed at https://github.com/bruAristimunha/mne-bids/issues/14 with feedback from @drammock (who suggested the merge approach)

### LINKED ISSUE #1550: ENH: write_raw_bids() should not overwrite existing README
## Describe the problem

`write_raw_bids()` creates a generic boilerplate README file ("This dataset was created with...") that contains no useful dataset-specific information. If a user has already created a comprehensive README with dataset-specific documentation, `write_raw_bids()` overwrites it.

Related: #1042

## Describe your solution

Two possible approaches:

1. **Don't overwrite existing README**: If a README file already exists with content beyond the boilerplate, skip writing it.
2. **Accept a `readme` parameter**: Add an optional `readme` string parameter to `write_raw_bids()` that, when provided, is used instead of the boilerplate.

Option 1 is the least disruptive change and most backwards-compatible.

## Describe possible alternatives

The current workaround is to write the README **after** all `write_raw_bids()` calls, overwriting the boilerplate. This works but requires careful ordering of operations.

## Current downstream workaround (MOABB)

```python
# After all write_raw_bids calls, overwrite the boilerplate README
readme_path = Path(root) / "README"
readme_path.write_text(_build_readme(dataset), encoding="utf-8")
```

Where `_build_readme()` generates a comprehensive document with paradigm info, channel setup, references, license, etc.

## Additional context

- Originally discussed at https://github.com/bruAristimunha/mne-bids/issues/16 with feedback from @drammock

bruAristimunha:
PR Description
--------------

Two MOABB-driven friction points in ``write_raw_bids``:

- **README** (#1550): add ``readme=True/False`` parameter; default keeps current append-refs behavior; ``readme=False`` skips creating/modifying the file entirely. Also fix racy ``lock→read→unlock→write`` in ``_readme`` to a single lock for the full read-modify-write.
- **dataset_description.json** (#1548): preserve BIDS-spec keys not modeled by ``make_dataset_description`` (e.g. ``Description``, ``DatasetLinks``) when merging with an existing file.

Closes #1548. Closes #1550.

Merge checklist
---------------

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -590,12 +590,12 @@ def _events_json(
     _write_json(fname, new_data, overwrite=overwrite)
 
 
-def _readme(datatype, fname, overwrite=False):
+def _readme(datatype, fname):
     """Create a README file and save it.
 
-    This will write a README file containing an MNE-BIDS citation.
-    If a README already exists, the behavior depends on the
-    ``overwrite`` parameter, as described below.
+    If a README already exists, append an MNE-BIDS citation to it
+    unless one is already present. Otherwise, create a new README
+    containing an MNE-BIDS citation.
 
     Parameters
     ----------
@@ -603,17 +603,11 @@ def _readme(datatype, fname, overwrite=False):
         The type of data contained in the raw file ('meg', 'eeg', 'ieeg')
     fname : str | mne_bids.BIDSPath
         Filename to save the README to.
-    overwrite : bool
-        Whether to overwrite the existing file (defaults to False).
-        If overwrite is True, create a new README containing an
-        MNE-BIDS citation. If overwrite is False, append an
-        MNE-BIDS citation to the existing README, unless it
-        already contains that citation.
     """
     # Hold the lock across read and write so concurrent writers cannot
     # observe a partially written file.
     with _open_lock(fname):
-        if fname.is_file() and not overwrite:
+        if fname.is_file():
             with open(fname, encoding="utf-8-sig") as fid:
                 orig_data = fid.read()
             mne_bids_ref = REFERENCES["mne-bids"] in orig_data
@@ -2359,7 +2353,7 @@ def write_raw_bids(
     # XXX: can include README overwrite in future if using a template API
     # XXX: see https://github.com/mne-tools/mne-bids/issues/551
     if readme:
-        _readme(bids_path.datatype, readme_fname, False)
+        _readme(bids_path.datatype, readme_fname)
 
     # save all participants meta data
     _participants_tsv(
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -607,23 +607,20 @@ def _readme(datatype, fname):
     # Hold the lock across read and write so concurrent writers cannot
     # observe a partially written file.
     with _open_lock(fname):
+        orig_data = ""
         if fname.is_file():
             with open(fname, encoding="utf-8-sig") as fid:
                 orig_data = fid.read()
-            mne_bids_ref = REFERENCES["mne-bids"] in orig_data
-            datatype_ref = REFERENCES[datatype] in orig_data
-            if mne_bids_ref and datatype_ref:
-                return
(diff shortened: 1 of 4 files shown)


mne_bids/write.py
```diff
@@ -2345,7 +2358,8 @@ def write_raw_bids(
     # save readme file unless it already exists
     # XXX: can include README overwrite in future if using a template API
     # XXX: see https://github.com/mne-tools/mne-bids/issues/551
-    _readme(bids_path.datatype, readme_fname, False)
+    if readme:
+        _readme(bids_path.datatype, readme_fname, False)
```
larsoner: So the private `_readme` function is only ever called with `overwrite=False`, right? If so we should simplify the logic there and remove the param from `_readme`

bruAristimunha: Done in 5d9216d5 — dropped the `overwrite` parameter and the dead `overwrite=True` branch.

bruAristimunha: Follow-up in 976cbd04: after dropping the param, the two branches were still building essentially the same string, so I unified them into a single assembly path (the missing-file case becomes `orig_data = ""`, ref checks evaluate False, early-return cannot fire, prefix collapses to `""`). Output bytes are identical for the new/append/early-return/partial-ref cases.

mne_bids/write.py
```diff
@@ -590,44 +590,38 @@ def _events_json(
...
-    if fname.is_file() and not overwrite:
-        with _open_lock(fname, encoding="utf-8-sig") as fid:
-            orig_data = fid.read()
+    # Hold the lock across read and write so concurrent writers cannot
+    # observe a partially written file.
+    with _open_lock(fname):
+        orig_data = ""
+        if fname.is_file():
+            with open(fname, encoding="utf-8-sig") as fid:
+                orig_data = fid.read()
```
larsoner:
```suggestion
            orig_data = fname.read_text("utf-8-sig")
```

larsoner (approved):
Simplified some logic and added `.. versionadded::` directive, marking for merge-when-green. Thanks in advance @bruAristimunha !

PR #1593, opened 2026-04-29
FIX: detect TSV encoding before reading

### LINKED ISSUE #1594: BUG: _from_tsv crashes on Latin-1 encoded TSV files (e.g., µ character)
## Description of the problem

`_from_tsv` uses `encoding="utf-8-sig"` which crashes on Latin-1 encoded TSV files containing characters like `µ` (micro-sign, common in European datasets with µV units in `channels.tsv`).

Found while processing OpenNeuro datasets in batch with [eegdash](https://github.com/sccn/EEGDash). A scan of 577 cloned EEG/iEEG datasets (~413k TSV files) found **767 non-UTF-8 TSV files across 8 datasets**: `ds006233` (347), `ds004621` (167), `ds003620` (160), `ds005692` (59), `ds004598` (20), `ds005691` (8, iEEG), `ds003574` (5), `ds004588` (1).

## Steps to reproduce

Minimal:
```python
from pathlib import Path
from mne_bids.tsv_handler import _from_tsv

tsv_path = Path("/tmp/test.tsv")
tsv_path.write_bytes(b"name\tunit\nEEG\t\xb5V\n")  # 'µV' in latin-1
_from_tsv(tsv_path)  # UnicodeDecodeError
```

End-to-end on a real OpenNeuro dataset:
```python
import openneuro

openneuro.download(
    dataset="ds004621",
    target_dir="/tmp/ds004621",
    include=["dataset_description.json", "participants.tsv",
             "sub-13/eeg/sub-13_task-rest_*"],
)
bp = BIDSPath(subject="13", task="rest", datatype="eeg",
              suffix="eeg", extension=".vhdr", root="/tmp/ds004621")
read_raw_bids(bp, on_ch_mismatch="rename")
```

## Expected results

The file is read successfully (encoding detected and applied), or a clear warning is emitted before falling back.

## Actual results

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb5 in position 25: invalid start byte` — raised deep in `numpy.loadtxt`, no recovery path for users.

## Proposed fix

Detect the encoding deterministically (BOM check + UTF-8 validity), then pass the detected encoding to `np.loadtxt`. Falls back to `latin-1` (which never raises) when UTF-8 decoding fails. No new dependency.

Fix in #1593.

## References

- Original report on fork: bruAristimunha/mne-bids#20
- Related upstream issue: #1530 (BOM in TSV files)
- Related upstream PR: #1531 (utf-8 instead of utf-8-sig)

bruAristimunha:
Closes #1594.

Real-world OpenNeuro datasets sometimes ship `channels.tsv` (and friends) in `latin-1` rather than UTF-8 — typically because of the `µ` sign in the `unit` column. `_from_tsv` hardcoded `utf-8-sig` and crashed.

Add a small deterministic encoding detector and feed the result straight into `np.loadtxt`:

```python
def _detect_file_encoding(fname):
    raw = Path(fname).read_bytes()
    if raw.startswith(codecs.BOM_UTF8):
        return "utf-8-sig"
    if raw.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
        return "utf-16"
    try:
        raw.decode("utf-8")
        return "utf-8"
    except UnicodeDecodeError:
        return "latin-1"
```

I think it could potentially be used as complemented of #1531 by @scott-huberty.

## Reproducer (end-to-end `read_raw_bids` on a real OpenNeuro dataset)

```python
import openneuro

openneuro.download(
    dataset="ds004621",
    target_dir="/tmp/ds004621",
    include=["dataset_description.json", "participants.tsv",
             "sub-13/eeg/sub-13_task-rest_*"],
)
bp = BIDSPath(subject="13", task="rest", datatype="eeg",
              suffix="eeg", extension=".vhdr", root="/tmp/ds004621")
raw = read_raw_bids(bp, on_ch_mismatch="rename")
print(f"OK — {len(raw.ch_names)} channels, {raw.n_times} samples")
```

| | result |
|--|--|
| `main` | `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb5 in position 25: invalid start byte` (deep in `numpy.loadtxt`) |
| this PR | `OK — 127 channels, 711840 samples` |

The offending bytes are `\xb5V` (`µV`) in the `units` column of `sub-13_task-rest_channels.tsv`. Same pattern affects `ds003620` (160 files), `ds005692` (59 files), `ds004598` (20 files), `ds005691` (8 iEEG files), and more.

Tests cover the 5 encoding cases (ASCII / UTF-8 / UTF-8-BOM / UTF-16 / latin-1) plus an integration test that verifies the warning fires and `µV` is read correctly.

### DIFF
--- a/mne_bids/tsv_handler.py
+++ b/mne_bids/tsv_handler.py
@@ -8,6 +8,7 @@ from collections import OrderedDict
 from copy import deepcopy
 
 import numpy as np
+from mne.utils import logger
 
 from mne_bids._fileio import _open_lock
 
@@ -172,7 +173,7 @@ def _from_tsv(fname, dtypes=None):
 
     encoding = _detect_file_encoding(fname)
     if not encoding.startswith("utf-8"):
-        warn(f"TSV file is not UTF-8 encoded, reading as latin-1: '{fname}'")
+        logger.info(f"Reading non-UTF-8 TSV as {encoding}: '{fname}'")
     data = np.loadtxt(
         fname, dtype=str, delimiter="\t", ndmin=2, comments=None, encoding=encoding
     )
--- a/mne_bids/tsv_handler.py
+++ b/mne_bids/tsv_handler.py
@@ -3,12 +3,28 @@
 # Authors: The MNE-BIDS developers
 # SPDX-License-Identifier: BSD-3-Clause
 
+import codecs
 from collections import OrderedDict
 from copy import deepcopy
+from pathlib import Path
 
 import numpy as np
 
 from mne_bids._fileio import _open_lock
+
+
+def _detect_tsv_encoding(fname):
+    """Deterministically detect the encoding of a TSV file."""
+    raw = Path(fname).read_bytes()
+    if raw.startswith(codecs.BOM_UTF8):
+        return "utf-8-sig"
+    if raw.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
+        return "utf-16"
+    try:
+        raw.decode("utf-8")
+        return "utf-8"
+    except UnicodeDecodeError:
+        return "latin-1"
 
 
 def _combine_rows(data1, data2, drop_column=None):
@@ -147,8 +163,11 @@ def _from_tsv(fname, dtypes=None):
     """
     from .utils import warn  # avoid circular import
 
+    encoding = _detect_tsv_encoding(fname)
+    if encoding == "latin-1":
+        warn(f"TSV file is not UTF-8 encoded, reading as latin-1: '{fname}'")
     data = np.loadtxt(
-        fname, dtype=str, delimiter="\t", ndmin=2, comments=None, encoding="utf-8-sig"
+        fname, dtype=str, delimiter="\t", ndmin=2, comments=None, encoding=encoding
     )
     # Handle empty files - data may be empty or only have a header
     if data.size == 0:
--- a/mne_bids/tsv_handler.py
+++ b/mne_bids/tsv_handler.py
@@ -13,15 +13,15 @@ import numpy as np
 from mne_bids._fileio import _open_lock
 
 
-def _detect_tsv_encoding(fname):
-    """Deterministically detect the encoding of a TSV file."""
-    raw = Path(fname).read_bytes()
-    if raw.startswith(codecs.BOM_UTF8):
+def _detect_file_encoding(fname):
+    """Deterministically detect the text encoding of a file."""
+    byte_content = Path(fname).read_bytes()
+    if byte_content.startswith(codecs.BOM_UTF8):
         return "utf-8-sig"
-    if raw.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
+    if byte_content.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
         return "utf-16"
     try:
-        raw.decode("utf-8")
+        byte_content.decode("utf-8")
         return "utf-8"
     except UnicodeDecodeError:
         return "latin-1"
@@ -163,7 +163,7 @@ def _from_tsv(fname, dtypes=None):
     """
     from .utils import warn  # avoid circular import
 
-    encoding = _detect_tsv_encoding(fname)
+    encoding = _detect_file_encoding(fname)
     if encoding == "latin-1":
         warn(f"TSV file is not UTF-8 encoded, reading as latin-1: '{fname}'")
     data = np.loadtxt(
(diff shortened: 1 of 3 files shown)


mne_bids/tsv_handler.py
```diff
@@ -147,8 +163,11 @@ def _from_tsv(fname, dtypes=None):
     """
     from .utils import warn  # avoid circular import
 
+    encoding = _detect_file_encoding(fname)
+    if encoding == "latin-1":
+        warn(f"TSV file is not UTF-8 encoded, reading as latin-1: '{fname}'")
     data = np.loadtxt(
-        fname, dtype=str, delimiter="\t", ndmin=2, comments=None, encoding="utf-8-sig"
+        fname, dtype=str, delimiter="\t", ndmin=2, comments=None, encoding=encoding
     )
```
larsoner, line 179: This will cause two full reads of the data for all files, which is a bummer. How about this pattern, which will be faster for properly formatted datasets:
```
try:
    data = np.loadtxt(..., encoding="utf-8-sig")
except UnicodeDecodeError:
    encoding = _detect...
    warn(f"TSV file is not UTF-8 encoded, attempting to read as {encoding}: {fname}")
    data = np.loadtxt(..., encoding=encoding)
```

mne_bids/tsv_handler.py
```diff
@@ -3,14 +3,30 @@
...
 from mne_bids._fileio import _open_lock
 
 
+def _detect_file_encoding(fname):
+    """Deterministically detect the text encoding of a file."""
+    byte_content = Path(fname).read_bytes()
+    if byte_content.startswith(codecs.BOM_UTF8):
+        return "utf-8-sig"
+    if byte_content.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
+        return "utf-16"
```
drammock: this branch could get away with only reading the first few bytes instead of the whole file. Worth considering doing this part using `file.read(n_bytes)`.

Of course I'd guess that the bulk of the examples in the wild do not have a BOM (?) so savings in the (rare) BOM case versus reading a few bytes twice in the (common) no-BOM case... 🤷🏻

mne_bids/tsv_handler.py
bruAristimunha: I think in this case, it would not be a problem to read twice, especially if I can check only a few bytes. The function can be potentially used in the PR #https://github.com/mne-tools/mne-bids/pull/1531

mne_bids/tsv_handler.py
bruAristimunha: I think we're gonna have this situation more often because of #https://github.com/mne-tools/mne-bids/pull/1531, and based on my experience with openneuro... it is better to infer with a small amount of bytes the encoding that assume everybody will follow the same system.

bruAristimunha:
See how you feel, @larsoner and @drammock

larsoner (approved):
Okay for me!

scott-huberty:
Sorry for the delay reviewing now.

bruAristimunha:
Heyyyy @scott-huberty! Supper don't worry, the team is merging many PRs here, so I am super happy about the speed :)

drammock (approved):

mne_bids/tsv_handler.py
```diff
@@ -147,8 +170,11 @@ def _from_tsv(fname, dtypes=None):
     """
     from .utils import warn  # avoid circular import
 
+    encoding = _detect_file_encoding(fname)
+    if encoding == "latin-1":
```
scott-huberty:
```suggestion
    if not encoding.startswith("utf-8"):
```

[BIDS specifies that tabular files must be utf-8](https://bids-specification.readthedocs.io/en/v1.5.0/02-common-principles.html#tabular-files), so might as well make the warning more broad

scott-huberty: But heads up that you will need to catch the warning thrown by`test_detect_file_encoding` id `"utf16"`

mne_bids/tsv_handler.py
```diff
@@ -147,8 +171,11 @@ def _from_tsv(fname, dtypes=None):
     """
     from .utils import warn  # avoid circular import
 
+    encoding = _detect_file_encoding(fname)
+    if not encoding.startswith("utf-8"):
+        logger.info(f"Reading non-UTF-8 TSV as {encoding}: '{fname}'")
```
scott-huberty, line 176: Sorry @bruAristimunha just to clarify, I wasn't suggesting that we demote the warning to a logger INFO message. But if you think that INFO is the right level for the message, then I am okay with it.

bruAristimunha: I think it would be better to reduce....

PR #1596, opened 2026-04-30
FIX: pass on_split_missing="warn" for split FIF files

### LINKED ISSUE #1595: BUG: read_raw_bids crashes on split FIF files when not all splits are present
## Description of the problem

`read_raw_bids` crashes when a recording is split across multiple FIF files (FIF has a 2 GB per-file limit) and any later split is missing. The bug surfaces in real-world scenarios:

- **Partial OpenNeuro downloads** — users filter `include="**/*split-01*"` to grab a quick sample, or scripts crash mid-fetch.
- **DataLad / git-annex without `get`** — symlinks present (mne-bids glob finds them), but only some annex blobs were fetched.
- **Tiered storage migration** — large datasets sometimes have later splits moved to cold storage.

MNE-Python has supported `on_split_missing="warn"` since v0.22 to load the available split and warn about the missing one. `read_raw_bids` does not opt into it. There is a long-standing `FIXME` in `mne_bids/path.py:928` noting this gap.

## Affected datasets (real-world)

A scan of 57 OpenNeuro MEG datasets found 4 that ship split FIF data: `ds004998`, `ds005241`, `ds005261`, `ds005356` (300 split files / 150 base paths). Whenever a user pulls only `split-01_meg.fif` (without `split-02_meg.fif`) from any of these, `read_raw_bids` hard-crashes.

## Steps to reproduce

```python
import s5cmd_or_aws_or_openneuro_py  # download just split-01 + sidecars

bp = BIDSPath(subject="13", task="gloups", run="03",
              datatype="meg", root="/tmp/ds005261")
read_raw_bids(bp)
# ValueError: Split raw file detected but next file
# /tmp/ds005261/sub-13/meg/sub-13_task-gloups_run-03_split-02_meg.fif
# does not exist...
```

## Expected results

`read_raw_bids` loads the available split(s) and emits a `RuntimeWarning` about the missing one, rather than raising `ValueError`. Users keep the strict behavior with `extra_params={"on_split_missing": "raise"}`.

## Actual results

`ValueError: Split raw file detected but next file ... does not exist.` — raised by MNE's FIF reader, no recovery path inside `read_raw_bids`.

## Proposed fix

Pass `on_split_missing="warn"` as a default for `.fif` files inside `read_raw_bids`, alongside the existing `allow_maxshield=True` default. Also drop the FIXME in `path.py`.

Fix in #1596.

## References

- Original report on fork: bruAristimunha/mne-bids#19
- Related FIXME: `mne_bids/path.py:928` (now resolved)

bruAristimunha:
Closes #1595.

Real-world large MEG recordings ship as multiple FIF files (FIF has a 2 GB per-file limit). Whenever later splits are absent — partial downloads, git-annex without `get`, tiered storage — `read_raw_bids` raises `ValueError: Split raw file detected but next file ... does not exist`. MNE-Python has supported `on_split_missing="warn"` since v0.22 specifically for this case; this PR opts into it.

```python
if raw_path.suffix == ".fif":
    if "allow_maxshield" not in extra_params:
        extra_params["allow_maxshield"] = True
    if "on_split_missing" not in extra_params:
        extra_params["on_split_missing"] = "warn"
```

Also drops the long-standing `FIXME` in `mne_bids/path.py:928` ("This will break ... e.g. with FIFF data split across multiple files"). Users keep the strict behavior with `extra_params={"on_split_missing": "raise"}`.

## Reproducer (real OpenNeuro dataset)

A scan of 57 OpenNeuro MEG datasets found 4 ship split FIF data: `ds004998`, `ds005241`, `ds005261`, `ds005356` (300 files across 150 base paths). The repro below uses `ds005261/sub-13/run-03` and intentionally simulates a partial download by deleting `split-02`:

```python
from pathlib import Path

# Pre-downloaded /tmp/ds005261/sub-13/meg/sub-13_task-gloups_run-03_*
# from s3://openneuro.org/ds005261/. Delete split-02 to simulate a
# partial download.
Path("/tmp/ds005261/sub-13/meg/"
     "sub-13_task-gloups_run-03_split-02_meg.fif").unlink()

bp = BIDSPath(subject="13", task="gloups", run="03",
              datatype="meg", root="/tmp/ds005261")
raw = read_raw_bids(bp)
print(f"OK — n_times={raw.n_times} ({raw.times[-1]:.1f}s, "
      f"{len(raw.ch_names)} ch)")
```

| | result |
|--|--|
| `main` | `ValueError: Split raw file detected but next file ... does not exist` |
| this PR | `OK — n_times=2163205 (1063.3s, 248 ch)` (with a `RuntimeWarning` about the missing split) |

A self-contained unit test (`test_read_raw_bids_split_fif_missing_warns`) reproduces this with a synthetic ~16 MB Raw split at 5 MB into 4 splits, then deletes the last one.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -67,7 +67,7 @@ Detailed list of changes
 - Fix bug with :meth:`mne_bids.BIDSPath.find_matching_sidecar` not searching parent directories properly, by `Eric Larson`_ (:gh:`1565`)
 - Fix :func:`mne_bids.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
 - Fix :func:`mne_bids.events_file_to_annotation_kwargs` to fall back to the ``value`` column when ``trial_type`` is entirely ``n/a`` but ``value`` contains trigger codes, instead of dropping all events, by `Bruno Aristimunha`_ (:gh:`947`)
-- Pass ``on_split_missing="warn"`` by default for split FIF files so :func:`mne_bids.read_raw_bids` warns and returns the available data instead of crashing on partial downloads, by `Bruno Aristimunha`_
+- Document that :func:`mne_bids.read_raw_bids` accepts ``extra_params=dict(on_split_missing="warn")`` (or ``"ignore"``) so users can opt in to loading split FIF recordings whose later splits are missing on disk (e.g. partial downloads, tiered storage); the default behavior is preserved, by `Bruno Aristimunha`_
 
 ⚕️ Code health
 ^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


larsoner:
Hmm given this is more likely a download rather than dataset curation/creation error -- or a particularly bad data creation error (data will be missing!) -- I am inclined to have the default be `"raise"`. And if people want to opt in to having missing data, they should pass `extra_params=dict(on_split_missing="ignore" | "warn")`.

bruAristimunha:
okay @larsoner, make sense your point, i will close here.

PR #1597, opened 2026-05-01
FIX: write *_electrodes.json with SpatialReference for derivative datasets

### LINKED ISSUE #1545: BUG: Missing *_electrodes.json sidecar causes BIDS validator failure
## Description of the problem

`write_raw_bids()` writes `*_electrodes.tsv` (with `space-CapTrak` entity) but does **not** create the companion `*_electrodes.json` sidecar. BIDS validator v2.4.0+ requires a `SpatialReference` key in that JSON sidecar whenever a `space` entity is present.

This causes a BIDS validator error on any dataset with digitized electrode positions.

## Steps to reproduce

```python
import mne
from mne_bids import write_raw_bids, BIDSPath

# Create sample raw with standard montage
raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() / "MEG" / "sample" / "sample_audvis_raw.fif")
raw.pick_types(eeg=True)
raw.set_montage("standard_1020")

bids_path = BIDSPath(subject="01", task="test", root="/tmp/bids_test")

# Check: electrodes.json should exist but doesn't
import glob
print(glob.glob("/tmp/bids_test/**/*electrodes*", recursive=True))
# Only *_electrodes.tsv and *_coordsystem.json are created — no *_electrodes.json
```

Then run:
```bash
bids-validator /tmp/bids_test
```

## Expected results

A `*_electrodes.json` file should be created next to `*_electrodes.tsv` containing at minimum:

```json
{
    "SpatialReference": "n/a"
}
```

For standard scalp EEG with CapTrak coordinates (defined by nasion, LPA, RPA landmarks), there is no external template image, so `SpatialReference` should be `"n/a"`.

## Actual results

No `*_electrodes.json` is created. BIDS validator v2.4.0+ reports an error about the missing `SpatialReference` field.

## Proposed fix

In `_write_dig_bids()` (in `mne_bids/dig.py`), after writing `*_electrodes.tsv`, also create the companion `*_electrodes.json` sidecar with `{"SpatialReference": "n/a"}` whenever a `space` entity is present and the coordinate system is CapTrak (or any head-based system without a template image).

## Additional information

- BIDS validator version: 2.4.0+
- Originally discussed at https://github.com/bruAristimunha/mne-bids/issues/11 with feedback from @larsoner and @scott-huberty

bruAristimunha:
Closes #1545.

Derivative datasets with a `space` entity fail the BIDS validator because the required `*_electrodes.json` (with `SpatialReference`) isn't written. This adds it and infers the value from the coordinate frame: template name for standard templates, a `bids::` URI to the subject T1w for `ACPC`/`ScanRAS`/`Pixels`, and `"n/a"` for head-based frames.

## Reproducer

```python
from mne_bids import write_raw_bids, BIDSPath, make_dataset_description

bp = BIDSPath(subject="01", task="test", root="/tmp/x", datatype="eeg")
write_raw_bids(raw, bp, overwrite=True)
make_dataset_description(path="/tmp/x", name="x", dataset_type="derivative", overwrite=True)
```

| | `deno -A jsr:@bids/validator /tmp/x` |
|--|--|
| `main` | `[ERROR] SIDECAR_KEY_REQUIRED: SpatialReference` |
| this PR | 0 errors |

### DIFF
--- a/mne_bids/dig.py
+++ b/mne_bids/dig.py
@@ -738,6 +738,10 @@ def _write_dig_bids(
         overwrite=overwrite,
     )
     if bids_path.datatype != "nirs":
+        # NIRS uses *_optodes.{tsv,json}; for EEG/iEEG the BIDS spec defines
+        # a *_electrodes.json sidecar whose SpatialReference field the
+        # validator requires for derivative datasets (see #1545):
+        # https://bids-specification.readthedocs.io/en/stable/modality-specific-files/electroencephalography.html#electrodes-description-_electrodestsv
         electrodes_json_path = BIDSPath(
             **electrode_file_entities, suffix="electrodes", extension=".json"
         )
--- a/mne_bids/dig.py
+++ b/mne_bids/dig.py
@@ -25,11 +25,13 @@ from mne_bids.config import (
     ALLOWED_SPACES,
     BIDS_COORD_FRAME_DESCRIPTIONS,
     BIDS_COORDINATE_UNITS,
+    BIDS_IEEG_COORDINATE_FRAMES,
     BIDS_STANDARD_TEMPLATE_COORDINATE_SYSTEMS,
     BIDS_TO_MNE_FRAMES,
     MNE_FRAME_TO_STR,
     MNE_STR_TO_FRAME,
     MNE_TO_BIDS_FRAMES,
+    coordsys_standard_template_deprecated,
 )
 from mne_bids.path import BIDSPath
 from mne_bids.tsv_handler import _from_tsv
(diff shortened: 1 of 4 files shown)


mne_bids/dig.py
```diff
@@ -688,6 +737,15 @@ def _write_dig_bids(
         datatype=bids_path.datatype,
         overwrite=overwrite,
     )
+    if bids_path.datatype != "nirs":
```
larsoner, line 740: A comment here would be helpful to say why and/or link to spec

bruAristimunha: should be fine now @larsoner :)

larsoner:
Thanks @bruAristimunha !

Issue #1519, opened 2026-02-12
Case mismatches in entity values between data files and sidecars in real-world datasets


bruAristimunha:
### Context

While processing ~200 OpenNeuro datasets with an automated pipeline, we encountered datasets where data files and BIDS sidecar files have case mismatches in entity values. For example, in [ds003751](https://openneuro.org/datasets/ds003751):

- Data file: `sub-01_task-Emotion_eeg.set`
- Sidecar: `sub-01_task-emotion_eeg.json`

Since BIDS entity matching is case-sensitive (correctly per the spec), strict matching cannot associate these files with their sidecars.

### This is not a feature request

We understand the BIDS specification treats entity values as case-sensitive and that MNE-BIDS correctly enforces this. We are **not** requesting case-insensitive matching — that would mask data quality issues.

We wanted to document this as a real-world pattern and ask two questions:

1. Is this a known issue with specific OpenNeuro datasets? Are there existing recommendations for handling this during automated processing of many datasets?
2. Would it be useful for functions like `find_matching_paths()` to optionally emit a warning when near-matches exist that differ only in case? For example: *"Found 'task-emotion_eeg.json' which differs from 'task-Emotion' only in case — possible sidecar mismatch."*

### Additional context

- MNE-BIDS version: 0.18.0
- Dataset: [ds003751](https://openneuro.org/datasets/ds003751) on OpenNeuro
- We work around this in our pipeline with case-insensitive comparison, but thought the real-world occurrence was worth reporting in case it's useful for the project.

bruAristimunha:
I have systematically contacted the authors and even contacted the openneuro itself about one dataset that wasn't even a valid format (*.tar instead of edf), for example. What I've noticed is that the authors ignore it or say they will fix it and never do. 

Many of the errors I've encountered are extremely simple to resolve, but I have little hope that the authors will be able to fix them.

Maybe we can run something that use mne bids as validator within open neuro, and fix everything that needs to be fix.

My feeling is that they correct several MRI and fMRI datasets according to the issues I found, but the EEG/IEEG/MEG part is more complicated to have that iterability, so several of these errors prevent the datasets from being bid on.

larsoner:
@bruAristimunha if you get on a roll fixing datasets and want extra credit... MNE-BIDS-Pipeline has a few workarounds for datasets we test with as well

[...]

PR #1601, opened 2026-05-04
FIX: allow non-numeric run values in BIDSPath if check=False

bruAristimunha:
One-character fix in `mne_bids/path.py`: gate the "run is not an index" check on `self.check`, so `BIDSPath(..., run="5H", check=False)` succeeds while the strict default still rejects.

```diff
-                if isinstance(val, str) and not val.isdigit():
+                if isinstance(val, str) and not val.isdigit() and self.check:
                     raise ValueError(f"{key} is not an index (Got {val})")
```

This matches the existing `check=False` escape hatch already used for `suffix`, `space`, `datatype`, and `extension`.

## Real-world dataset blocked today

[`ds003190`](https://openneuro.org/datasets/ds003190) ships VHDR files with `run-5F` and `run-5H` entities. eegdash hit this and silently dropped 104 files via a workaround ([eegdash#263](https://github.com/eegdash/EEGDash/pull/263), later extended in [eegdash#277](https://github.com/eegdash/EEGDash/pull/277)).

```python
from mne_bids import BIDSPath

BIDSPath(subject="01", run="5H")
# main: ValueError: run is not an index (Got 5H)

bp = BIDSPath(subject="01", run="5H", check=False)
# this PR: bp.basename == "sub-01_run-5H"
```

|        | result                                              |
|--------|-----------------------------------------------------|
| `main` | `ValueError: run is not an index (Got 5H)`          |
| this PR| `BIDSPath(...).basename == "sub-01_run-5H"`         |

## End-to-end `read_raw_bids` works

The fix carries through to `read_raw_bids`. Synthetic dataset with `run-5H` files on disk:

```python

bp = BIDSPath(subject="01", task="t", run="5H", datatype="eeg",
              suffix="eeg", extension=".edf", root=root, check=False)
raw = read_raw_bids(bp)
# OK: 139 channels, 3072 samples, channels.tsv sidecar resolved correctly
```

`_find_matching_sidecar` is unaffected — it already sets `check=False` internally on the bids_path copy it uses.

## Tests

Folded into existing `test_path.py` tests rather than a new file:

- `test_make_filenames`: positive (`check=False`) + negative (default raises) for `run="5H"`.
- `test_bids_path`: positive + negative for `datatype="fnirs"` (covering the same `check=False` escape hatch already used by some real-world datasets that ship `fnirs/` instead of the spec's `nirs/`).

Strict default behavior is unchanged.

## Related

Part of the broader effort to read non-spec OpenNeuro datasets (#1591, #1593, #1599). The other items in the table at [#1519](https://github.com/mne-tools/mne-bids/issues/1519) (case-mismatched entities, separator typos, fuzzy participant_id) are matching/parsing concerns, not BIDSPath construction — those need separate PRs.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -70,6 +70,7 @@ Detailed list of changes
 - Fix :func:`mne_bids.events_file_to_annotation_kwargs` to drop rows with invalid ``onset`` values (``n/a``, ``nan``, ``NaN``, empty string) before float conversion. This makes reading real-world OpenNeuro datasets (e.g. ``ds004841``, ``ds004842``, ``ds004843``) succeed instead of either raising or returning ``NaN`` onsets, by `Bruno Aristimunha`_ (:gh:`1547`)
 - Fix :func:`mne_bids.events_file_to_annotation_kwargs` to fall back to the ``value`` column when ``trial_type`` is entirely ``n/a`` but ``value`` contains trigger codes, instead of dropping all events, by `Bruno Aristimunha`_ (:gh:`947`)
 - :func:`mne_bids.read_raw_bids` now tolerates malformed ``scans.tsv`` entries and ISO 8601 ``acq_time`` variants, by `Bruno Aristimunha`_ (:gh:`1591`)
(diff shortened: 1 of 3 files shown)


larsoner (approved):
Seems straightforward, thanks @bruAristimunha !

PR #1602, opened 2026-05-04
ENH: add keywords parameter to make_dataset_description

### LINKED ISSUE #1549: ENH: Add Keywords parameter to make_dataset_description()
## Describe the problem

`make_dataset_description()` accepts a fixed set of keyword arguments (`name`, `authors`, `doi`, `funding`, `acknowledgements`, etc.) but does not support `Keywords`, which is a valid BIDS field in `dataset_description.json`. Users who want to include keywords must manually patch the JSON after calling `make_dataset_description()`.

Related: #302

## Describe your solution

Add a `keywords` parameter to `make_dataset_description()`:

```python
def make_dataset_description(
    path,
    name=None,
    ...,
    keywords=None,  # NEW: list of str
    ...,
):
```

When provided, `keywords` should be written as the `"Keywords"` field in `dataset_description.json`.

## Describe possible alternatives

1. **Manual JSON patching** (current workaround) — read, modify, write the JSON after `make_dataset_description()`.
2. **Only add `extra_fields`** — a single generic parameter instead of explicitly adding `keywords`.

## Current downstream workaround (MOABB)

```python
mne_bids.make_dataset_description(path=str(root), overwrite=True, **desc_kwargs)

# Then patch in Keywords manually
desc_path = Path(root) / "dataset_description.json"
with open(desc_path) as f:
    desc = json.load(f)
if keywords and "Keywords" not in desc:
    desc["Keywords"] = keywords
    with open(desc_path, "w") as f:
        json.dump(desc, f, indent="\t")
```

## Additional context

- `Keywords` is defined in the BIDS specification: https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#dataset-description
- Originally discussed at https://github.com/bruAristimunha/mne-bids/issues/15 with feedback from @drammock

bruAristimunha:
Closes #1549.

Add a ``keywords`` parameter to ``make_dataset_description`` so users can populate the BIDS ``Keywords`` field directly instead of post-patching ``dataset_description.json``.

```python
make_dataset_description(
    path=root, name="my-dataset",
    keywords=["eeg", "motor-imagery"],  # or "eeg, motor-imagery"
)
```

Accepts list / comma-separated str / ``None``, matching the neighbouring list-valued params.

### DIFF
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1598,9 +1598,10 @@ def make_dataset_description(
         List of references to publication that contain information on the
         dataset, or links.  Must be a list of str (e.g., ['a', 'b', 'c'])
         or a single comma-separated str (e.g., 'a, b, c').
-    keywords : list
+    keywords : list | str | None
         List of keywords describing the dataset (BIDS ``Keywords`` field).
-        Must be a list of str (e.g., ``['eeg', 'motor-imagery']``).
+        Must be a list of str (e.g., ``['eeg', 'motor-imagery']``) or a single
+        comma-separated str (e.g., ``'eeg, motor-imagery'``).
     doi : str | None
         The Digital Object Identifier of the dataset (not the corresponding
         paper). Must be of the form ``doi:<insert_doi>`` (e.g.,
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1602,6 +1602,8 @@ def make_dataset_description(
         List of keywords describing the dataset (BIDS ``Keywords`` field).
         Must be a list of str (e.g., ``['eeg', 'motor-imagery']``) or a single
         comma-separated str (e.g., ``'eeg, motor-imagery'``).
+
+        .. versionadded:: 0.19
     doi : str | None
         The Digital Object Identifier of the dataset (not the corresponding
         paper). Must be of the form ``doi:<insert_doi>`` (e.g.,
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1550,7 +1550,7 @@ def make_dataset_description(
     funding=None,
     ethics_approvals=None,
     references_and_links=None,
-    keywords=(),
+    keywords=None,
     doi=None,
     generated_by=None,
     source_datasets=None,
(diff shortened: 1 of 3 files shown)


mne_bids/write.py
```diff
@@ -1597,6 +1598,10 @@ def make_dataset_description(
         List of references to publication that contain information on the
         dataset, or links.  Must be a list of str (e.g., ['a', 'b', 'c'])
         or a single comma-separated str (e.g., 'a, b, c').
+    keywords : list | str | None
+        List of keywords describing the dataset (BIDS ``Keywords`` field).
+        Must be a list of str (e.g., ``['eeg', 'motor-imagery']``) or a single
+        comma-separated str (e.g., ``'eeg, motor-imagery'``).
```
larsoner, line 1604: I wouldn't allow the comma-separated `str`. There should be one clear/obvious way of doing things, when possible/reasonable, and `.split` on strings is easy enough.
```suggestion
    keywords : list
        List of keywords describing the dataset (BIDS ``Keywords`` field).
        Must be a list of str (e.g., ``['eeg', 'motor-imagery']``).
```
default should be `()` (we allow `tuple` when we say `list`, most of the time)

mne_bids/write.py
```diff
@@ -1550,6 +1550,7 @@ def make_dataset_description(
     funding=None,
     ethics_approvals=None,
     references_and_links=None,
+    keywords=None,
```
larsoner, line 1553: Designing the API from scratch I would do this... but I guess since we have `None` everywhere already I guess this could stay `list | None`
```suggestion
    keywords=(),
```

larsoner (approved):
Ugh actually never mind, I we already allow comma-separated `str` so this should, too

larsoner:
@bruAristimunha sorry I should have read more closely to see we already allow comma-separated `str`, if you revert those commits I'll merge :facepalm:

bruAristimunha:
reverting...

bruAristimunha:
hey @larsoner, don't worry ;)

PR #1603, opened 2026-05-04
FIX: more generous channels.tsv reading

bruAristimunha:
Two friction points in `_handle_channels_reading` when reading real-world OpenNeuro datasets, plus a new opt-in mismatch mode.

- **Empty / headerless `channels.tsv`** — used to raise `KeyError: 'name'`. Now warns and skips channel metadata, same as if the sidecar were absent.
- **Duplicate channel names in `channels.tsv`** — used to raise `ValueError: New channel names are not unique`. Now treated as a sub-type of channel mismatch and routed through `on_ch_mismatch`:
  - `"raise"` (default): raises with a pointer to the rename/warn options.
  - `"rename"`: deduplicates by appending `-0/-1/...` suffixes so the rename succeeds.
  - `"warn"`: warns and skips channels.tsv-derived metadata, leaves `raw` untouched.
  - `"reorder"`: raises (reorder is ambiguous with duplicates).
- **New `on_ch_mismatch="warn"`** for the general name-mismatch case: warns and skips channels.tsv-derived metadata instead of raising.

```python
# before
KeyError: 'name'
ValueError: New channel names are not unique, renaming failed

# after (defaults unchanged: still raises on duplicates / mismatch)
read_raw_bids(..., on_ch_mismatch="rename")  # dedupes -> renames
read_raw_bids(..., on_ch_mismatch="warn")    # warns -> skips metadata
```

Tests cover all three branches (rename/raise/warn) for duplicates, plus the empty / wrong-header paths and the new `warn` mode for ordinary mismatches. Added next to the existing `test_channel_mismatch_*` family in `test_read.py` — no new test file.


bruAristimunha:
ping to not forget this PR too @scott-huberty ;)

mne_bids/read.py
```diff
@@ -1024,17 +1031,27 @@ def _handle_channel_mismatch(raw, on_ch_mismatch, ch_names_tsv, channels_fname):
...
 
     Updates status (bad) and types of channels.
     """
     logger.info(f"Reading channel info from {channels_fname}.")
     channels_dict = _from_tsv(channels_fname)
+    if not channels_dict.get("name"):
+        warn(
+            f"{channels_fname} is empty or has no 'name' column; "
+            "skipping channel metadata."
+        )
```
larsoner: This will emit two warnings for an empty file, right? That seems unnecessary... so how about
```suggestion
    if not channels_dict.get("name"):
        if len(channels_dict):
            warn(
                f"{channels_fname} has no 'name' column; "
                "skipping channel metadata."
            )
```

mne_bids/read.py
```diff
@@ -1165,7 +1184,7 @@ def read_raw_bids(
     extra_params=None,
     *,
     return_event_dict=False,
-    on_ch_mismatch="raise",
+    on_ch_mismatch="warn",
```
larsoner: I'd rather keep the current default, it's safer. Then if people really want they can set to `warn` or `rename`

mne_bids/read.py
```diff
@@ -13,6 +13,7 @@
 import mne
 import numpy as np
 from mne import events_from_annotations, io, pick_channels_regexp, read_events
+from mne._fiff.meas_info import _unique_channel_names
```
larsoner: Ideally we wouldn't use any private functions from MNE... can we get away with something simpler? A little helper here?

drammock:
not sure I'm comfortable with calling the dedup function all the time.  Can we treat duplicate ch names in channels.tsv as a sub-type of "mismatch", and *only* call the dedup function for certain values of `on_ch_mismatch`? E.g. maybe dedup only happens if `on_ch_mismatch="rename"`

bruAristimunha:
should be fine now @drammock :)

mne_bids/read.py
```diff
@@ -1024,7 +1031,32 @@ def _handle_channel_mismatch(raw, on_ch_mismatch, ch_names_tsv, channels_fname):
...
         raw.rename_channels(dict(zip(raw.ch_names, ch_names_tsv)))
     else:
-        raise ValueError("on_ch_mismatch must be one of {'reorder','raise','rename'}")
+        raise ValueError(
+            "on_ch_mismatch must be one of {'reorder','raise','rename','warn'}"
+        )
+
+
+def _dedupe_channel_names(ch_names):
+    """Rename duplicate channel names in place ."""
```
scott-huberty, line 1040:
```suggestion
    """Rename duplicate channel names in place."""
```

mne_bids/read.py
```diff
@@ -1024,7 +1031,32 @@ def _handle_channel_mismatch(raw, on_ch_mismatch, ch_names_tsv, channels_fname):
...
+    positions = defaultdict(list)
+    for i, name in enumerate(ch_names):
+        positions[name].append(i)
+    dups = {name for name, idxs in positions.items() if len(idxs) > 1}
+    if not dups:
+        return
+    warn(
```
scott-huberty, line 1047: @bruAristimunha and @drammock - this function gets called when 1) There are duplicate channel names in channels.tsv and 2) the user passes `on_ch_mismatch='rename'` (other valid arguments are `'warn'`, `'raise'`, or `'reorder'`).  

This makes me assume that when the user passes `'rename'`, they are expressing that they don't want a warning, yet we will throwing one anyway. Should we change this to an INFO statement or keep it as a warning?

mne_bids/read.py
```diff
@@ -1111,8 +1163,10 @@ def _handle_channels_reading(channels_fname, raw, on_ch_mismatch="raise"):
         )
     else:
         orig_names = list(raw.ch_names)
-        if orig_names != ch_names_tsv:
-            _handle_channel_mismatch(raw, on_ch_mismatch, ch_names_tsv, channels_fname)
+        if orig_names != ch_names_tsv and _handle_channel_mismatch(
+            raw, on_ch_mismatch, ch_names_tsv, channels_fname
```
scott-huberty, line 1167:
```suggestion
        if raw.ch_names != ch_names_tsv and _handle_channel_mismatch(
            raw, on_ch_mismatch, ch_names_tsv, channels_fname
```

Why did we ever do `orig_names = list(raw.ch_names)` ? Isn't `raw.ch_names` guaranteed to be a list?

drammock (approved):

drammock:
thanks @bruAristimunha

scott-huberty (approved):
It's Monday morning for me but, but the process of reading the `channels.tsv` is getting hard for me to reason about.

Reconciliation between raw channel names and `channels.tsv` happens in multiple places: Deduplication is handled in `_handle_channels_reading`, mismatches between `channels.tsv` and `raw.ch_names` happens in `_handle_channels_mismatch`.. Except for the special-case mismatch of `STI 014` being present in `raw` but not `channels.tsv`.. That is handled in `_handle_channels_reading`..

Also, now, `_handle_channels_mismatch is True` means "skip setting channels metadata on the raw object", which isn't very intuitive..

It might be worthwhile to centralize all this data munging in a helper function that is called very early in `_handle_channels_reading`. Then the rest of `_handle_channels_reading` can focus on applying channel types/units and `'bads'` etc.

Perhaps the MNE sprint this summer will be a good time to do such a refactor.

scott-huberty:
Oops - disregard. I did not see that this got merged mid-review.

Issue #1176, opened 2023-10-11
Feature request: Support of derivatives


christian-oreilly:
### Describe the problem

I am currently writing a grant proposal where I will suggest a aim that looks like this (the wording will probably change between now and the submission, but the idea should remain more or less the same):

Aim 1: Improve MNE-BIDS and EEGNET to support derivatives BIDS datasets of EEG/MEG source data. We will extend MNE to ease the creation and automated analyses of shared neurophysiological datasets. Particularly, we will focus on adding support for derivatives datasets, including EEG source data. We will demonstrate this integration in the context of EEGNET web platforms EEGNET and the PyLossless pipeline.

### Describe your solution

For this grant, and for our PyLossless pipeline (https://github.com/lina-usc/pylossless), I would be interested in MNE-BIDS supporting more transparently BIDS derivatives, including the writing/loading of derivatives in subdirectories as outlined here: https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#derived-dataset-and-pipeline-description

I would also be interested in supporting more transparently BIDS mechanisms for storing EEG/MEG source data derivatives. I understand that this may also require, for consistency, discussion with the BIDS project to make sure that the norm itself supports these data, not just that software like MNE-BIDS supports an ad hoc version of it. 

To be clear, this feature request ticket is not for one of the MNE-BIDS developers to write such a PR; the grant would support the salary for someone to contribute to this on my side. I would just want, at this step, to initiate a discussion so that I can know your level of interest in integrating such functionalities in MNE-BIDS and get your opinion about the best way to integrate these features.

### Describe possible alternatives

None.

agramfort:
awesome initiative !

so far the mne-bids-pipeline has hacked his way for derivatives and it's
not done in a standardized way at all.
I think it makes your project very relevant to move forward here.

as you said it's important to coordinate this with the wider EEG / MEG BIDS
folks to make it adopted beyond
mne-bids

hoechenberger:
I'd like to second what Alex said: this is a great initiative! 👍👍👍

sappelhoff:
sounds wonderful!

bruAristimunha:
Can I take care of this? I need it because I have some datasets at OpenNeuro that are saved as epochs. 

| Dataset ID |
|---|
| ds003805 |
| ds004785 |
| ds004368 |
| ds004771 |
| ds004554 |
| ds004520 |
| ds004521 |
| ds007081 |
| ds004519 |
[...]

PR #1605, opened 2026-05-04
ENH: Add read_epochs_bids for pre-epoched EEGLAB datasets

bruAristimunha:
Closes (in part) #1176.

## Summary

Adds `mne_bids.read_epochs_bids()` for reading BIDS datasets whose data sidecar JSON declares `RecordingType: \"epoched\"`. EEGLAB `.set` is supported in this first PR — that covers every OpenNeuro epoched dataset I'm aware of.

`read_raw_bids()` now peeks at `RecordingType` and raises an informative `RuntimeError` pointing to the new function rather than silently mis-loading epoched data as `Raw`.

The shared sidecar attachment (channels / electrodes / coordsystem / info / scans / participants) is factored into a private `_attach_sidecars()` used by both readers, so the new public function stays small.

## Why no derivatives layout?

The BIDS EEG spec already allows `RecordingType: \"epoched\"` in the main subject tree (not under `derivatives/`); every OpenNeuro example I tested uses that layout. Discussion on #1176 (Arno + drammock) confirmed this is spec-compliant. The broader \"derivatives subdir\" story Christian raised in the original issue is orthogonal and deferred.

## Real-world validation

Pulled one subject from each of the 13 OpenNeuro datasets [Bru listed in #1176](https://github.com/mne-tools/mne-bids/issues/1176#issuecomment-3137820800) and ran the new reader end-to-end:

| Dataset | Result |
|---|---|
| ds003805 | ❌ upstream MNE (`IndexError` at `eeglab.py:705`) |
| ds004785 | ✅ 48 ep · 32 ch · 500 Hz |
| ds004368 | ❌ upstream MNE (`AttributeError: 'Bunch' object has no attribute 'eventtype'`) |
| ds004771 | ✅ 480 ep · 34 ch · 256 Hz |
| ds004554 | ✅ 270 ep · 99 ch · 1000 Hz |
| ds004520 | ✅ 784 ep · 62 ch · 250 Hz |
| ds004521 | ✅ 896 ep · 62 ch · 250 Hz |
| ds007081 | ❌ upstream MNE (`IndexError` at `eeglab.py:705`) |
| ds004519 | ✅ 864 ep · 62 ch · 250 Hz |
| ds005946 | ✅ 336 ep · 60 ch · 1000 Hz |
| ds006367 | ✅ 562 ep · 30 ch · 1000 Hz |
| ds006370 | ✅ 1280 ep · 30 ch · 1000 Hz |
| ds004166 | ❌ upstream scipy (`buffer is too small`; matches mne-tools/mne-python#10012) |

**9/13 read end-to-end including all sidecars.** Confirmed the 4 failures by re-running each on a second subject and via bare `mne.io.read_epochs_eeglab` — they fail identically without `mne-bids` involved, so they need MNE-side fixes.

The validation surfaced one real bug in this PR: `_handle_scans_reading` calls `inst.set_meas_date(None)`, and MNE's implementation pokes `inst.annotations._orig_time` — which is `None` on `Epochs` by default. Added a one-line guard in `_attach_sidecars`.

## Followups (separate PRs)

- `.fif` epoched reader (one line in the dispatch table once a fixture exists).
- Segmented epoching of EDF/BDF/BrainVision via the `EpochLength` sidecar field.
- Per-epoch events.tsv → `epochs.metadata` mapping (currently we keep MNE's EEGLAB-derived events).

## Test plan
- [x] `pytest mne_bids/tests/test_read_epochs.py` (new, 1 test, exercises round-trip + the `read_raw_bids` rejection)
- [x] `pytest mne_bids/tests/test_read.py` (75 existing tests still pass — `_attach_sidecars` is byte-equivalent to the original sequence, just hoisted into a helper)
- [x] Manual validation against 13 OpenNeuro datasets (table above)


mne_bids/read.py
```diff
@@ -1442,15 +1469,61 @@ def read_raw_bids(
...
+def read_epochs_bids(
+    bids_path, extra_params=None, *, on_ch_mismatch="raise", verbose=None
+):
+    """Read pre-epoched BIDS data (RecordingType="epoched") as :class:`mne.Epochs`.
+
+    EEGLAB ``.set`` only for now. The same sidecars as
+    :func:`read_raw_bids` are applied (channels, electrodes, coordsystem,
+    scans, participants). For continuous formats flagged ``"epoched"`` in
+    the sidecar (``.edf`` / ``.bdf`` / ``.vhdr``), :func:`read_raw_bids`
+    will load them as :class:`mne.io.Raw`; segment them yourself.
```
```suggestion
    Currently, only EEGLAB ``.set`` files are supported.  For continuous formats flagged ``"epoched"`` in
    the sidecar (``.edf`` / ``.bdf`` / ``.vhdr``), please use :func:`read_raw_bids`
    to load them as :class:`mne.io.Raw` and. then segment them yourself.
```

mne_bids/read.py
```diff
@@ -1442,15 +1469,61 @@ def read_raw_bids(
...
+        Forwarded to the underlying MNE reader.
+    on_ch_mismatch : str
+        See :func:`read_raw_bids`.
+    %(verbose)s
+
+    Returns
+    -------
+    epochs : mne.Epochs
+        The epoched data with BIDS sidecar metadata applied.
+    """
```
scott-huberty, line 1599:
```suggestion
        The epoched data with BIDS sidecar metadata applied.
    """
    
    Notes
    ------
    This function will apply the same sidecars as :func:`read_raw_bids` (``channels.tsv``, ``electrodes.tsv``, ``coordsystem.json``,
    ``scans.tsv``, ``participants.tsv``).
```

scott-huberty: Sorry there docstrings edits will surely cause some Ruff or numpydoc / "pydocstyle" flakes!

scott-huberty (commented):
Looks pretty good to me! First pass. @drammock let's look at this tomorrow, merge it if it looks good, and cut a release?

drammock:
My take on this is that it should wait until after the release. If we're going to implement support for reading epoched data, we should do it for all datatypes (not just EEG) and all file formats that support storing epoched data (i.e., FIF, Brainvision, EDF/BDF at least, IIUC).

larsoner:
Okay for me! And release process hopefully isn't too onerous so we can cut another quickly if more complete support is added quickly

bruAristimunha:
We would need to implement the other epoch readers first in MNE, and I think it would not be ideal.

At the moment, we only have the https://mne.tools/stable/generated/mne.read_epochs_eeglab.html, and the .fif reader that doesn't seem to work well with BIDS. I don't know why. I think it's because of the file name, but I'm not sure.

What I noticed is that, for brain vision and for EDF/BDF, reading the "epoch" file is possible with read_raw!

I could only find epoch files that I could not read with .set, but the others I could read with the raw without much problem.

drammock:
> We would need to implement the other epoch readers first in MNE

okay, that doesn't seem like a problem though

> What I noticed is that, for brain vision and for EDF/BDF, reading the "epoch" file is possible with read_raw!

but doesn't it return a `Raw` object though (with Epochs glued together as though they were continuous)?  We should make it do the right thing.

bruAristimunha:
super true @drammock! I validated with the OpenNeuro datasets, and in fact, there is a delivery error. I think I found an easy way to implement all this....

bruAristimunha:
heyyyyy! I was wondering @drammock, @scott-huberty, and @larsoner, did you have any time to take a look at this? I worked and now is working on all the types of epoch data :)

mne_bids/read.py
```diff
@@ -1336,6 +1339,18 @@ def read_raw_bids(
     if suffix is None:
         bids_path.update(suffix=datatype)
 
+    sidecar_json_fname = _find_matching_sidecar(
+        bids_path, suffix=datatype, extension=".json", on_error="ignore"
+    )
+    if sidecar_json_fname is not None:
+        with _open_lock(sidecar_json_fname, encoding="utf-8-sig") as fin:
+            recording_type = json.load(fin).get("RecordingType")
```
larsoner: Do we need to lock this open? Seems like we only need to lock when doing things that we expect to potentially *write* to files simultaneously. I don't see any problem with simultaneous reads for example

larsoner (commented):
LGTM other than one idea about not needing `_open_lock` here!

But would be good for @scott-huberty or @drammock to look, too, since they looked originally

mne_bids/read.py
scott-huberty: Also since #1531 I think we agreed that moving forward we should open JSON files with  UTF-8, because the JSON standard does not permit writing JSON files with a BOM.

bruAristimunha: solved!

mne_bids/read.py
```diff
@@ -1506,15 +1535,136 @@ def read_raw_bids(
...
+    return _attach_sidecars(epochs, bids_path, on_ch_mismatch=on_ch_mismatch)
+
+
+@verbose
+def _read_epochs_from_continuous(bids_path, *, extra_params=None, verbose=None):
+    """Slice a continuous file flagged ``RecordingType="epoched"`` into Epochs."""
+    sidecar_fname = _find_matching_sidecar(
+        bids_path, suffix=bids_path.datatype, extension=".json"
+    )
+    with _open_lock(sidecar_fname, encoding="utf-8-sig") as fp:
```
larsoner: Same comment here I think about both `sig` and `_open_lock`

larsoner (commented):
@scott-huberty @bruAristimunha any thoughts on getting this in for 0.19 vs early in 0.20 cycle?

mne_bids/read.py
bruAristimunha: done

mne_bids/read.py
```diff
@@ -1506,15 +1535,136 @@ def read_raw_bids(
...
-    if return_event_dict:
-        return raw, event_id
     return raw
 
 
+@verbose
+def read_epochs_bids(
+    bids_path, extra_params=None, *, on_ch_mismatch="raise", verbose=None
+):
+    """Read pre-epoched BIDS data (RecordingType="epoched") as :class:`mne.Epochs`.
```
```suggestion
    """Read epoched BIDS data (RecordingType="epoched") as :class:`mne.Epochs`.
```

Total nitpick but I initially thought pre-epoch was referring to the Unix timestamp 😂

mne_bids/read.py
```diff
@@ -1506,15 +1535,136 @@ def read_raw_bids(
...
+    epochs : mne.Epochs
+        The epoched data with BIDS sidecar metadata applied.
+
+    Notes
+    -----
+    This function applies the same sidecars as :func:`read_raw_bids`
+    (``channels.tsv``, ``electrodes.tsv``, ``coordsystem.json``,
+    ``scans.tsv``, ``participants.tsv``).
+    """
+    if not isinstance(bids_path, BIDSPath):
```
scott-huberty: @larsoner are we allowed to use `mne.utils._validate_type` here? it's used elsewhere in MNE-BIDS but I know from previous discussions that we try not to use MNE-Python private helpers in this repo.

mne_bids/read.py
```diff
@@ -1506,15 +1535,136 @@ def read_raw_bids(
...
+        epochs = epoch_reader[ext](
+            bids_path.fpath, verbose=verbose, **(extra_params or {})
+        )
+    elif ext in _continuous_epoched_reader:
+        epochs = _read_epochs_from_continuous(
+            bids_path, extra_params=extra_params, verbose=verbose
+        )
+    else:
+        raise RuntimeError(
+            f"read_epochs_bids does not support {ext!r} epoched files; "
```
scott-huberty, line 1622: Just curious.. Is there a reason tht we want to evaluate this expression via `repr()` instead of `str()`?

scott-huberty (approved):
LGTM thanks @bruAristimunha !! My only request is that we add an example/tutorial to the `examples` directory. Would this be easy to do?

mne_bids/read.py
bruAristimunha: Applied, thanks! Also updated the changelog entry to match.

mne_bids/read.py
bruAristimunha: Switched to `_validate_type` in 4da555b — it's already used in `write.py` and `sidecar_updates.py` for this exact check, so I think we're fine on that front.

mne_bids/read.py
bruAristimunha: `!r` wraps the extension in quotes in the error message (`does not support '.fif'`), which reads more clearly than the bare string — same convention MNE uses elsewhere. Happy to change it if you prefer `str()`!

bruAristimunha:
@scott-huberty done! Added `examples/read_epochs_bids.py` in 4da555b — it builds a small epoched EEGLAB BIDS dataset from an EEGBCI motor-imagery run, shows `read_raw_bids` refusing it with the helpful error, reads it back with `read_epochs_bids`, and plots the evoked response. Also applied the docstring suggestion and switched to `_validate_type`. Should be ready for another look :)

larsoner:
@bruAristimunha sorry there is a conflict now, can you fix it?

larsoner:
Actually I took care of the conflict, looks like you addressed comments, marking for merge-when-green. Thanks in advance @bruAristimunha !

PR #1607, opened 2026-05-04
FIX: warn instead of raise when subject row missing from participants.tsv

### LINKED ISSUE #1606: read_raw_bids raises ValueError when subject row is missing from participants.tsv
### Problem

When `participants.tsv` exists but does not contain a row for the subject being read, `read_raw_bids` raises a `ValueError` from inside `_handle_participants_reading`, aborting the read entirely:

```
File ".../mne_bids/read.py", line 297, in _handle_participants_reading
    row_ind = _verbose_list_index(subjects, subject, allow_all=True)
File ".../mne_bids/read.py", line 289, in _verbose_list_index
    raise ValueError(f"{exc}{extra}") from None
ValueError: 'sub-XX' is not in list. Did you mean one of [...]?
```

This is asymmetric with the case where `participants.tsv` is **entirely absent**, which already warns and proceeds:

```python
# read.py:1440-1446
if participants_tsv_path.exists():
    raw = _handle_participants_reading(...)   # raises if subject row absent
else:
    warn(f"participants.tsv file not found for {raw_path}")
    raw.info["subject_info"] = dict()
```

`_handle_participants_reading` only writes to `raw.info["subject_info"]` (sex, hand, height, weight, age, …). It does not touch the signal, channels, or timing — so a missing row has the same practical impact as a missing file.

### Affected OpenNeuro datasets (from a recent EEG/MEG read sweep)

| Dataset | Records affected | Pattern |
|---|---|---|
| `ds003645` | 8 | `sub-emptyroom` sessions not listed in participants.tsv |
| `ds004816` | 5 | `sub-15..sub-19` missing (participants.tsv lists only sub-13/14/15) |
| `ds005540` | 3 | `sub-57..sub-59` missing |
| `ds005697` | 1 | `sub-16` missing |

In each case the recording itself is fine — only the participants.tsv row is absent.

### Proposed fix

Treat the missing-row case the same as the missing-file case: emit a warning and leave `raw.info["subject_info"]` empty. No new public parameter; the helper just early-returns.

PR forthcoming.

bruAristimunha:
Closes #1606.

## Summary

When `participants.tsv` exists but does not contain a row for the requested subject, `read_raw_bids` raised a `ValueError` and aborted the read. Now it emits a warning and leaves `raw.info[\"subject_info\"]` empty — the same behavior as when `participants.tsv` is entirely absent.

`_handle_participants_reading` only writes `raw.info[\"subject_info\"]` (sex, hand, height, weight, age, …); it never touches the signal, channels, or timing. So a missing row is functionally equivalent to a missing file, and treating them the same way unblocks reading datasets where the curators forgot to update `participants.tsv` after adding subjects.

## Impact

Datasets observed in a recent EEG/MEG read sweep that this unblocks:

| Dataset | Records affected |
|---|---|
| `ds003645` | 8 (`sub-emptyroom` sessions) |
| `ds004816` | 5 |
| `ds005540` | 3 |
| `ds005697` | 1 |

## Diff

Three files, +15 lines, 0 deletions:

- `mne_bids/read.py` (+4): four-line guard in `_handle_participants_reading` that early-returns with a warning when the subject is absent from the TSV. No signature change, no new parameter.
- `mne_bids/tests/test_read.py` (+10): regression case appended to `test_read_participants_data` — rewrites `participants.tsv` to drop the subject and asserts the warning + empty `subject_info`.
- `doc/whats_new.rst` (+1): bug-fix entry under 0.19.

## Backward compatibility

Anyone catching `ValueError` from `read_raw_bids` for this specific case will lose the exception, but a `RuntimeWarning` is emitted in its place.

## Test plan

- [x] `pytest mne_bids/tests/test_read.py -k participants` — 7 passed locally
- [x] New regression case `test_read_participants_data` covers the missing-row branch

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -75,6 +75,7 @@ Detailed list of changes
 - Allow non-numeric ``run`` entities (e.g. ``run-5H``) in :class:`mne_bids.BIDSPath` when ``check=False``, by `Bruno Aristimunha`_ (:gh:`1601`)
 - :func:`mne_bids.write_raw_bids` now writes an ``*_electrodes.json`` sidecar so derivative datasets pass the BIDS validator, by `Bruno Aristimunha`_ (:gh:`1545`)
 - :func:`mne_bids.read_raw_bids` now strips whitespace-padded ``n/a`` cells and normalizes European-locale decimal commas in TSV sidecars, by `Bruno Aristimunha`_ (:gh:`1599`)
+- :func:`mne_bids.read_raw_bids` now warns (instead of raising ``ValueError``) when ``participants.tsv`` exists but does not list the requested subject, leaving ``raw.info["subject_info"]`` empty. This unifies behavior with the existing missing-``participants.tsv`` path, by `Bruno Aristimunha`_ (:gh:`1606`)
 
 ⚕️ Code health
 ^^^^^^^^^^^^^^
--- a/mne_bids/read.py
+++ b/mne_bids/read.py
@@ -297,6 +297,10 @@ def _verbose_list_index(lst, val, *, allow_all=False):
 def _handle_participants_reading(participants_fname, raw, subject):
     participants_tsv = _from_tsv(participants_fname)
     subjects = participants_tsv["participant_id"]
+    if subject not in subjects:
+        warn(f"Subject {subject!r} is not listed in {participants_fname.name}")
+        raw.info["subject_info"] = dict()
+        return raw
     row_ind = _verbose_list_index(subjects, subject, allow_all=True)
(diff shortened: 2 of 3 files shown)


scott-huberty (approved):
LGTM! 

I know `participant.tsv` is optional, though I think that _if it does exist_, then there MUST be an entry for each participant: 

> The participant_id entries MUST be a superset of all subject directories and all participant_id entries found among phenotypic and assessment data in the phenotype/ directory.

From https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html

In any case, this aligns with our "permissive on read, strict on write" approach for MNE-BIDS

Issue #1608, opened 2026-05-13
refactor / dedup channel names getting reconciled


drammock:
the process of reading the `channels.tsv` is getting hard for me to reason about.

Reconciliation between raw channel names and `channels.tsv` happens in multiple places: Deduplication is handled in `_handle_channels_reading`, mismatches between `channels.tsv` and `raw.ch_names` happens in `_handle_channels_mismatch`.. Except for the special-case mismatch of `STI 014` being present in `raw` but not `channels.tsv`.. That is handled in `_handle_channels_reading`..

Also, now, `_handle_channels_mismatch is True` means "skip setting channels metadata on the raw object", which isn't very intuitive..

It might be worthwhile to centralize all this data munging in a helper function that is called very early in `_handle_channels_reading`. Then the rest of `_handle_channels_reading` can focus on applying channel types/units and `'bads'` etc.

Perhaps the MNE sprint this summer will be a good time to do such a refactor.

_Originally posted by @scott-huberty in https://github.com/mne-tools/mne-bids/pull/1603#pullrequestreview-4265404427_

PR #1610, opened 2026-05-19
MAINT: extract _reconcile_channel_names from _handle_channels_reading

bruAristimunha:
Pure no-op refactor addressing #1608.

Channel-name reconciliation between `raw.ch_names` and `channels.tsv` was spread across four blocks of `_handle_channels_reading`:

1. duplicate names within `channels.tsv` (dedup / warn / raise)
2. dropping a synthesized `STI 014` not present in the sidecar
3. length mismatch between sidecar and raw (warn-and-skip)
4. name-order mismatch (dispatch to `_handle_channel_mismatch`)

This PR moves all four into a new `_reconcile_channel_names(raw, channels_dict, on_ch_mismatch, channels_fname) -> bool` helper called once near the top of `_handle_channels_reading`. The helper returns `True` when the caller should bail out and skip applying types / units / bads. After the call, `_handle_channels_reading` focuses on what its name suggests — applying channels.tsv-derived metadata to an already-aligned `raw`.

Drive-by: `_handle_channel_mismatch` now returns an explicit `False` on the reorder/rename paths so the call site doesn't need a `bool(...)` cast to interpret `True | None`.

### Why now

@scott-huberty flagged the dispersion in https://github.com/mne-tools/mne-bids/pull/1603#pullrequestreview-4265404427 and Dan filed #1608 suggesting the sprint as a possible window. Doing the structural part now is cheap because the bodies move byte-for-byte; behavior-changing follow-ups (e.g., dropping the defensive `if ch_name in raw.ch_names` filters now that alignment is guaranteed when the helper returns `False`-and-non-length-mismatch) can land separately.

### Verification

- All 79 tests in `mne_bids/tests/test_read.py` pass unchanged.
- Channel-related tests in `mne_bids/tests/test_write.py` pass unchanged.
- Bodies of moved code are byte-for-byte identical to the pre-refactor state; only the `return raw` → `return True` adjustment for the dedup-warn path and the added `return False` in `_handle_channel_mismatch` change.
- Pre-commit (ruff + ruff-format) clean.

### Not included

- No `whats_new.rst` entry — pure internal refactor with no user-visible effect.
- The defensive `if ch_name in raw.ch_names` filters in the types/units/bads loops are kept as-is; tightening them is a behavior change worth its own PR.


larsoner (approved):
Okay for me, but maybe a good early PR for 0.20 since it's an internal refactoring

scott-huberty (approved):
Thanks for doing this @bruAristimunha ! This LGTM.

One (style) nitpick.. Right now `_reconcile_channel_names == True` or `_handle_channel_mismatch == True` means _don't_ apply channel metadata to the raw object (because something went wrong). What if we inverted that logic?

Then the function call would read like

```python
if not _reconcile_channel_names(raw, ...):
  return raw

# code that sets types, units, coils, etc. below
```

This reads more intuitively to me, but it is admittedly a style preference. WDYT?

bruAristimunha:
@scott-huberty done in 266d5ab4 — inverted the return semantics as suggested: `_reconcile_channel_names` (and `_handle_channel_mismatch`) now return `True` when channels.tsv metadata can be applied, so the call site reads `if not _reconcile_channel_names(...): return raw`.

PR #1611, opened 2026-05-22
Prep for 0.19 release

scott-huberty:


doc/_static/versions.json
```diff
@@ -1,12 +1,22 @@
 [
     {
-        "name": "0.18 (devel)",
```
scott-huberty, line 3: Maybe I'm missing the point but I don't think this was updated after the last release (0.18). I would expect this to say 0.19 (devel) and I would expect the dict below to say "0.18 (stable)"

drammock: I think you're right. Need to add another new entry, so end result is:

- 0.20 (dev)
- 0.19 (stable)
- 0.18
- 0.17 
- ...etc

drammock (approved):
ok @scott-huberty feel free to merge when you're ready to finish the release process!

PR #1612, opened 2026-05-22
FIX: montage deprecation

scott-huberty:
We need to address some[ CI failures](https://github.com/mne-tools/mne-bids/actions/runs/26308246239/job/77451229046?pr=1611) after https://github.com/mne-tools/mne-python/pull/13903/


larsoner (approved):
Nice, thanks @scott-huberty

PR #1620, opened 2026-06-15
Speed up write_raw_bids for recordings with many channels

sappelhoff:
(PR and content generated with the help of Claude)

Three small, behavior-preserving optimizations on the channel-writing hot path in `write_raw_bids`. Each was verified to produce identical output to the previous code.

### Changes

1. **`pick.get_coil_types()` — cache the constant dict.**
   `coil_type()` is called once per channel and rebuilt this large dict of FIFF constants every time. It is now built once via `functools.cache`. (read-only, internal use only)

2. **`_sidecar_json()` — single-pass channel counting.**
   Replaced 11 separate `len([ch for ch in raw.info["chs"] if ch["kind"] == ...])` comprehensions (11 full passes building throwaway lists) with one `Counter` pass.

3. **`_write_raw_brainvision()` — fix quadratic unit loop.**
   The `"NA" -> "n/a"` fixup comprehension lived *inside* the per-channel loop, rebuilding the entire `unit` list on every iteration (O(n²)). Moved it to a single pass after the loop. The fixup is idempotent, so the result is identical.

### Measured impact (microbenchmarks, representative channel counts)

| Change | Before | After | Speedup |
|---|---|---|---|
| coil_type ×300 channels | 0.320 ms | 0.161 ms | ~2× |
| channel counting (279 ch) | 111.6 µs | 19.3 µs | ~5.8× |
| BrainVision unit loop (270 ch) | 655 µs | 35 µs | ~18.8× |

### Testing

- Equivalence checks confirm the new channel-counting and unit-loop code produce identical results to the originals across all channel kinds (including the `FIFF_UNIT_NONE -> "NA"` case).
- `test_write.py`, `test_pick.py`, `test_read.py`: 215 passed, 0 new failures. (The 7 pre-existing `curry` failures and skips are environmental and also fail on `main`.)

No behavior or output changes.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -51,6 +51,6 @@ Detailed list of changes
 ⚕️ Code health
 ^^^^^^^^^^^^^^
 
-- None yet
+- Sped up writing of recordings with many channels by avoiding redundant per-channel work in :func:`mne_bids.write_raw_bids` (single-pass channel-type counting, cached coil-type lookup, and a fixed quadratic loop when writing BrainVision units), by `Stefan Appelhoff`_ (:gh:`1620`)
 
 :doc:`Find out what was new in previous releases <whats_new_previous_releases>`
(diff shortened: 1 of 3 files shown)


larsoner (approved):
These are simple enough to get in, thanks @sappelhoff !

PR #1627, opened 2026-06-24
Adds support for writing "Other" sex

### LINKED ISSUE #1626: Supporting writing "Other" Sex from MNE-BIDS
Is there a way that writing a value of "Other" for sex can be supported in MNE-BIDS?

PR #1567 made changes to how sex metadata is written, including "Other" in the `participants.tsv` sidecar, but as far as I can tell there is no way for MNE-BIDS to write "Other" to the sex column of the tsv file.

This is partly due to the in-built mapping from MNE to BIDS not including a value for "Other"

https://github.com/mne-tools/mne-bids/blob/2d60804689edee3d9689d09e700b4fb727af91a2/mne_bids/config.py#L684-L691

Looking deeper into this, I suspect it may be because EDF files only support M&F as options for Sex and passing values other than 0,1, or 2 raises an error.

ubdbra001:
PR Description
--------------

Resolves #1626 

- Adds the ability to write "Other" sex
- Raises errors when writing to file formats that do not support "Other" sex

Merge checklist
---------------

If applicable:

- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_bids/config.py
+++ b/mne_bids/config.py
@@ -682,12 +682,13 @@ HAND_BIDS_TO_MNE = {
 HAND_MNE_TO_BIDS = {0: "n/a", 1: "R", 2: "L", 3: "A"}
 
 SEX_BIDS_TO_MNE = {
-    ("n/a", "other", "o", "O", "OTHER", "Other"): 0,
+    ("n/a"): 0,
     ("male", "m", "M", "MALE", "Male"): 1,
     ("female", "f", "F", "FEMALE", "Female"): 2,
+    ("other", "o", "O", "OTHER", "Other"): 3,
 }
 
-SEX_MNE_TO_BIDS = {0: "n/a", 1: "M", 2: "F"}
+SEX_MNE_TO_BIDS = {0: "n/a", 1: "M", 2: "F", 3: "O"}
 
 
 def _map_options(what, key, fro, to):
--- a/mne_bids/write.py
+++ b/mne_bids/write.py
@@ -1506,6 +1506,12 @@ def _write_raw_edf_bdf(raw, bids_fname, overwrite, *, physical_range="auto"):
     """
     ext = bids_fname.suffix[1:].upper()
     assert ext in ("EDF", "BDF")
+
+    subject_info = raw.info.get("subject_info", None)
+
+    if subject_info is not None and subject_info.get("sex", None) == 3:
+        raise ValueError("EDF/BFD files only support M (1) and F (2) sex options")
+
     if raw.info["meas_date"] is not None and raw.info["meas_date"].year < 1985:
         warn(
             f"Attempting to write a {ext} file with a meas_date of "


ubdbra001:
Looking at the specs for the EEG file types:
- `EDF`/`BDF`: can only handle M/F so need to raise an error
- `BrainVision`: doesn't store any info about Sex, so no change needed
- `EEGLAB`: doesn't store any info about Sex, so no change needed

(Happy to be corrected about any of these)

I'm less familiar with file formats for other recording types, are these the additional formats that need to be accounted for:
- MEG: `.con`, `.sqd`, `.fif`, `.pdf`, `.ds`
- iEEG:  `.mefd`, `.nwb`
- NIRS: `.snirf`
- Motion: `.tsv`

drammock:
> I'm less familiar with file formats for other recording types, are these the additional formats that need to be accounted for

Sorry for the slow response! Can I just ask a clarifying question here: I thought the goal was getting MNE-BIDS to write "Other" into the `participants.tsv` file... so why do we need to care about *writing* sex metadata in the various raw data formats supported by BIDS?  IIUC the json sidecars / .tsvs are supposed to be the source-of-truth for things like participant sex, and the metadata in the ephys data files only matters at the time when you're using MNE-BIDS to create the BIDS dataset from the original recordings (in other words, we need to be able to *read* the sex metadata in ephys files, but don't need to care about *writing* it).  Can @scott-huberty, @sappelhoff, or perhaps @hoechenberger confirm that intuition?

ubdbra001:
My experience with using MNE-BIDS and writing "Other" to E/BDF files is that it will fail as MNE-BIDS will try and write the sex to the data file.

There are two ways to go with this:
1. Error on trying to write an invalid sex
2. Don't write sex at all, and (as you say) use the metadata files as the source of truth

I'm happy with either to be honest.

scott-huberty:
> Can @scott-huberty, @sappelhoff, or perhaps @hoechenberger confirm that intuition?

Yes, MNE-BIDS will typically defer to what is written in sidecar metadata over what is written in modality specific data files (.fif, .edf etc)



> My experience with using MNE-BIDS and writing "Other" to E/BDF files is that it will fail as MNE-BIDS will try and write the sex to the data file.
> 
> There are two ways to go with this:
> 
> 1. Error on trying to write an invalid sex
> 2. Don't write sex at all, and (as you say) use the metadata files as the source of truth
> 
> I'm happy with either to be honest.

How about 2:,  if a file format (e.g. edf) cannot represent a BIDS metadata  value ("Other"), we don't write anything at all. In the case of "Male" we can write "M". In either case, MNE-BIDS will use BIDS sidecar files (e.g. `participants.tsv`) as the source of truth, and the presence of this metadata in the M/EEG files (.fif, .edf) is only for posterity/convenience.

ubdbra001:
Sure, that works. I'll update the code.
Does this mean I shouldn't worry too much about other potential output file formats?

scott-huberty:
> Does this mean I shouldn't worry too much about other potential output file formats?

Yeah I personally think so (you dont have to worry). Since storing this info in the data file itself is not mission critical to MNE-BIDS, I think that this PR can focus on the bug you found in `_write_raw_edf_bdf`! Can you add a test too? Since you had to report this regression, it sounds like we don't have a test to catch it.

PR #1631, opened 2026-07-10
MAINT: enable ssh debugging

scott-huberty:
Allows for ssh tunneling into the CI environments, for interactive debugging. I would like this feature so e.g. I can gain better visibility into the failures over at #1630, which I can't replicate locally.

I basically just copied the `warpbuild/action-debugger` configuration over at [MNE-Python](https://github.com/mne-tools/mne-python/blob/31a73266fb6ee0aed10de777ddaa4d8474da315f/.github/workflows/tests.yml#L109).

The only difference between here and mne-python is that I set `limit-access-to-actor: true` in the `action-debugger` config. My understanding from their [documentation](https://github.com/WarpBuilds/action-debugger#use-registered-public-ssh-keys) is that by default, if the ssh session is triggered by a user who _does not have any public ssh keys on their GitHub profile_, then `action-debugger` will create an ssh link that _anyone can access_. `limit-access-to-actor: true` requires one to have a public ssh key associated with their GitHub profile before they can trigger an interactive ssh job, which ensures that only they can access the shh tunnel.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,6 +12,14 @@ on:
     branches: ['**']
   schedule:
     - cron: "0 4 * * 1"  # every Monday at 4:00 UTC
+  # adapted from mne-python which adapted theirs from spyder-ide/spyder
+  workflow_dispatch:
+    inputs:
+      ssh:
+        description: 'Enable ssh debugging'
+        required: false
+        default: false
+        type: boolean
 
 jobs:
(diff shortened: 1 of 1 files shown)


scott-huberty:
ssh tunneling works!! 

Before merge, needs a fix for #1632 either in this PR or a new PR. I will wait for a consensus among devs regarding the best approach to #1632

PR #1633, opened 2026-07-10
FIX: Don't let lock files persist in BIDS datasets

### LINKED ISSUE #1632: Lock files should not persist in BIDS datasets

https://github.com/mne-tools/mne-bids/blob/73ee63730e4462d6b45407cde7e1af297ad1f58e/mne_bids/_fileio.py#L122-L125

This behavior is causing our tests to fail. We should not write persistent `*.lock` files to BIDS datasets because that breaks BIDS compliance (i.e., `bids-validator` will raise). We should either figure out a way to teardown the lock files after writing is completely finished, or write these persistent lock files to some temporary directory that is outside the BIDS dataset root..(?)

### Steps to reproduce

```Python
pytest -x mne_bids/tests/test_write.py in an environment with `filelock>=3.20.4` installed
```

### Expected results

pass

### Actual results

`subprocess.CalledProcessError: Command '['/usr/local/bin/deno', '-A', 'jsr:@bids/validator', PosixPath('/private/var/folders/z1/jw4bwv7s13b66b4cz2_m6pzw0000gn/T/pytest-of-scotterik/pytest-12/test_write_participants0')]' returned non-zero exit status 16.`

### Additional information

n/a

scott-huberty:
Fixes #1632 

Found `SoftFileLock` on the filelock doc:

- https://py-filelock.readthedocs.io/en/latest/api.html#filelock.SoftFileLock
- https://py-filelock.readthedocs.io/en/latest/how-to.html#detect-stale-locks-soft-locks-only

IIUC this will remove the lock file for us when the process that created it finishes. 

The important thing for the team to know is that `SoftFileLock` uses a different locking mechanism than `FileLock`. `FileLock` will use OS-level locking mechanisms, `SoftFileLock` uses the existence of the lock file itself as the locking mechanism.

Using `SoftFileLock` makes the tests pass again locally. If the CI's pass too, I think this is one way to fix the issue, but maybe in the medium-to-long-term there are better ways, (EDIT) as I imagine that the [concerns about unlinking ](https://py-filelock.readthedocs.io/en/latest/api.html#filelock.UnixFileLock)a lock file on unix systems still applies when using `SoftFileLock`

### DIFF
--- a/mne_bids/_fileio.py
+++ b/mne_bids/_fileio.py
@@ -95,7 +95,7 @@ def _get_lock_context(path, *, timeout=None, lock=True):
         try:
             # Ensure parent directory exists
             Path(lock_path).parent.mkdir(parents=True, exist_ok=True)
-            lock_context = filelock.FileLock(
+            lock_context = filelock.SoftFileLock(
                 str(lock_path),
                 timeout=timeout,
             )
--- a/mne_bids/tests/test_fileio.py
+++ b/mne_bids/tests/test_fileio.py
@@ -144,7 +144,7 @@ def test_get_lock_context_oserror(tmp_path):
     test_file = tmp_path / "test.txt"
     test_file.write_text("test")
 
-    with patch.object(filelock, "FileLock", side_effect=_mock_filelock_oserror):
+    with patch.object(filelock, "SoftFileLock", side_effect=_mock_filelock_oserror):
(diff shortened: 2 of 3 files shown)


larsoner (approved):
Seems okay to me and solves a real problem, so I'll mark for merge-when-green. Thanks for investigating and fixing @scott-huberty !

PR #1637, opened 2026-07-18
Fix BIDSPath.match roots for nested datasets

### LINKED ISSUE #1561: BIDSPath.match() can return duplicates (and files that don't exist)
See

https://github.com/mne-tools/mne-bids/pull/1560/changes#diff-fa823e8c258bfc63ca666a1954e56b00ceb6816e16fecef3ba6cbffaa33de9b0R326

Where it returns hundreds of values when there are only 36 unique ones. This is due to finding paths like
```
a/b/something
a/c/something
```
in the initial scan and then replacing them with
```
bids_root/something
bids_root/something
```
I am not sure the correct fix here... deduplication is possible, but it's weird because the `a/b` and `a/c` are being replaced with `bids_root` so the `bids_root/something` file might not actually even exist in the end!

viranovskaya:
Fixes #1561.

`BIDSPath.match()` now keeps the root inferred from each matched filename when replacing it with the requested root would point somewhere else. This keeps same-named files in `sourcedata` and `derivatives` distinct and ensures that the returned paths refer to files that exist.

The regression test covers identical filenames in two nested BIDS roots. The existing dense-tree benchmark now also checks that every match is unique and exists.

Checks:

- `ruff check mne_bids/path.py mne_bids/tests/test_path.py`
- `ruff format --check mne_bids/path.py mne_bids/tests/test_path.py`
- 19 related path matching, filtering, and parsing tests passed
- dense-tree path benchmark passed

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -48,7 +48,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
-- Preserve the actual root of files returned by :meth:`mne_bids.BIDSPath.match`, avoiding duplicate paths and paths that do not exist when matching inside nested BIDS roots, by `Daria Agafonova`_ (:gh:`1561`)
+- Preserve the actual root of files returned by :meth:`mne_bids.BIDSPath.match`, avoiding duplicate paths and paths that do not exist when matching inside nested BIDS roots, by `Daria Agafonova`_ (:gh:`1637`)
 - :func:`mne_bids.write_raw_bids` no longer raises a ``TypeError`` when writing ANT Neuro eego recordings (``.cnt``), by `Vincent Gao`_ (:gh:`1617`)
 
 ⚕️ Code health
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -48,6 +48,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
+- Preserve the actual root of files returned by :meth:`mne_bids.BIDSPath.match`, avoiding duplicate paths and paths that do not exist when matching inside nested BIDS roots, by `Daria Agafonova`_ (:gh:`1561`)
 - :func:`mne_bids.write_raw_bids` no longer raises a ``TypeError`` when writing ANT Neuro eego recordings (``.cnt``), by `Vincent Gao`_ (:gh:`1617`)
 
 ⚕️ Code health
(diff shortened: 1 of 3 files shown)


larsoner:
Let me know when I should look! I restarted the failing CI, it just had a download error

viranovskaya:
Thanks! I added one more regression check covering both the root-level file and the same filename under `sourcedata` and `derivatives`. The path tests and style checks pass, so this is ready for review.

mne_bids/tests/test_path.py
```diff
@@ -492,13 +491,9 @@ def test_path_benchmark(bids_root_dense, monkeypatch, path_counter):
         root=tmp_bids_root,
     )
     paths = path.match()
-    # TODO: We *should* have n_subjects * n_sessions of these, but we get way more:
     assert len(paths) == 684
```
larsoner: Was the TODO wrong here? If so can you update it to say why 684 is correct (or if it's incorrect, what we should expect it to be)? The goal is to eventually get these to the expected number and if it's not there at least say what it should be...

viranovskaya: Yes, the old TODO was misleading. The fixture creates one .ds recording for each subject-session pair in the main dataset, sourcedata, and 17 derivative roots: 9 × 4 × 19 = 684. I’ll make that calculation explicit in the test instead of leaving 684 as an unexplained constant.

viranovskaya: I updated the test to calculate the expected number from the fixture instead of leaving 684 as an unexplained constant.

doc/whats_new.rst
```diff
@@ -48,6 +48,7 @@ Detailed list of changes
 🪲 Bug fixes
 ^^^^^^^^^^^^
 
+- Preserve the actual root of files returned by :meth:`mne_bids.BIDSPath.match`, avoiding duplicate paths and paths that do not exist when matching inside nested BIDS roots, by `Daria Agafonova`_ (:gh:`1561`)
```
larsoner: We usually cite the PR not the issue
```suggestion
- Preserve the actual root of files returned by :meth:`mne_bids.BIDSPath.match`, avoiding duplicate paths and paths that do not exist when matching inside nested BIDS roots, by `Daria Agafonova`_ (:gh:`1637`)
```

larsoner (approved):
I'll push this tiny change and merge, thanks in advance @viranovskaya !

Issue #1342, opened 2024-11-25
Eyetracking BIDS


scott-huberty:
### Describe the problem

xref [BEP 020](https://github.com/bids-standard/bids-specification/pull/1128) : The community is converging on a BIDS specification for eye-tracking data. MNE-Python has also supported eye-tracking data since [v1.4] (https://mne.tools/1.5/whats_new.html#version-1-4-0-2023-05-10) (May 2023).

_Once BEP 020 is merged_, actually implementing software to help researchers read/write BIDS compliant eye-tracking data is the natural next step.

### Describe your solution

Since more folks are working with eyetracking data in MNE, I propose that MNE-BIDS is a good place to implement this support.

I imagine that only the new schema validator will be able to validate eye tracking BIDS data (assuming that the legacy validator  that is written in JS will not be updated). So if #1302 is not merged soon, I'd also be willing to help bring that across the finish line.

How do folks feel about supporting I/O for eyetracking BIDS?

### Describe possible alternatives

Don't support I/O for BIDS compliant eyetracking data, at least not yet.

### Additional context

I will be applying for grant funding (USSRI) to request funding that would allow me to dedicate a portion of my time to do this.
If I am not granted the award, I will still be interested in this to the extent that my projects would benefit from it, but I may not be able to dedicate as much time to the effort.

sappelhoff:
> Since more folks are working with eyetracking data in MNE, I propose that MNE-BIDS is a good place to implement this support.

I'd be very happy to see that!

> I imagine that only the new schema validator will be able to validate eye tracking BIDS data

100%

>  (assuming that the legacy validator that is written in JS will not be updated)

yes, that will not happen

> So if https://github.com/mne-tools/mne-bids/pull/1302 is not merged soon, I'd also be willing to help bring that across the finish line.

Help in the form of commits pushed to that branch (or PRs against my branch / etc.) would be very very welcome!

> How do folks feel about supporting I/O for eyetracking BIDS?

I think we need it!

Your support and contributions would be very welcome @scott-huberty!

PR #1642, opened 2026-08-10
ENH: write eyetracking data

scott-huberty:
Towards #1342 

This strips out the previously written eyetrack writing code from #1512 in order to keep the diff manageable.

FYI this Adds a parameter `eyetrack_calibration` to `write_raw_bids`, which must be passed when the Raw data include eyetrack channels, as BIDS wants metadata about the experimental display that can't be automatically inferred.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,12 +13,19 @@ repos:
         # D205: 1 blank line required between summary line and description
         args: ["--ignore=D103,D400,D205"]
         files: ^examples/
+      - id: ruff
+        name: ruff doc/
+        files: ^doc/
+        args: ["--fix"]
       - id: ruff-format
         name: ruff format mne_bids/
         files: ^mne_bids/
       - id: ruff-format
         name: ruff format examples/
         files: ^examples/
+      - id: ruff-format
+        name: ruff format doc/
+        files: ^doc/
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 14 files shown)


scott-huberty:
OK ready for review. Diff is big but about 500 LOC are the tutorial and tests. The tests cover writing monocular data, binocular data, eyetracking data alonsgside EEG, and they al pass the BIDS validator. 

@bruAristimunha do you have time to provide a review?

EDIT: [rendered tutorial](https://output.circle-artifacts.com/output/job/cbd30a36-3f52-4786-886c-761247282b97/artifacts/0/dev/auto_examples/convert_eyetracking_to_bids.html), [rendered API-doc](https://output.circle-artifacts.com/output/job/cbd30a36-3f52-4786-886c-761247282b97/artifacts/0/dev/generated/mne_bids.write_raw_bids.html)

EDIT2: I also directed QWEN3 coder to provide a PR review, and I fixed issues that were flagged by it (hopefully to save reviewer time).

larsoner (approved):
@scott-huberty I had a look and it all looks reasonable to me! Can you see if you're okay with the commit I pushed? A couple of things are just `doc/` ruff-formatting, but the more substantial one has to do with saving the annotations per-eye instead of both eye annotations landing in the file for each eye.

scott-huberty:
Ah good catch.. This BEP was annoying to implement!! Thanks @larsoner !

Issue #1371, opened 2025-02-16
Attention! Removal of INSPECT module (inspect_dataset)


sappelhoff:
We are planning to move the mne-bids INSPECT module, see also:

- https://github.com/mne-tools/mne-bids/issues/1366

If you are actively using this module and would like to continue doing so, please respond to this thread.

https://mne.tools/mne-bids/stable/generated/mne_bids.inspect_dataset.html

drammock:
URL in issue description is 404 (extra `.` before `html` suffix, but just removing it is also a 404 page). @sappelhoff what was the intended link?

sappelhoff:
Thanks Dan, it was not what I intended. Edited now to point to https://mne.tools/mne-bids/stable/generated/mne_bids.inspect_dataset.html

PR #1648, opened 2026-08-18
Parallelize CI testing and doc build

larsoner:
I had Claude Opus 5 port over the recent MNE-Python changes to make tests parallel using `pytest-xdist` and docs via `-j auto`. Should speed up both here.

### DIFF
--- a/mne_bids/conftest.py
+++ b/mne_bids/conftest.py
@@ -2,6 +2,7 @@
 
 # Authors: The MNE-BIDS developers
 # SPDX-License-Identifier: BSD-3-Clause
+import os
 from datetime import datetime
 
 import mne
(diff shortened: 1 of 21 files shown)


mne_bids/conftest.py
```diff
@@ -13,6 +14,15 @@ def pytest_configure(config):
     """Configure pytest options."""
     # Fixtures
     config.addinivalue_line("usefixtures", "monkeypatch_mne")
+    # inspect_dataset() calls plt.show(block=True) on any backend but agg, so a test
+    # that reaches it before something else has switched backends hangs forever
+    os.environ.setdefault("MPLBACKEND", "Agg")
```
scott-huberty, line 19: Sounds like a bug (or at least, something that can be improved). But are we still planning on removing the inspect module? xref https://github.com/mne-tools/mne-bids/issues/1371

scott-huberty (approved):

PR #1658, opened 2026-09-02
MAINT: migrate additional path handling to pathlib

gugu1031:
## Summary

- migrate additional path construction in `mne_bids/path.py` from `os.path` helpers to `pathlib`
- preserve string boundaries used by scans TSV handling and legacy BTi path inference
- add a regression test for legacy BTi directory inference
- add the required first-time contributor metadata

This is part of #1268.

## Testing

- `pre-commit run --files mne_bids/path.py mne_bids/tests/test_path.py doc/authors.rst CITATION.cff doc/whats_new.rst`
- `python -m pytest mne_bids/tests/test_path.py -q --disable-pytest-warnings`
  - 63 passed, 38 skipped; all skipped tests require the MNE testing dataset
- `git diff --check`

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -20,6 +20,7 @@ Version 0.20 (unreleased)
 The following authors contributed for the first time. Thank you so much! 🤩
 
 * `Daria Agafonova`_
+* `Jiayue Gu`_
 * `Vincent Gao`_
 
 The following authors had contributed before. Thank you for sticking around! 🤘
@@ -56,6 +57,10 @@ Detailed list of changes
 
 ⚕️ Code health
 ^^^^^^^^^^^^^^
+
+- Modernize path construction in :mod:`mne_bids.path` by replacing selected
+  ``os.path`` operations with ``pathlib.Path``, by `Jiayue Gu`_
+  (:gh:`1268`)
 
 - Sped up writing of recordings with many channels by avoiding redundant per-channel work in :func:`mne_bids.write_raw_bids` (single-pass channel-type counting, cached coil-type lookup, and a fixed quadratic loop when writing BrainVision units), by `Stefan Appelhoff`_ (:gh:`1620`)
 - Run the test suite with ``pytest-xdist`` and build the documentation with parallel Sphinx and Sphinx-Gallery workers, by `Eric Larson`_ (:gh:`1648`)
--- a/mne_bids/path.py
+++ b/mne_bids/path.py
@@ -135,7 +135,7 @@ def _find_empty_room_candidates(bids_path):
     candidate_er_fnames = []
     for session_dir in emptyroom_session_dirs:
         dir_contents = glob.iglob(
-            op.join(session_dir, datatype, f"sub-emptyroom_*_{datatype}*")
+            str(session_dir / datatype / f"sub-emptyroom_*_{datatype}*")
         )
         for item in dir_contents:
             item = Path(item)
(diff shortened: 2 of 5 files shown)


doc/whats_new.rst
```diff
@@ -57,6 +58,10 @@ Detailed list of changes
 ⚕️ Code health
 ^^^^^^^^^^^^^^
 
+- Modernize path construction in :mod:`mne_bids.path` by replacing selected
```
scott-huberty, line 61:
```suggestion
- Modernize path construction in  ``mne_bids.path`` by replacing selected
```

The use of ``:mod:`` here was causing the doc build to fail, and anyways I don't think we need to hyperlink to the source code (probably irrelevant for most users).

mne_bids/path.py
```diff
@@ -135,7 +135,7 @@ def _find_empty_room_candidates(bids_path):
     candidate_er_fnames = []
     for session_dir in emptyroom_session_dirs:
         dir_contents = glob.iglob(
-            op.join(session_dir, datatype, f"sub-emptyroom_*_{datatype}*")
+            str(session_dir / datatype / f"sub-emptyroom_*_{datatype}*")
```
scott-huberty, line 138:
```suggestion
            (session_dir / datatype).glob(f"sub-emptyroom_*_{datatype}*")
```

scott-huberty: 2 things: 

1) in Modern Python `glob.iglob` can handle pathlib objects, so no need for the `str` wrap. 

2) Better yet, since `session_dir` is now a `Path` object, we might as well use it's convenience methods!

mne_bids/path.py
```diff
@@ -530,15 +530,15 @@ def directory(self):
         """
         # Create the data path based on the available entities:
         # root, subject, session, and datatype
-        data_path = "" if self.root is None else self.root
+        data_path = Path() if self.root is None else self.root
```
scott-huberty, line 533: @larsoner my hunch is that since we already accept `self.root == None`, then it would be better to keep it `None` instead of resolving it to `Path()` (which actually points to the current directory). 

On the other hand, the old code provided a fallback (``""``), so the contributors suggestion keeps the behavior the same..

WDYT?

mne_bids/path.py
```diff
@@ -883,7 +883,7 @@ def rm(self, *, safe_remove=True, verbose=None):
             if not scans_fpath.exists():
                 continue
             # get the relative datatype of these bids files
-            bids_fnames = [op.join(p.datatype, p.fpath.name) for p in bids_paths]
+            bids_fnames = [str(Path(p.datatype) / p.fpath.name) for p in bids_paths]
```
scott-huberty, line 886:
```suggestion
            bids_fnames = [Path(p.datatype / p.fpath.name) for p in bids_paths]
```

Does downstream code allow us to keep `bids_fnames` as a list of Path objects?

mne_bids/path.py
```diff
@@ -1445,14 +1444,14 @@ def _get_matching_bidspaths_from_filesystem(bids_path):
...
+    bti_dir = data_dir / Path(basename).stem
+    if bti_dir_with_ext.is_dir():
         logger.info(f"Assuming BTi data in {bti_dir_with_ext}")
-        matching_paths = [bti_dir_with_ext]
-    elif op.isdir(bti_dir):
+        matching_paths = [str(bti_dir_with_ext)]
+    elif bti_dir.is_dir():
         logger.info(f"Assuming BTi data in {bti_dir}")
-        matching_paths = [bti_dir]
+        matching_paths = [str(bti_dir)]
```
scott-huberty, line 1454:
```suggestion
        matching_paths = [bti_dir]
```

Does downstream code allow us to use Path objects instead of str?

scott-huberty (changes requested):
Great start!

